Issue #5, opened 2021-09-23
namespace shouldn't be hard coded


subhamkrai:
currently, the namespace is hard coded as `rook-ceph`  we should be reading it from the argument.

subhamkrai:
@travisn can I get permission to assign myself as `Assigness`?

PR #12, opened 2021-10-25
core: add option to run command from other namespace

subhamkrai:
Earlier, `rook-ceph` namespace was hardcoded.
Now, we have the option to run commands from any namespace.

- [x] test the changes
- [x] update CI

Closes: https://github.com/rook/kubectl-rook-ceph/issues/5


subhamkrai:
with #14 we can run commands in another namespace. I'll close. Thanks for the reviews

PR #13, opened 2021-10-27
core: suggest a new skeleton script template

BlaineEXE:
The short of this is that I would really like to avoid spending time doing code reviews for the command and argument parsing boilerplate that this script is going to need. I'd rather start off with a robust argument/command parsing structure that we should be able to user with little to no modification for the forseeable future and start focusing code reviews on feature work.

I have seen bash argument parsing go terribly awry and be hugely cumbersome, and I really want to avoid that here if possible. I really can't stress enough how terribly difficult it is to write good argument parsing in Bash, and I'm sure there are some people who would say my suggestion here is bad as well, though I have done my best to make it have the features we need and be as easy to read and use as possible. @subhamkrai if you have questions about how or why things are implemented here, please feel free to ask. 

I don't anticipate that this PR will be merged, as it suggests a top-level `skeleton.sh` file that we don't want. Instead, please feel free to copy-paste the parts of this to a PR that revises the main script, and feel free to refer back to this when adding new commands and subcommands. 

@leseb or @travis, if you have suggestions/questions/comments/improvements, I'm happy to update what I'm doing here to make the example skeleton better.

-----
##### Commit text below
-----

Handling argument and command parsing in Bash is really difficult, and I
am interested in proposing a new skeleton script that is primarily aimed
at handling argument parsing for various types of commands we will likely
want to implement in the future. My goal is that this skeleton has an
easily-extensible structure to accept new commands easily and that the
argument parsing is robust and easy to reuse while being as easy to read
as possible.

It's my goal that we can port the existing functionality to
this script with little effort and have a good foundation for expanding
features without having to go through detailed back-and-forth about how
to implement the boilerplate for argument parsing in a piece-by-piece
manner. This will likely only result in us having a cumbersome script in
need of large refactoring down the line.

I designed the argument parser to accept long and short arguments both
with and without values.

I designed the skeleton to support and show quality examples of commands
including commands that have arguments and subcommands.

The following can be used to set the namespace to 'my-ns', for example.
   skeleton.sh -nmy-ns
   skeleton.sh -n my-ns
   skeleton.sh --namespace=my-ns
   skeleton.sh --namespace my-n

Or, this would enable verbose mode:
   skeleton.sh -v
   skeleton.sh --verbose

I implemented -h/--help flags and a 'help' command.

I implemented a placeholder 'ceph' command to run a ceph CLI command
with arbitrary args.

I implemented a placeholder 'status' command that supports a -l/--long
argument.

I implemented a placeholder 'operator' command with an additional
'restart' subcommand.

The skeleton also throws errors on common corner cases like
- a flag's value is another flag
- a flag's value is empty
- a flag isn't supported
- a command isn't supported
- a command wasn't given
- extraneous commands were given at the end of input

Signed-off-by: Blaine Gardner <blaine.gardner@redhat.com>

### DIFF
--- a/skeleton.sh
+++ b/skeleton.sh
@@ -1,0 +1,270 @@
+#!/usr/bin/env bash
+
+set -eEuo pipefail
+
+####################################################################################################
+# HELPER FUNCTIONS
+####################################################################################################
+
+function print_usage () {
+  echo ""
+  echo "DESCRIPTION"
+  echo "  blah blah blah"
+  echo "USAGE"
+  echo "  kubectl rook-ceph <main args> <command> <command args>"
+  echo "MAIN ARGS"
+  echo "  -v, --verbose               : output verbose info"
+  echo "  -h, --help                  : output help text"
+  echo "  -n, --namespace='rook-ceph' : the namespace of the CephCluster"
+  echo "COMMANDS"
+  echo "  help           : output help text"
+  echo "  ceph <args>    : call a 'ceph' CLI command with arbitrary args"
+  echo "  status         : show status of the CephCluster and its resources"
+  echo "    -l, --long     : show long status"
+  echo "  operator <subcommand>..."
+  echo "    restart      : restart the Rook-Ceph operator"
+  echo ""
+}
+
+function fail_error () {
+  print_usage >&2
+  echo "ERROR: $*" >&2
+  exit 1
+}
+
+# return failure if the input is not a flag
+function is_flag () {
+  [[ "$1" == -* ]]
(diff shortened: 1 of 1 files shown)


subhamkrai:
> > The approach sounds good to me, great to have this initial effort flushed out for reliable args parsing. So the next step is for Subham to integrate in a new PR, then we close this?
> 
> That is what I would suggest, yes. We can always refer back to this PR as an example for implementing subcommands (possibly with args) later down the line.

okay, I'll start working on this soon

PR #15, opened 2022-01-20
docs: update the Readme doc to add commands

subhamkrai:
Unfortunately, we forgot to add the list of commands
we should have with our plugin. This commit updates
the doc to add commands this plugin supports.
Like, running any ceph command and restart operator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,27 +2,74 @@
 
 # kubectl-rook-ceph
 
-Provide common management and troubleshooting tools for the Rook Ceph storage provider as a Krew plugin.
+Provide common management and troubleshooting tools for the [Rook Ceph](https://github.com/rook/rook) storage provider as a [Krew](https://github.com/kubernetes-sigs/krew) plugin.
 
 ## Install
 
-To install tool
+To install the plugin, run:
 
-1. Run the command below to use it with krew
+  ```kubectl krew install rook-ceph```
 
-    ```kubectl krew install rook-ceph```
+## Usage
 
-Now, use the tool with `kubectl`
+`kubectl rook-ceph <root-args> <command> <command-args>`
+
+### Root args
+
+- `--namespace` | `-n`: the Kubernetes namespace in which the CephCluster resides (default: rook-ceph)
+- `--help` | `-h`: Output help text
+
+### Commands
+
+- `ceph`: Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](https://docs.ceph.com/en/pacific/start/intro/) for more.
+
+- `operator`: run sub-commands against the Rook-Ceph operator
+  -  `restart`: restart the Rook-Ceph operator
+
+- `help`: Output help text
+
+
+## Examples
+
+### Run a Ceph Command
+
+Any `ceph` command can be run with the plugin. This example gets the ceph status:
 
 ```console
-# example: `kubectl rook-ceph` as prefix with `ceph` command
 kubectl rook-ceph ceph status
 ```
 
+>```text
+>  cluster:
+>    id:     a1ac6554-4cc8-4c3b-a8a3-f17f5ec6f529
+>    health: HEALTH_OK
+>
+>  services:
+>    mon: 3 daemons, quorum a,b,c (age 11m)
+>    mgr: a(active, since 10m)
+>    mds: 1/1 daemons up, 1 hot standby
+>    osd: 3 osds: 3 up (since 10m), 3 in (since 8d)
+>
+>  data:
+>    volumes: 1/1 healthy
+>    pools:   6 pools, 137 pgs
+>    objects: 34 objects, 4.1 KiB
+>    usage:   58 MiB used, 59 GiB / 59 GiB avail
+>    pgs:     137 active+clean
+>
+>  io:
+>    client:   1.2 KiB/s rd, 2 op/s rd, 0 op/s wr
+>```
 
-## Commands
+### Restart the Operator
 
-TBD
+```console
+kubectl rook-ceph operator restart
+```
+
+>```text
+>deployment.apps/rook-ceph-operator restarted
+>```
 
 ## Contributing
 


README.md
```diff
@@ -22,7 +22,18 @@ kubectl rook-ceph ceph status
...
+
+       1.
+        ```console
+        # we could also provide optional flag namespace with `-n` or `--namespace` after `kubect rook-ceph` and before main commands.
+        # example: `kubectl rook-ceph` as prefix with `ceph` command
+        kubectl rook-ceph <-n/--namespace=<namespace> ceph status
+        ```
+        2. `kubectl rook-ceph ceph --status --format=json-pretty`
+
+2. Restart rook-ceph operator `kubectl rook-ceph operator restart`.
```
BlaineEXE, line 36: I think it might be easier for users to see at a glance if this is written more like a `man` page where the command is given first. 

Something like this explains overall usage to users first, then calls out each supported command, which is pretty standard for how commands are documented for users. IMO, it's fine to put examples either first or last. I have seen both.

```md
### Overall usage
`kubectl rook-ceph <root-args> <command> <command-args>`

### Root args
- `--namespace` | `-n`: the Kubernetes namespace in which the CephCluster resides (default: rook-ceph)

### Commands
- `ceph`: Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](link) for more.
- `operator`: run sub-commands against the Rook-Ceph operator
  -  `restart`: restart the Rook-Ceph operator

### Examples
- Run a ceph command: `kubectl rook-ceph ceph --status --format=json-pretty`
- Run a command against a CephCluster in a different namespace: `kubectl rook-ceph --namespace my-cluster ceph --status` 
- Restart the operator: `kubectll rook-ceph operator restart`
```

There are different strategies for keeping documentation up-to-date with a CLI tool. It seems beneficial to note for users what commands the tool supports so they can hopefully decide if it's worth using. Maybe we should just copy the usage text here, either manually or with an automated build script.

travisn: That approach looks good to me too. To keep it simple, maybe we just keep it like that with the most common commands listed here, then point out to use --help for more options.

README.md
```diff
@@ -2,7 +2,7 @@
 
 # kubectl-rook-ceph
 
-Provide common management and troubleshooting tools for the Rook Ceph storage provider as a Krew plugin.
+Provide common management and troubleshooting tools for the [Rook Ceph](https://github.com/rook/rook) storage provider as a [Krew](https://github.com/kubernetes-sigs/krew) plugin.
 
 ## Install
```
travisn, line 8: To simplify the install instructions, how about combining lines 9-11 and no bullet point?
```
To install the plugin, run the command below:
```

travisn: Lines 15-20 should be moved to the next section

README.md
```diff
@@ -22,7 +22,43 @@ kubectl rook-ceph ceph status
 
 ## Commands
```
travisn, line 23:
```suggestion
## Usage
```

README.md
```diff
@@ -22,7 +22,43 @@ kubectl rook-ceph ceph status
...
 
-TBD
+1. We support running Ceph command with `rook-ceph` [krew]([Krew](https://github.com/kubernetes-sigs/krew)) plugin
```
travisn, line 25: How about removing lines 25-36? They seem redundant with the other section.

README.md
```diff
@@ -22,7 +22,43 @@ kubectl rook-ceph ceph status
...
+### Commands
+
+- `ceph`: Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](https://docs.ceph.com/en/pacific/start/intro/) for more.
+
+- `operator`: run sub-commands against the Rook-Ceph operator
+  -  `restart`: restart the Rook-Ceph operator
+
+- `help`: Output help text
+
+### Examples
```
travisn, line 58: How about moving the examples section to be first under the Usage section? Then they can see at a glance what commands can be run.

README.md
```diff
@@ -22,7 +22,43 @@ kubectl rook-ceph ceph status
...
+
+- Run a ceph command: `kubectl rook-ceph ceph --status --format=json-pretty`
```
travisn, line 60: Let's keep the examples with ceph commands simple. JSON output isn't that pretty in reality anyway. :) 
```suggestion
- Run a ceph command: `kubectl rook-ceph ceph status 
```

travisn (changes requested):

README.md
subhamkrai: done.

README.md
subhamkrai: removed.

README.md
```diff
@@ -2,15 +2,15 @@
 
 # kubectl-rook-ceph
 
-Provide common management and troubleshooting tools for the Rook Ceph storage provider as a Krew plugin.
+Provide common management and troubleshooting tools for the [Rook Ceph](https://github.com/rook/rook) storage provider as a [Krew](https://github.com/kubernetes-sigs/krew) plugin.
 
 ## Install
 
-To install tool
+To install the plugin, run the command below:
```
travisn, line 9: We might just remove this line completely, or perhaps more concisely:
```suggestion
To install the plugin, run:
```

README.md
```diff
@@ -19,10 +19,30 @@ Now, use the tool with `kubectl`
 kubectl rook-ceph ceph status
 ```
 
+### Examples
```
travisn, line 22: Let's make this a higher level section
```suggestion
## Examples
```

README.md
```diff
@@ -19,10 +19,30 @@ Now, use the tool with `kubectl`
 kubectl rook-ceph ceph status
```
travisn, line 19: Let's move the status under the examples

README.md
```diff
@@ -19,10 +19,30 @@ Now, use the tool with `kubectl`
...
 ```
 
+### Examples
+
+- Run a ceph command: `kubectl rook-ceph ceph`
+- Restart the operator: `kubectll rook-ceph operator restart`
+
+### Overall usage
```
travisn, line 27: How about removing the "Usage" title above and rename this to Usage?
```suggestion
## Usage
```

README.md
```diff
@@ -19,10 +19,30 @@ Now, use the tool with `kubectl`
 kubectl rook-ceph ceph status
 ```
 
+### Examples
+
```
travisn, line 23: What about the following changes for examples?
- Move the "Examples" after the "Usage" section (I know, another reorder :)
- Each example gets its own 3rd-level header and full output included with the command. For example:

## Examples
### Run a Ceph Command
Any `ceph` command can be run with the plugin. This example gets the ceph status:

```
$ kubectl rook-ceph ceph status

<paste full output from a sample cluster>
```

### Restart the Operator
```
$ kubectll rook-ceph operator restart

<paste output>
```
...

travisn (changes requested):

README.md
subhamkrai: updated

travisn: Did you push the changes? I don't see it

subhamkrai: can see it now?

README.md
```diff
@@ -2,15 +2,15 @@
...
 ## Install
 
-To install tool
+To install the plugin, run:
 
-1. Run the command below to use it with krew
+  ```kubectl krew install rook-ceph```
 
-    ```kubectl krew install rook-ceph```
+## Usage
```
travisn, line 13: Let's remove lines 13-20, it seems duplicative.

README.md
```diff
@@ -19,10 +19,68 @@ Now, use the tool with `kubectl`
...
+
+- `ceph`: Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](https://docs.ceph.com/en/pacific/start/intro/) for more.
+
+- `operator`: run sub-commands against the Rook-Ceph operator
+  -  `restart`: restart the Rook-Ceph operator
+
+- `help`: Output help text
+
+
+## Examples
```
travisn, line 41: The new examples section looks good to me!

README.md
```diff
@@ -19,10 +19,68 @@ Now, use the tool with `kubectl`
...
+
+### Run a Ceph Command
+
+Any `ceph` command can be run with the plugin. This example gets the ceph status
```
travisn, line 45:
```suggestion
Any `ceph` command can be run with the plugin. This example gets the ceph status:
```

README.md
```diff
@@ -19,10 +19,68 @@ Now, use the tool with `kubectl`
...
+
+```console
+kubectl rook-ceph ceph status
+```
+
+>```text
+>  cluster:
+>    id:     a1ac6554-4cc8-4c3b-a8a3-f17f5ec6f529
+>    health: HEALTH_WARN
```
travisn, line 54: How about a healthy example? :) 
```suggestion
>    health: HEALTH_OK
```
And you can remove the following two warning lines

travisn (changes requested):

travisn (approved):

PR #18, opened 2022-01-27
core: add more commands to the plugin

subhamkrai:
Adding more to the plugin like rook version, rook status,
get-mon-endpoints, set-mon-endpoints and etc.

Todo:

- [x] `kubectl rook-ceph get-mon-endpoints`
- [ ] `kubectl rook-ceph set-mon-endpoints`: For disaster recovery, update the `rook-ceph-mon-endpoints` configmap and restart the operator
-  [ ] `kubectl rook-ceph force-mon-quorum <mon>`: Automate the disaster recovery guide for forcing the mon quorum to a single mon
- [ ] `kubectl rook-ceph get-volumes [rwx|rwo|cephfs|rbd]`
- [x] `kubectl rook-ceph ceph <args>`: Run any ceph command in the toolbox
- [x] `kubectl rook-ceph operator restart`: Restart the operator
- [x] `kubectl rook-ceph operator set <property> <value>`: Set the property in the `rook-ceph-operator-config` configmap, for example to enable debug logging or configure the CSI driver
- [x] `kubectl rook-ceph rook version`: Print the version of Rook and the version of Ceph daemons from the CephCluster CR status
- [x] `kubectl rook-ceph rook status`: Print the phase and/or conditions of every Ceph CR
- [ ] `kubectl rook-ceph rook purge-osd <id>`: Replace the purge-osd.yaml example job
- [ ] `kubectl rook-ceph set-pool-failure-domain <pool> <domain>`: Replace the advanced topic for updating a pool's failure domain

Closes: #17 
Signed-off-by: subhamkrai <srai@redhat.com>

PR #23, opened 2022-01-31
core: add auto-completion when installed as script

subhamkrai:
This commit adds auto-completion feature when installed
as script and also updating the doc for the same.


subhamkrai:
on top of #18

README.md
```diff
@@ -6,10 +6,23 @@ Provide common management and troubleshooting tools for the [Rook Ceph](https://
 
 ## Install
 
-To install the plugin, run:
+### To install the plugin with krew, run:
 
   ```kubectl krew install rook-ceph```
 
+### To install the plugin as script (has auto-completion), run:
```
travisn, line 13: For now I'd suggest we not document this install method, let's just keep the krew install instructions. While it will be more convenient especially for the completion, the krew install instructions are still the simpler method to get going.

travisn (changes requested):

README.md
BlaineEXE: We could do what kubectl does and have a plugin command that outputs the completion script. E.g., `kubectl rook-ceph completion bash`

subhamkrai:

Issue #24, opened 2022-02-07
No ability to specify namespace


DomantasGiedraitis:
Hey all,

I will start by saying this plugin is a great idea. Good work folks! 

**Issue**
I am unable to specify the namespace. It looks like it defaults to `rook-ceph` and `--namespace <custom-namespace>` doesn't override it.

**Reproduction Steps**
Run `kubectl rook-ceph ceph status --namespace my-custom-namespace`

**Expected Output**
The rook-ceph would check for OSDs in to `my-custom-namespace` namespace and return the value.

**Actual Output**
```
Error from server (NotFound): namespaces "rook-ceph" not found
```

subhamkrai:
@DomantasGiedraitis thanks for opening the issue, I'll look into this

subhamkrai:
@DomantasGiedraitis also, for custom namespace the syntax is 
```console
kubectl rook-ceph  --namespace my-custom-namespace ceph status
```

but in future, we'll remove this limitation.

DomantasGiedraitis:
Nice one! Thanks for a quick response. I will give it a go 👍

PR #26, opened 2022-02-09
ceph: add support to read custom namespace

subhamkrai:
Earlier, we have the support to read custom namespace
but there was bug that it was reading from wrong config
file. This commit fixes that bug.

Closes: https://github.com/rook/kubectl-rook-ceph/issues/24

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -161,7 +161,7 @@ function run_help_command () {
 
 function run_ceph_command () {
   # do not call end_of_command_parsing here because all remaining input is passed directly to 'ceph'
-  kubectl --namespace "$NAMESPACE" exec deploy/rook-ceph-operator -- ceph "$@" --conf=/var/lib/rook/rook-ceph/rook-ceph.config
+  kubectl --namespace "$NAMESPACE" exec deploy/rook-ceph-operator -- ceph "$@" --conf="$CEPH_CONF_PATH"
 }
 
 ####################################################################################################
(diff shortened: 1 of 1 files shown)


kubectl-rook-ceph.sh
```diff
@@ -161,7 +161,7 @@ function run_help_command () {
 
 function run_ceph_command () {
   # do not call end_of_command_parsing here because all remaining input is passed directly to 'ceph'
-  kubectl --namespace "$NAMESPACE" exec deploy/rook-ceph-operator -- ceph "$@" --conf=/var/lib/rook/rook-ceph/rook-ceph.config
+  kubectl --namespace "$NAMESPACE" exec deploy/rook-ceph-operator -- ceph "$@" --conf=/var/lib/rook/"$NAMESPACE"/"$NAMESPACE".config
```
leseb, line 164: Should we have a single variable for the path of the ceph conf instead? Like a top value:

`CEPH_CONF_PATH=/var/lib/rook/"$NAMESPACE"/"$NAMESPACE".config`

So future reference can be changed in one place?

subhamkrai: make sense, changed

leseb (approved):

travisn (approved):

travisn:
As a separate PR it would be nice to run the CI in a different namespace than the default.

subhamkrai:
> As a separate PR it would be nice to run the CI in a different namespace than the default, or potentially one action in the default namespace and another in a different namespace.

will add that.

subhamkrai:
> > As a separate PR it would be nice to run the CI in a different namespace than the default, or potentially one action in the default namespace and another in a different namespace.
> 
> will add that.

#28

Issue #25, opened 2022-02-07
Support operator and cluster in different namespaces


travisn:
Currently the plugin assumes that the operator and cluster are in the same namespace. While this is most often the case, it is also possible to run multiple clusters each in their own namespace, while the operator is in a separate namespace. So the plugin should allow setting the namespace of the operator to be different from the cluster namespace.

Consider the following design:
- `--namespace`: If specified, is the namespace for the cluster. If not specified, look for the env var `$ROOK_CLUSTER_NAMESPACE`. If the env var is not found, default to `rook-ceph`.
- `--operator-namespace`: If specified, is the operator namespace. If not specified, look for the env var `$ROOK_OPERATOR_NAMESPACE`. If the env var is not found, default to the same namespace as the cluster described in the previous bullet.

PR #27, opened 2022-02-09
core: support operator and cluster in different namespaces

subhamkrai:
currently, the tool can't handle different namespace for operator and
cluster.

With this path, we'll have two namespaces
1. `--namespace` | `-n`: namespace of CephCluster (default: rook-ceph):
   it will first read from CLI arg, if not passed then it will read from
   env var `ROOK_CLUSTER_NAMESPACE` if that's empty  default to `rook-ceph`.

2. `--operator-namespace` | `-op-ns`: namespace of rook operator
   (default: <namespace> i.e rook-ceph). it will first read from cli arg,
   if not passed then it will read from env var `ROOK_OPERATOR_NAMESPACE`
   if that's empty default to <namespace>(the one in first step) i.e
   `rook-ceph`.

Closes: https://github.com/rook/kubectl-rook-ceph/issues/25

### DIFF
--- a/.github/workflows/ci-for-diff-ns.yaml
+++ b/.github/workflows/ci-for-diff-ns.yaml
@@ -1,0 +1,50 @@
+name: Plugin test
+on:
+  pull_request:
+
+defaults:
+  run:
+    # reference: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell
+    shell: bash --noprofile --norc -eo pipefail -x {0}
+
+jobs:
+  # This test is required to test latest changes or the changes that not present
+  # with current version of rook-ceph krew plugin
+  with-pr-changes-in-custom-namespace:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: setup cluster
+        uses: ./.github/workflows/cluster-setup
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          op-ns: "test-operator"
+          cluster-ns: "test-cluster"
+
+      - name: install script
+        run: sudo install kubectl-rook-ceph.sh /usr/local/bin/kubectl-rook_ceph
+
+      - name: Test Plugin
+        run: |
+          kubectl rook_ceph -o test-operator -n test-cluster ceph status
+          kubectl rook_ceph -o test-operator -n test-cluster ceph status -f json
+          kubectl rook_ceph -o test-operator -n test-cluster ceph status --format json-pretty
+          kubectl rook_ceph -o test-operator -n test-cluster rbd ls replicapool
+          kubectl rook_ceph -o test-operator -n test-cluster operator restart
+
+          # let's wait for operator pod to be restart
+          tests/github-action-helper.sh wait_for_operator_pod_to_be_ready_state_custom
+          kubectl rook_ceph -o test-operator -n test-cluster operator set ROOK_LOG_LEVEL DEBUG
+          kubectl rook_ceph -o test-operator -n test-cluster mons
+          kubectl rook_ceph -o test-operator -n test-cluster rook version
+          kubectl rook_ceph -o test-operator -n test-cluster rook status
+          kubectl rook_ceph -o test-operator -n test-cluster rook status all
+          kubectl rook_ceph -o test-operator -n test-cluster rook status cephobjectstores
+
+      - name: setup tmate session for debugging when event is PR
+        if: failure() && github.event_name == 'pull_request'
+        uses: mxschmitt/action-tmate@v3
--- a/.github/workflows/cluster-setup/action.yaml
+++ b/.github/workflows/cluster-setup/action.yaml
@@ -4,6 +4,12 @@ inputs:
   github-token:
     description: GITHUB_TOKEN from the calling workflow
     required: true
+  op-ns:
+    description: operator namespace where rook operator will deploy
+    required: false
+  cluster-ns:
+    description: cluster namespace where ceph cluster will deploy
+    required: false
 
 runs:
   using: "composite"
@@ -11,14 +17,14 @@ runs:
     - name: setup minikube
       uses: manusa/actions-setup-minikube@v2.4.2
       with:
-        minikube version: 'v1.23.2'
-        kubernetes version: 'v1.22.2'
+        minikube version: "v1.23.2"
+        kubernetes version: "v1.22.2"
         start args: --memory 6g --cpus=2
         github token: ${{ inputs.github-token }}
 
     - name: print k8s cluster status
       shell: bash --noprofile --norc -eo pipefail -x {0}
-      run:  |
+      run: |
         minikube status
         kubectl get nodes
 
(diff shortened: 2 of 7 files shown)


subhamkrai:
need testing

kubectl-rook-ceph.sh
```diff
@@ -27,21 +27,23 @@ function print_usage () {
   echo "USAGE"
   echo "  kubectl rook-ceph <main args> <command> <command args>"
   echo "MAIN ARGS"
-  echo "  -h, --help                  : output help text"
-  echo "  -n, --namespace='rook-ceph' : the namespace of the CephCluster"
+  echo "  -h, --help                                : output help text"
+  echo "  -n, --namespace='rook-ceph'               : the namespace of the CephCluster"
+  echo "  -op-ns, --operator-namespace='rook-ceph'  : the namespace of the CephCluster"
```
travisn, line 32: How about we keep the short names to one (or maybe two) letters?
```suggestion
  echo "  -o, --operator-namespace='rook-ceph'  : the namespace of the rook operator"
```

kubectl-rook-ceph.sh
```diff
@@ -27,21 +27,23 @@ function print_usage () {
   echo "USAGE"
   echo "  kubectl rook-ceph <main args> <command> <command args>"
   echo "MAIN ARGS"
-  echo "  -h, --help                  : output help text"
-  echo "  -n, --namespace='rook-ceph' : the namespace of the CephCluster"
+  echo "  -h, --help                                : output help text"
```
travisn, line 30:
```suggestion
  echo "  -h, --help                                : output help"
```

kubectl-rook-ceph.sh
```diff
@@ -27,21 +27,23 @@ function print_usage () {
...
+  echo "  -n, --namespace='rook-ceph'               : the namespace of the CephCluster"
+  echo "  -op-ns, --operator-namespace='rook-ceph'  : the namespace of the CephCluster"
+
   echo "COMMANDS"
-  echo "  help                        : output help text"
-  echo "  ceph <args>                 : call a 'ceph' CLI command with arbitrary args"
-  echo " rbd <args>                   : call a 'rbd' CLI command with arbitrary args"
+  echo "  help                                      : output help text"
```
travisn, line 35: Since `--help` is above, seems like we don't need to document it here as well, even if it's an option.
```suggestion
```

kubectl-rook-ceph.sh
```diff
@@ -328,8 +330,18 @@ function run_main_command () {
 # MAIN: PARSE MAIN ARGS AND CALL MAIN COMMAND HANDLER
 ####################################################################################################
 
-# Default values for flag-controlled settings.
-NAMESPACE='rook-ceph' # namespace of the cluster
+# Default values
+NAMESPACE=$(printenv ROOK_CLUSTER_NAMESPACE ) # namespace of the cluster
+OPERATOR_NAMESPACE=$(printenv ROOK_OPERATOR_NAMESPACE ) # operator namespace
+CEPH_CONF_PATH="/var/lib/rook/$NAMESPACE/$NAMESPACE.config" # path of ceph config
```
travisn, line 336: Need to move this down a few lines until after the namespace vars are finally determined?

travisn: Seems like lines 336-344 should move down around 377 so these defaults will be set only after the command line args are evaluated?

kubectl-rook-ceph.sh
```diff
@@ -339,12 +351,17 @@ function parse_main_flag () {
     "-n"|"--namespace")
       val_exists "$val" || return 1 # val should exist
       NAMESPACE="${val}"
+      CEPH_CONF_PATH="/var/lib/rook/$NAMESPACE/$NAMESPACE.config"
```
travisn, line 354: I believe this isn't necessary if you move the other lines move as suggested above.

travisn (changes requested):

kubectl-rook-ceph.sh
subhamkrai: removed

kubectl-rook-ceph.sh
```diff
@@ -360,4 +348,17 @@ if [[ "${#REMAINING_ARGS[@]}" -eq 0 ]]; then
   fail_error "No command to run"
 fi
 
+# Default values
+NAMESPACE=$(printenv ROOK_CLUSTER_NAMESPACE ) # namespace of the cluster
```
travisn, line 352: Lines 352-353 need to be before line 345, otherwise the env vars would overwrite the parse_flags, right?

subhamkrai:
@travisn when I'm trying to read from env var let's say `ROOK_CLUSTER_NAMESPACE` and suppose it is not set then I'll get `unbound variable` error.

leseb:
> @travisn when I'm trying to read from env var let's say `ROOK_CLUSTER_NAMESPACE` and suppose it is not set then I'll get `unbound variable` error.

You can use:

```sh
```

subhamkrai:
I'll add ci for this in different PR to keep PR small and also some discussion is going on one of CI pr #28

leseb:
> I'll add ci for this in different PR to keep PR small and also some discussion is going on one of CI pr #28

If we don't add the CI, how do we know that this is working?

subhamkrai:
> > I'll add ci for this in different PR to keep PR small and also some discussion is going on one of CI pr #28
> 
> If we don't add the CI, how do we know that this is working?

okay, adding CI here and we can discuss on #28 if want some close that pr or add some new test.

travisn (approved):

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          custom-ns: "true"
+
+      - name: install script
+        run: sudo install kubectl-rook-ceph.sh /usr/local/bin/kubectl-rook_ceph
+
+      - name: Test Plugin
+        run: |
+          kubectl rook_ceph -o test-operator -n test-cluster  ceph status
```
leseb, line 32:
```suggestion
          kubectl rook_ceph -o test-operator -n test-cluster ceph status
```

same for the lines below

.github/workflows/cluster-setup/action.yaml
```diff
@@ -4,21 +4,24 @@ inputs:
   github-token:
     description: GITHUB_TOKEN from the calling workflow
     required: true
+  custom-ns:
```
leseb, line 7: why a bool and not a string so that we can choose the name of the namespace?

tests/github-action-helper.sh
```diff
@@ -33,12 +33,35 @@ deploy_rook() {
...
   kubectl create -f cluster-test.yaml
+  wait_for_pod_to_be_ready_state_default
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/storageclass-test.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/pvc.yaml
 }
 
-# wait_for_pod_to_be_ready_state check for operator pod to in ready state
-wait_for_pod_to_be_ready_state() {
+deploy_rook_in_custom_namespace() {
+  kubectl create namespace test-operator # creating namespace manually beacuse rook common.yaml create one namespace and here we need 2
```
leseb, line 42:
```suggestion
  kubectl create namespace test-operator # creating namespace manually because rook common.yaml create one namespace and here we need 2
```

.github/workflows/cluster-setup/action.yaml
subhamkrai: actually, it is a string only, ` custom-ns: "true"` this is just passing a var if we want to deploy on custom namespace or not.

subhamkrai: okay, We have 2 namespaces so I'll create another input in action and will read both namespaces. I'll do the changes

subhamkrai: ready for review

leseb (approved):

PR #28, opened 2022-02-10
ci: test plugin when rook deployed in custom namespace

subhamkrai:
this commit adds ci to test plugin when rook deployed
in custom namespace.

### DIFF
--- a/tests/github-action-helper.sh
+++ b/tests/github-action-helper.sh
@@ -33,12 +33,35 @@ deploy_rook() {
   curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/cluster-test.yaml -o cluster-test.yaml
   sed -i "s|#deviceFilter:|deviceFilter: $(lsblk|awk '/14G/ {print $1}'| head -1)|g" cluster-test.yaml
   kubectl create -f cluster-test.yaml
+  wait_for_pod_to_be_ready_state_default
+  kubectl -n rook-ceph wait --for=condition=ready pod -l app=rook-ceph-osd --timeout=200s
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/storageclass-test.yaml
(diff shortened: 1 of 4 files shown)


tests/github-action-helper.sh
```diff
@@ -33,12 +33,35 @@ deploy_rook() {
...
+  wait_for_pod_to_be_ready_state_custom
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/storageclass-test.yaml -o storageclass-test.yaml
+  sed -i "s|provisioner: rook-ceph.rbd.csi.ceph.com |provisioner: rook.rbd.csi.ceph.com |g" storageclass-test.yaml
+  deploy_with_custom_ns storageclass-test.yaml
+  kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/pvc.yaml
+}
+
+deploy_with_custom_ns(){
+  sed -i "s|rook-ceph # namespace:operator|rook # namespace:operator|g" "$1"
+  sed -i "s|rook-ceph # namespace:cluster|rook # namespace:cluster|g" "$1"
```
travisn, line 60: How about different namespaces for the operator and cluster, and name them something like `test-cluster` and `test-operator`?

travisn (changes requested):

tests/github-action-helper.sh
subhamkrai: I think the aim of the PR is to test in different namespaces (from `rook-ceph`) and having different namespaces for operator and cluster PR is still in progress. So, we can wait for the other PR to get merged and I'll update this pr accordingly.

Or we can go ahead with this PR have a new test for different namespaces for the operator and cluster?

subhamkrai: I think now these changes are included in #27 only. should we close this?

subhamkrai:
Closing since these changes are included in other PR. Thanks!

Issue #21, opened 2022-01-31
Add command to purge osd 


subhamkrai:
Add command for:

- [ ] `kubectl rook-ceph rook purge-osd <id>`: Replace the purge-osd.yaml example job

subhamkrai:
@travisn since we have cr for this, do we want to add it to the tool?

travisn:
The purge-osd job example basically just runs a command that could be run in the operator pod. So I think this would actually be a nice addition to the krew plugin.

PR #33, opened 2022-05-26
core: add new rook subcommand `purge-osd <id>`

subhamkrai:
this commit adding new rook subcommand `purge-osd <id>`
This command will removing osd force fully when osd is down and
will also delete the deployment of the osd <id>.

the command will looks like this:
`kubectl rook-ceph rook purge-osd <id>`

Closes: https://github.com/rook/kubectl-rook-ceph/issues/21


kubectl-rook-ceph.sh
```diff
@@ -246,6 +250,15 @@ function run_rook_cr_status() {
   fi
 }
 
+function run_purge_osd() {
+   if [ "$#" -eq 1 ]; then
+    run_ceph_command osd purge osd."$1" --yes-i-really-mean-it --force
```
travisn, line 255: Instead of running the ceph purge command, let's just run the `rook ceph osd remove` command, the same as the osd purge job. Then it will delete the osd deployment and all other related resources such as PVCs.

README.md
```diff
@@ -38,6 +38,7 @@ To install the plugin, run:
   - `status`      : Print the phase and/or conditions of every CR in the namespace"
   - `status all`  : Print the phase and conditions of all CRs
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
+  - `purge-osd <osd-id>` : remove the osd from crush map, osd map and removes it keys given osd is down. **Note** this is destroy operation"
```
travisn, line 41:
```suggestion
  - `purge-osd <osd-id>` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
```

travisn (changes requested):

subhamkrai:
local testing 
```kubectl-rook_ceph rook purge-osd 0
2022-06-08 09:27:33.847621 I | rookcmd: starting Rook v1.9.0-alpha.0.322.g1acc0621a with arguments 'rook ceph osd remove --osd-ids=0 --force-osd-removal=true'
2022-06-08 09:27:33.847663 I | rookcmd: flag values: --force-osd-removal=true, --help=false, --log-level=DEBUG, --operator-image=, --osd-ids=0, --preserve-pvc=false, --service-account=
2022-06-08 09:27:33.847668 I | op-mon: parsing mon endpoints: a=10.111.249.156:6789
...
...
2022-06-08 09:27:39.837299 I | cephosd: no ceph crash to silence
2022-06-08 09:27:39.837323 I | cephosd: completed removal of OSD 0

File: kubectl-rook-ceph.sh (hunk 1 above)
Comment (subhamkrai) [reply kubectl-rook-ceph.sh]: good suggestion. 

changes done.

**Comment by subhamkrai** [comment]
on top of #34

File: .github/workflows/ci-for-default-ns.yaml [hunk 3]
```diff
@@ -67,7 +66,11 @@ jobs:
           kubectl rook_ceph rook status
           kubectl rook_ceph rook status all
           kubectl rook_ceph rook status cephobjectstores
+          kubectl rook_ceph rbd ls replicapool
 
+          # for testing osd purge scale the osd deplyment
+          kubectl --namespace rook-ceph scale deploy/rook-ceph-osd-0 --replicas=0
+          kubectl-rook_ceph rook purge-osd 0
```
travisn, line 73: You may need to sleep a few seconds between scaling down the osd and purging the osd, otherwise the purge will be refused since the osd is not down yet.

subhamkrai:
> LGTM except the CI seems to be failing.
> 
> After this is merged, will be good to get a release out, then update the Rook [OSD Management topic](https://github.com/rook/rook/blob/master/Documentation/Storage-Configuration/Advanced/ceph-osd-mgmt.md#remove-an-osd) with this tool.

CI error is strange,  I checked in runner command is running. Adding some sleep to check if there is race conditions
```runner@fv-az426-571:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ kubectl rook_ceph rbd ls replicapool
csi-vol-f5c75fc3-ec7b-11ec-ae40-0242ac110005```

**Comment by subhamkrai** [comment]
> > LGTM except the CI seems to be failing.
> > After this is merged, will be good to get a release out, then update the Rook [OSD Management topic](https://github.com/rook/rook/blob/master/Documentation/Storage-Configuration/Advanced/ceph-osd-mgmt.md#remove-an-osd) with this tool.
> 
> CI error is strange, I checked in runner command is running. Adding some sleep to check if there is race conditions
> 
> ```
> csi-vol-f5c75fc3-ec7b-11ec-ae40-0242ac110005```
> ```

adding sleep is working, so maybe it was early for the command to run.

File: kubectl-rook-ceph.sh [hunk 4]
```diff
@@ -245,6 +249,21 @@ function run_rook_cr_status() {
   fi
 }
 
+function run_purge_osd() {
+  if [ "$#" -eq 1 ]; then
+    mon_endpoints=$($TOP_LEVEL_COMMAND --namespace "$ROOK_CLUSTER_NAMESPACE" get cm rook-ceph-mon-endpoints -o jsonpath='{.data.data}' | cut -d "," -f1)
+    ceph_secret=$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- cat /var/lib/rook/"$ROOK_CLUSTER_NAMESPACE"/client.admin.keyring | grep "key" | awk '{print $3}')
+    $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- sh -c "export ROOK_MON_ENDPOINTS=$mon_endpoints \
+      ROOK_CEPH_USERNAME=client.admin \
```
BlaineEXE, line 257: Will this be affected by our current desires to change the client? Should we change something here for forwards compatibility?

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -67,7 +66,15 @@ jobs:
           kubectl rook_ceph rook status
           kubectl rook_ceph rook status all
           kubectl rook_ceph rook status cephobjectstores
+          # use of sleep is not incouraged but here we need sleep between scaling down the osd and purging the osd
```
travisn, line 69: let's just explain why we're sleeping, no need for explaining why we wish we didn't have to ;) 
```suggestion
to allow time for reconcile, sleep before listing the pools
```

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -67,7 +66,15 @@ jobs:
...
+          sleep 5
+          kubectl rook_ceph rbd ls replicapool
 
+          # for testing osd purge scale the osd deplyment
+          kubectl --namespace rook-ceph scale deploy/rook-ceph-osd-0 --replicas=0
+          # use of sleep is not incouraged but here we need sleep between scaling down the osd and purging the osd
```
travisn, line 75:
```suggestion
          # we need to sleep so the osd will be marked down before purging the osd
```

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -44,6 +43,15 @@ jobs:
           kubectl rook_ceph -o test-operator -n test-cluster rook status
           kubectl rook_ceph -o test-operator -n test-cluster rook status all
           kubectl rook_ceph -o test-operator -n test-cluster rook status cephobjectstores
+          # use of sleep is not incouraged but here we need sleep between scaling down the osd and purging the osd
```
travisn, line 46: In a future PR, can we refactor the script and take namespace as a var so we don't duplicate it?

README.md
```diff
@@ -38,6 +38,7 @@ To install the plugin, run:
   - `status`      : Print the phase and/or conditions of every CR in the namespace"
   - `status all`  : Print the phase and conditions of all CRs
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
+  - `purge-osd <osd-id>` :Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
```
travisn, line 41:
```suggestion
  - `purge-osd <osd-id>` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
```

kubectl-rook-ceph.sh
```diff
@@ -245,6 +249,21 @@ function run_rook_cr_status() {
...
+      ROOK_CEPH_SECRET=$ceph_secret \
+      ROOK_LOG_LEVEL=DEBUG \
```
travisn, line 259: Is debug level needed? Or we can remove this line?

travisn (changes requested):

kubectl-rook-ceph.sh
subhamkrai: I believe not, in the future we change the client we'll just have to change the client name here.

kubectl-rook-ceph.sh
subhamkrai: no, it is not required, removed

.github/workflows/ci-for-diff-ns.yaml
subhamkrai: yes, I'll refactor the ci yaml's

kubectl-rook-ceph.sh
travisn: For backward compatibility, we will need to keep the client.admin, so I don't see a problem if we keep this for now. Ideally, perhaps we even create a `client.krewplugin` user in the future after that PR, but we can decide later if it's worth it.

kubectl-rook-ceph.sh
```diff
@@ -245,6 +249,20 @@ function run_rook_cr_status() {
...
 
+function run_purge_osd() {
+  if [ "$#" -eq 1 ]; then
+    mon_endpoints=$($TOP_LEVEL_COMMAND --namespace "$ROOK_CLUSTER_NAMESPACE" get cm rook-ceph-mon-endpoints -o jsonpath='{.data.data}' | cut -d "," -f1)
+    ceph_secret=$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- cat /var/lib/rook/"$ROOK_CLUSTER_NAMESPACE"/client.admin.keyring | grep "key" | awk '{print $3}')
+    $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- sh -c "export ROOK_MON_ENDPOINTS=$mon_endpoints \
+      ROOK_CEPH_USERNAME=client.admin \
+      ROOK_CEPH_SECRET=$ceph_secret \
+      ROOK_CONFIG_DIR=/var/lib/rook && \
+    rook ceph osd remove --osd-ids=$1 --force-osd-removal=true"
```
travisn, line 260: How about we not pass the `--force-osd-removal=true` every time? To avoid accidental osd removal, we should have a separate option for forcing removal, something like: `purge-osd <osd-id> [--force-osd-removal=true]`

travisn (changes requested):
one more suggestion...

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -44,6 +43,15 @@ jobs:
...
           kubectl rook_ceph -o test-operator -n test-cluster rook status cephobjectstores
+          # to allow time for reconcile, sleep before listing the pools
+          sleep 5
+          kubectl rook_ceph -o test-operator -n test-cluster rbd ls replicapool
+
+          # for testing osd purge scale the osd deplyment
+          kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
+          # we need to sleep so the osd will be marked down before purging the osd
+          sleep 5
+          kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
```
subhamkrai, line 54: one of the ci will test `--force` and the other will be running without `--force` so that we test both scenarios.

subhamkrai: passing `--force` in both the ci test since the osd is not out, it will not able to purge the osd, and the operator will keep waiting for osd to be out and CI will keep on running forever

travisn: yeah, with the small cluster we will need to force

kubectl-rook-ceph.sh
```diff
@@ -245,6 +249,28 @@ function run_rook_cr_status() {
   fi
 }
 
+function run_purge_osd() {
+  if [ "$#" -eq 2 ] && [ "$2" = "--force" ]; then
```
travisn, line 253: To avoid the command duplication, how about setting a var for the force flag? For example:
```suggestion
  if [ "$#" -eq 2 ] && [ "$2" = "--force" ]; then
    force_removal=true
  else
    force_removal=false
  fi
```
Then there is a single command that uses the var

README.md
```diff
@@ -38,6 +38,7 @@ To install the plugin, run:
   - `status`      : Print the phase and/or conditions of every CR in the namespace"
   - `status all`  : Print the phase and conditions of all CRs
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
+  - `purge-osd <osd-id> <--force>` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
```
travisn, line 41: Optional flags are generally indicated by different brackets
```suggestion
  - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs. The force flag must be set if the OSD still contains data."
```

travisn (changes requested):

kubectl-rook-ceph.sh
subhamkrai: updated

kubectl-rook-ceph.sh
```diff
@@ -245,6 +249,20 @@ function run_rook_cr_status() {
   fi
 }
 
+function run_purge_osd() {
+  force_removal=false
+  if [ "$#" -eq 2 ] && [ "$2" = "--force" ]; then
```
travisn, line 254: In the future when we have more complex commands it would be nice to allow arguments in a different order, but this is ok for now.

travisn (approved):

kubectl-rook-ceph.sh
```diff
@@ -245,6 +249,20 @@ function run_rook_cr_status() {
...
+    force_removal=true
+  fi
+  mon_endpoints=$($TOP_LEVEL_COMMAND --namespace "$ROOK_CLUSTER_NAMESPACE" get cm rook-ceph-mon-endpoints -o jsonpath='{.data.data}' | cut -d "," -f1)
+  ceph_secret=$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- cat /var/lib/rook/"$ROOK_CLUSTER_NAMESPACE"/client.admin.keyring | grep "key" | awk '{print $3}')
+  $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- sh -c "export ROOK_MON_ENDPOINTS=$mon_endpoints \
+      ROOK_CEPH_USERNAME=client.admin \
```
BlaineEXE, line 260: Of note, the `rook-ceph-mon` secret has the name of the client used to access the cluster. Why don't we read that value and use it. Then this should be true now and in the future also.

kubectl-rook-ceph.sh
BlaineEXE: Good point. Why aren't we using the arg parsing function here?

kubectl-rook-ceph.sh
BlaineEXE: I also wonder if it wouldn't be better to start a version of the osd removal job, then tail the logs of that, instead of running the command in the operator. @travisn wdyt?

travisn: > Of note, the `rook-ceph-mon` secret has the name of the client used to access the cluster. Why don't we read that value and use it. Then this should be true now and in the future also. It would also better support running actions on external clusters.

In the future there will basically be multiple users in the secret instead of just a single user. We will still have to choose which user, so not sure there's an advantage of reading it.

> I also wonder if it wouldn't be better to start a version of the osd removal job, then tail the logs of that, instead of running the command in the operator since it seems like that is causing us to have to do some pretty deep inspection of resources. @travisn wdyt?

The osd removal job is just running a single ceph command, which is also what we're doing with the plugin in the operator pod. Since the pattern of the plugin is to run in the operator pod, I don't see a benefit to changing that pattern. We would just change the complexity of reading keys for the complexity of starting a job and tailing its logs, right?

kubectl-rook-ceph.sh
subhamkrai: IIUC, the `parse_flags` method is written, only accept flags starting before command eg `kuecttl rook-ceph --namesapce ...` and here we are passing `--force` in the end of the command.
But I do agree it is not the best code and also I found some minor bugs. I'm not planning to fix the minor bugs in the pr.  We can discuss about `parge_flags`

kubectl-rook-ceph.sh
```diff
@@ -44,6 +44,7 @@ function print_usage() {
   echo "    status                                  : print the phase and conditions of the CephCluster CR"
   echo "    status all                              : print the phase and conditions of all CRs"
   echo "    status <CR>                             : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
+  echo "    purge-osd <osd-id> [--force]            : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
```
BlaineEXE, line 47: Should this really be a subcommand of `rook`? What if we had this be a top-level command? @travisn thoughts?

kubectl-rook-ceph.sh
BlaineEXE: The currently commented-out status subcommand has flags after a command. See `parse_status_flag`.

PR #34, opened 2022-06-08
core: format script using `shell-format` in vscode

subhamkrai:
this commit format all the shell the script
according the `shell-format` extention in vscode.

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -20,7 +20,7 @@ set -eEuo pipefail
 # HELPER FUNCTIONS
 ####################################################################################################
 
-function print_usage () {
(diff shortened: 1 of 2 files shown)


subhamkrai:
@travisn `rbd ls ...` command was also failing in #34 sometime but seems like it requires sleep. is it okay to merge this will fix in #34 only?

Also, don't see this pr changing  `rbd ls...` part.

travisn:
Did you mean you'll fix it in #33? Sounds good to merge this first and fix the CI separately.

subhamkrai:
> Did you mean you'll fix it in #33? 
yes😅
Sounds good to merge this first and fix the CI separately.

Issue #35, opened 2022-07-12
Add command to put an OSD or Mon pod in debug mode


travisn:
OSD and Mon pods in some support cases may need to be put in a mode that will allow commands to be run against the OSD or Mon while the daemon is not running. Let's call this "debug" mode. This is not a normal part of a reconcile, so the krew plugin is the right place for the one-time operation that can be tuned according to need. 

When the admin decides to put an OSD or Mon in debug mode, the sequence of events would be:
1. `kubectl rook-ceph create-debug [--osd-id <id> | mon-id <id>] [--image <alternate-image>]`
2. Perform debug operations in the debug osd pod
3. `kubectl rook-ceph remove-debug [--osd-id <id> | mon-id <id>]`

# create-debug
The command to create the debug pod would make the following modifications to the copy of the deployment:
1. Verifies the deployment exists
1. Scales the original deployment down to 0
1. Copies the deployment spec and make modifications (see below)
1. Creates the debug deployment 

The following modifications are made in the debug deployment:
- Add the suffix "-debug" to the deployment name
- Add a label `ceph.rook.io/debug: "true"` to the deployment so it can be easily detected. 
- Modify the command in the main container to be only a placeholder so the admin can connect to the pod and execute ceph tool commands

# remove-debug

When done debugging, the remove-debug command:
1. Deletes the debug pod
1. Scales back up the original daemon deployment

parth-gr:
[Randy Martinez](https://github.com/randymtz) if you can explain your finding here,

Currently, as per the above details for debugging https://github.com/rook/kubectl-rook-ceph/issues/35#issue-1302623104, I created something  https://github.com/rook/kubectl-rook-ceph/pull/36, please have a look

randymtz:
> Scales the original deployment down to 0

This will need rook PR#10585 merged before we can do this cleanly. Otherwise, if the rook operator is restarted, osd deployment spec will reconcile & potentially interrupt in-progress maintenance commands.

> Copies the deployment spec and make modifications (see below)
Creates the debug deployment
The following modifications are made in the debug deployment:
Add the suffix "-debug" to the deployment name

This is part of `oc debug`. Note: downstream version >=4.10 have a new 'startup probe' we need to ignore. If alike flags are not part of `kubectl debug`, we'll need to catch when the env is ODF; which will be beneficial at the beginning where current default for namespace is defined. 

> Add a label ceph.rook.io/debug: "true" to the deployment so it can be easily detected.
[...]

PR #36, opened 2022-07-14
core: add debug mode for mon and osd

parth-gr:
Debug mode will offer a special case
where the operator will don't reconcile
on the specific deployments but works
same with other controllers and deployments

Closes:	https://github.com/rook/kubectl-rook-ceph/issues/35

### DIFF
--- a/.github/workflows/ci-for-default-ns.yaml
+++ b/.github/workflows/ci-for-default-ns.yaml
@@ -69,15 +69,29 @@ jobs:
           kubectl rook_ceph rook status all
           kubectl rook_ceph rook status cephobjectstores
           # to allow time for reconcile, sleep before listing the pools
-          sleep 5
+          sleep 10
           kubectl rook_ceph rbd ls replicapool
-
+          
+          # for testing start-debug and stop-debug deployment
+          # mons 
+          kubectl rook_ceph debug start rook-ceph-mon-a
+          sleep 5
+          kubectl rook_ceph debug stop rook-ceph-mon-a
+          # osd  
+          kubectl rook_ceph debug start rook-ceph-osd-0
+          sleep 5
+          kubectl rook_ceph debug stop rook-ceph-osd-0
+          
+          # sleep again so OSD can start
+          sleep 5
+          
           # for testing osd purge scale the osd deplyment
           kubectl --namespace rook-ceph scale deploy/rook-ceph-osd-0 --replicas=0
           # we need to sleep so the osd will be marked down before purging the osd
           sleep 5
-          kubectl-rook_ceph rook purge-osd 0 --force
+          kubectl rook_ceph rook purge-osd 0 --force
           kubectl rook_ceph health
+      
       - name: setup tmate session for debugging when event is PR
         if: failure() && github.event_name == 'pull_request'
         uses: mxschmitt/action-tmate@v3
--- a/.github/workflows/ci-for-diff-ns.yaml
+++ b/.github/workflows/ci-for-diff-ns.yaml
@@ -56,6 +56,18 @@ jobs:
           kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
           kubectl rook_ceph -o test-operator -n test-cluster health
 
+          # for testing start-debug and stop-debug deployment
+          # mon
+          kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-mon-a
+          # sleep till the debug pod get running
+          sleep 5
+          kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-mon-a
+          # osd
+          kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-osd-0
+          # sleep till the debug pod get running
+          sleep 5
+          kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
+
       - name: setup tmate session for debugging when event is PR
         if: failure() && github.event_name == 'pull_request'
         uses: mxschmitt/action-tmate@v3
--- a/README.md
+++ b/README.md
@@ -42,6 +42,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
 
+- `debug` : [Debug a deployment](#debug-mode)  by scaling it down and creating a debug copy. This is supported for mons and OSDs only
+  - `start  <deployment-name> ` 
+    `[--alternate-image <alternate-image>]` : Start debugging a deployment with an optional alternative ceph container image
+  - `stop  <deployment-name>` : Stop debugging a deployment
+
 - `help` : Output help text
 
 ## Examples
@@ -119,6 +124,52 @@ kubectl rook-ceph ceph versions
         "ceph version 16.2.7 (dd0603118f56ab514f133c8d2e3adfc983942503) pacific (stable)": 3
     }
 }
+```
+
+### Debug Mode
+
+Debug mode can be useful when a mon or OSD needs advanced maintenance operations that require the daemon to be stopped. Ceph tools such as `ceph-objectstore-tool`,`ceph-bluestore-tool`, or `ceph-monstore-tool` are commonly used in these scenarios. Debug mode will set up the mon or OSD so that these commands can be run.
+
+Debug mode will automate the following:
+1. Scale down the existing mon or OSD deployment
+2. Start a new debug deployment where operations can be performed directly against the mon or OSD without that daemon running
+   
+   a. The main container sleeps so you can connect and run the ceph commands
+   
+   b. Liveness and startup probes are removed
+
+   c. If alternate Image is passed by --alternate-image flag then the new debug deployment container will be using alternate Image.
+
+For example, start the debug pod for mon `b`:
+```console
+kubectl rook-ceph debug start rook-ceph-mon-b
+```
+```text
+setting debug mode for "rook-ceph-mon-b"
+setting debug command to main container
+deployment.apps/rook-ceph-mon-b scaled
+deployment.apps/rook-ceph-mon-b-debug created
+```
+
+Now connect to the daemon pod and perform operations:
+```console
+kubectl exec <debug-pod> -- <ceph command>
+```
+
+When finished, stop debug mode and restore the original daemon:
+```console
+kubectl rook-ceph debug stop rook-ceph-mon-b
+```
+```text
+setting debug mode for "rook-ceph-mon-b-debug"
+removing debug mode from "rook-ceph-mon-b-debug"
+deployment.apps "rook-ceph-mon-b-debug" deleted
+deployment.apps/rook-ceph-mon-b scaled
+```
+
+>Note: If you need to update the limits and request of the debug deployment that is created using debug command you can run:
+>```console
+>oc set resources deployment rook-ceph-osd-${osdid}-debug --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi
 ```
 
 ## Contributing
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -46,6 +46,10 @@ function print_usage() {
   echo "    status all                              : print the phase and conditions of all CRs"
   echo "    status <CR>                             : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
   echo "    purge-osd <osd-id> [--force]            : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo "  debug <subcommand>...                     : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
+  echo "    start  <deployment-name>                
+           [--alternate-image <alternate-image>]    : Start debugging a deployment with an optional container image"
+  echo "  stop <deployment-name>                    : Stop debugging a deployment"
   echo ""
 }
 
@@ -371,6 +375,125 @@ function check_mgr_pods_status_and_count() {
 }
 
 ####################################################################################################
+# 'kubectl rook-ceph debug commands
+####################################################################################################
+
+function remove_probe_from_deployment(){
+  deployment_spec=$1
+  deployment_spec=$(jq 'del(.template.spec.containers[0].livenessProbe)' <<< $deployment_spec)
+  deployment_spec=$(jq 'del(.template.spec.containers[0].startupProbe)' <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec 
+}
+
+function update_deployment_spec_command(){
+  deployment_spec=$1
+  main_command='["sleep","infinity"]'
+  deployment_spec=$(jq .template.spec.containers[0].command="$main_command" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.containers[0].args=[] <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
+}  
+
+function update_deployment_spec_image(){
+  deployment_spec=$1
+  alternalte_image=\"$2\"
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
+}
+
+function verify_debug_deployment(){
+  if [[ $deployment_name != "rook-ceph-mon-"* ]] && [[ $deployment_name != "rook-ceph-osd-"* ]]; then
+    fail_error "only mon or osd deployment name can passed for starting debug mode"
+  fi
+  if KUBECTL_NS_CLUSTER get deployment "$deployment_name" &>/dev/null; then
+    echo "setting debug mode for \"$deployment_name\"" 
+  else  
+    fail_error "deployment $deployment_name doesn't exist"
+  fi
+}
+
+function run_start_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing mon or osd deployment name"
+  deployment_name="$1"
+  alternate_image_flag="${2:-""}"
+  alternate_image="${3:-""}"
+  verify_debug_deployment "$deployment_name"
+  
+  # copy the deployment spec before scaling it down
+  deployment_spec=$(KUBECTL_NS_CLUSTER get  deployments  "$deployment_name" -o json | jq -r ".spec")
+  # copy the deployment labels before scaling it down
+  labels=$(KUBECTL_NS_CLUSTER get  deployments  "$deployment_name" -o json | jq -r ".metadata.labels")
+  # add debug label to the list
+  labels=$(echo "$labels" | jq '. + {"ceph.rook.io/do-not-reconcile": "true"}')
+  # remove probes from the deployment
+  deployment_spec=$(remove_probe_from_deployment "$deployment_spec")
+  # update the deployment_spec with new image if alternate-image is passed
+  if [ -n "$alternate_image_flag" ]; then
+    if [ "$alternate_image_flag" == "--alternate-image" ]; then
+      if [ "$#" -eq 3 ]; then
+        echo "setting debug image to $alternate_image"
+        deployment_spec=$(update_deployment_spec_image "$deployment_spec" "$alternate_image")
+      else
+        fail_error "if --alternate-image flag is passed then image argument is mandatory"   
+      fi
+    else
+      fail_error "--alternate-image should be passed as flag"
+    fi
+  fi
+  # update the deployment_spec main container to be only a placeholder,
+  echo "setting debug command to main container"
+  deployment_spec=$(update_deployment_spec_command "$deployment_spec")
+
+  # scale the deployment to 0
+  KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=0
+  
+  # create debug deployment
+  cat <<EOF | $TOP_LEVEL_COMMAND create -f -
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+        name: "$deployment_name-debug"
+        namespace: "$ROOK_CLUSTER_NAMESPACE"
+        labels:
+            ${labels}
+    spec:
+        $deployment_spec
+EOF
+}
+
+function run_stop_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing mon or osd deployment name"
+  deployment_name="$1"
+  if [[ $deployment_name != *"debug" ]]; then
+    deployment_name="$deployment_name-debug"
+  fi
+  verify_debug_deployment "$deployment_name"
+  echo "removing debug mode from \"$deployment_name\""
+
+  # delete the deployment debug pod
+  KUBECTL_NS_CLUSTER delete deployment "$deployment_name"
+  # remove the debug suffix from the deployment
+  deployment_name="${deployment_name%'-debug'}"
+  # scale the deployment to 1
+  KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=1
+}
+
+function run_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'debug' subcommand"
+  subcommand="$1"
+  shift # remove the subcommand from the front of the arg list
+  case "$subcommand" in
+  start)
+    run_start_debug "$@"
+    ;;
+  stop)
+    run_stop_debug "$@"
+    ;;
+  *)
+  fail_error "'debug' subcommand '$subcommand' does not exist"
+    ;;
+  esac  
+}
+####################################################################################################
 # 'kubectl rook-ceph status' command
 ####################################################################################################
 # Disabling it for now, will enable once it is ready implementation
@@ -430,6 +553,9 @@ function run_main_command() {
     ;;
   health)
     run_cluster_health "$@"
+    ;;
+  debug)
+    run_debug "$@"
     ;;
   # status)
   #   run_status_command "$@"


kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,54 @@ function run_purge_osd() {
...
+  cat <<eof | TOP_LEVEL_COMMAND create -f -
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+        name:"$deployment_name-debug"
+        namespace: "$ROOK_OPERATOR_NAMESPACE"
+        labels:
+            ceph.rook.io/debug: "true"
+    spec:
+        $deployment_spec
```
parth-gr, line 302: The thing to look ,
How to edit this var `$deployment_spec`  for changing the cmd

kubectl-rook-ceph.sh
```diff
@@ -27,24 +27,26 @@ function print_usage() {
...
-  echo "    purge-osd <osd-id> [--force]            : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo "    version                                                   : print the version of Rook"
+  echo "    status                                                    : print the phase and conditions of the CephCluster CR"
+  echo "    status all                                                : print the phase and conditions of all CRs"
+  echo "    status <CR>                                               : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
+  echo "    purge-osd <osd-id> [--force]                              : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo " create-debug [--name deployment] [--image <alternate-image>] : Make the OSD or mon in debug mode"
+  echo " remove-debug [--name deployment]                             : Make the OSD or mon out of debug mode"
```
parth-gr, line 49: @travisn 
Here it should be the deployment name[rook-ceph-mon-b] or the debug deployment name[rook-ceph-mon-b-debug].

Currently using the  `deployment name[rook-ceph-mon-b]`

parth-gr (commented):
The code works fine now,
Debug pod is started running,

Maybe we can try testing this with the rook-PR or [Randy Martinez](https://github.com/randymtz) you can also try this or first merge this and test completely

Thanks

parth-gr:
Process:
```
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh create-debug rook-ceph-mon-b
create debug mode for rook-ceph-mon-b
deployment rook-ceph-mon-b exist, NAME              READY   UP-TO-DATE   AVAILABLE   AGE
rook-ceph-mon-b   1/1     1            1           13d
alternate-image is not passed, so using the same container image and updating main container command
deployment.apps/rook-ceph-mon-b scaled
deployment.apps/rook-ceph-mon-b-debug created
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh remove-debug rook-ceph-mon-b
remove debug mode
deployment rook-ceph-mon-b-debug exist, NAME                    READY   UP-TO-DATE   AVAILABLE   AGE
rook-ceph-mon-b-debug   0/1     1            0           14s
deployment.apps "rook-ceph-mon-b-debug" deleted
deployment.apps/rook-ceph-mon-b scaled
[rider@localhost kubectl-rook-ceph]$ 

```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+    echo "deployment $deployment_name doesn't exist"
+    exit 1
+  fi
+  
+  # copy the deployment spec before scaling it down
+  deployment_spec=$($TOP_LEVEL_COMMAND get  deployments --namespace "$ROOK_OPERATOR_NAMESPACE"  "$deployment_name"   -o json | jq -r ".spec")
+  
+  # update the deployment_spec main container to be only a placeholder,
+  # also update image if alternate-image is passed
+  if [[ -z "${2:-""}" ]]; then
```
BlaineEXE, line 302: Instead of using `$1` and `$2`, please use `local $var="$1"` at the top of a function for better readability. Or `local $var2="${2:-""}"`

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+
+  # scale the deployment to 0
+  $TOP_LEVEL_COMMAND scale deployments --namespace "$ROOK_OPERATOR_NAMESPACE" "$deployment_name" --replicas=0
+  
+  # create debug deployment
+  cat <<eof | $TOP_LEVEL_COMMAND create -f -
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+        name: "$deployment_name-debug"
```
BlaineEXE, line 319: I think we should leave the deployment name the same. Unless I missed a detail, I don't think the operator knows to check whether a deployment with the name `*-debug` exists. I think it will only look for the deployment name coded into Rook. Does this make sense @travisn ?

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+function run_remove_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'mon/osd' ID"
+  deployment_name=$1
+  echo "remove debug mode"
+  if output=$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" get deployment "$deployment_name-debug"); then
+    echo "deployment $deployment_name-debug exist, $output" 
+  else  
+    echo "deployment $deployment_name-debug doesn't exist"
+    exit 1
+  fi
```
BlaineEXE, line 337: I see this same logic above. Maybe a helper function to get a deployment by name would be helpful?

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+
+function run_create_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'mon/osd/mgr/mds' deployment name"
+  deployment_name=$1
+  echo "create debug mode for $deployment_name"
+  if output=$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" get deployment "$deployment_name"); then
+    echo "deployment $deployment_name exist, $output" 
+  else  
+    echo "deployment $deployment_name doesn't exist"
+    exit 1
```
BlaineEXE, line 294: Don't forget that the `fail_error` helper exists. This will help keep error output consistent.

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+
+function update_deployment_spec_image(){
+  deployment_spec=$1
+  alternalte_image=$2
+  deployment_spec=$(jq .template.spec.containers[0].env[0].value="$alternalte_image" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+}
+
+function run_create_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'mon/osd/mgr/mds' deployment name"
```
BlaineEXE, line 287: Do we need to specify mon/osd/mgr/mds here? Do we support mgr/mds at this point in time?

```suggestion
  [[ -z "${1:-""}" ]] && fail_error "Missing deployment name"
```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+  deployment_name=$1
+  echo "create debug mode for $deployment_name"
```
BlaineEXE, line 289:
```suggestion
  echo "setting debug mode for $deployment_name"
```

kubectl-rook-ceph.sh
```diff
@@ -27,24 +27,26 @@ function print_usage() {
...
-  echo "    status <CR>                             : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
-  echo "    purge-osd <osd-id> [--force]            : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo "    version                                                   : print the version of Rook"
+  echo "    status                                                    : print the phase and conditions of the CephCluster CR"
+  echo "    status all                                                : print the phase and conditions of all CRs"
+  echo "    status <CR>                                               : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
+  echo "    purge-osd <osd-id> [--force]                              : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo " create-debug [--name deployment] [--image <alternate-image>] : Make the OSD or mon in debug mode"
```
BlaineEXE, line 48: I think `set-debug` may be more clear here. WDYT @travisn ?

BlaineEXE, line 48: I feel like `create-debug` is similar in scope to `purge-osd` which is under the `rook` subcommand. Should this move under that? @travisn ?

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+function update_deployment_spec_image(){
+  deployment_spec=$1
+  alternalte_image=$2
+  deployment_spec=$(jq .template.spec.containers[0].env[0].value="$alternalte_image" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+}
+
+function run_create_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'mon/osd/mgr/mds' deployment name"
+  deployment_name=$1
```
BlaineEXE, line 288:
```suggestion
  deployment_name="$1"
```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+    echo "updating the alternalte image and main container command"
+    update_deployment_spec_command "$deployment_spec"
+    update_deployment_spec_image "$deployment_spec" "$2"
+  fi  
+
+  # scale the deployment to 0
+  $TOP_LEVEL_COMMAND scale deployments --namespace "$ROOK_OPERATOR_NAMESPACE" "$deployment_name" --replicas=0
+  
+  # create debug deployment
+  cat <<eof | $TOP_LEVEL_COMMAND create -f -
```
BlaineEXE, line 315: EOF is generally given in all caps. Let's keep common patterns.

```suggestion
  cat <<EOF | $TOP_LEVEL_COMMAND create -f -
```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+        labels:
+            ceph.rook.io/debug: "true"
+    spec:
+        $deployment_spec
+eof
+}
+
+function run_remove_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'mon/osd' ID"
+  deployment_name=$1
```
BlaineEXE, line 330:
```suggestion
  deployment_name="$1"
```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+  echo "remove debug mode"
```
BlaineEXE, line 331:
```suggestion
  echo "removing debug mode from $deployment_name"
```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+        namespace: "$ROOK_OPERATOR_NAMESPACE"
+        labels:
+            ceph.rook.io/debug: "true"
+    spec:
+        $deployment_spec
+eof
+}
+
+function run_remove_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing 'mon/osd' ID"
```
BlaineEXE, line 329:
```suggestion
  [[ -z "${1:-""}" ]] && fail_error "Missing deployment name"
```

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+  deployment_name=$1
+  echo "remove debug mode"
+  if output=$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" get deployment "$deployment_name-debug"); then
+    echo "deployment $deployment_name-debug exist, $output" 
+  else  
+    echo "deployment $deployment_name-debug doesn't exist"
+    exit 1
+  fi
+  deployment_name=$1
```
BlaineEXE, line 338:
```suggestion
  deployment_name="$1"
```

kubectl-rook-ceph.sh
BlaineEXE, line 302: We need to parse arguments for this function since it can take a flag and takes 1 arg.

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,83 @@ function run_purge_osd() {
...
+  # update the deployment_spec main container to be only a placeholder,
+  # also update image if alternate-image is passed
+  if [[ -z "${2:-""}" ]]; then
+    echo "alternate-image is not passed, so using the same container image and updating main container command"
+    update_deployment_spec_command "$deployment_spec"
+  else
+    echo "updating the alternalte image and main container command"
+    update_deployment_spec_command "$deployment_spec"
+    update_deployment_spec_image "$deployment_spec" "$2"
+  fi
```
BlaineEXE, line 309: What if this instead did something like this...

```suggestion
  if [[ -n "$desired_image" ]]; then 
    echo "setting debug image to $desired_image"
    update_deployment_spec_image "$deployment_spec" "$desired_image"
  fi
  echo "setting debug command to main container"
```

If "setting debug image" is not present, it implies that it was left at its usual value. I want to make sure we don't log too much for the user.

BlaineEXE (changes requested):
Make sure you are quoting variables so that any values that might contain spaces aren't lost.

Also, it would be helpful for me to see what the CLI output looks like for the commands with various options and for various error conditions. It's good to make sure the tool outputs in a similar style for all commands. Consistency is usually good for the end user.

kubectl-rook-ceph.sh
travisn: > I feel like `create-debug` is similar in scope to `purge-osd` which is under the `rook` subcommand. Should this move under that? @travisn ?

What about `start-debug` and `stop-debug`?

travisn: Instead of increasing the indentation of all the commands, how about moving the `--image` param to the next line? We will surely have commands with multiple args in the future, so this seems like a good pattern to start.

kubectl-rook-ceph.sh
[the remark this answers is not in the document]
travisn:
```suggestion
  echo " start-debug [--name deployment] [--image <alternate-image>] : Create the OSD or mon debug deployment"
  echo " stop-debug [--name deployment]                             : Remove the OSD or mon debug deployment"
```

kubectl-rook-ceph.sh
travisn: I think we should only support the debug pods for mons and osds. Everything else should be doable in the toolbox. So maybe we can add an error check that it's only for mons and osds
```suggestion
  [[ -z "${1:-""}" ]] && fail_error "Missing mon or OSD deployment name"
```

travisn (changes requested):

subhamkrai (changes requested):
@parth-gr please add the command for CI too. Thank!

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,87 @@ function run_purge_osd() {
...
+# 'kubectl rook-ceph debug commands
+####################################################################################################
+
+function update_deployment_spec_command(){
+  deployment_spec=$1
+  main_command="[\"sleep\",\"infinity\"]"
+  deployment_spec=$(jq .template.spec.containers[0].command="$main_command" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.initContainers[1].command="$main_command" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.initContainers[0].command="$main_command" <<< $deployment_spec)
+}
```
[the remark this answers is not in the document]
parth-gr: No it doesn't returns,

We cannot return string in shell so it just update the `deployment_spec` which is passed as a parameter

kubectl-rook-ceph.sh
parth-gr: we can discuss over this

kubectl-rook-ceph.sh
parth-gr: how about both as a argument, currently code works for that

kubectl-rook-ceph.sh
BlaineEXE: I think using `*-debug` makes sense. I forgot the detail about the original being labeled and scaled down.

kubectl-rook-ceph.sh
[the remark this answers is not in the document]
BlaineEXE: Functions in bash operate essentially the same way as external commands like `grep`. You can capture subshell output from calling a function to get a string return.

```sh
#! /usr/bin/env bash
set -eEuo pipefail

function do_math() {
  res="$(( $1 ))"
  echo "$res"
}

res="$(do_math "1 + 2")"

echo "result was: $res"

```

kubectl-rook-ceph.sh
BlaineEXE: I would say it should accept either.

kubectl-rook-ceph.sh
```diff
@@ -263,6 +266,82 @@ function run_purge_osd() {
...
+  deployment_spec=$1
+  alternalte_image=$2
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+}
+
+function verify_deployment_exist(){
+  if [[ $deployment_name != *"mon"* ]] && [[ $deployment_name != *"osd"* ]]; then
+    fail_error "only mon or osd deployment name can passed for starting debug mode"
+  fi
+  if output="$($TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" get deployment "$deployment_name")"; then
```
BlaineEXE, line 290: Should this be `CLUSTER` namespace? I see a lot of instances of the operator namespace, but I don't think we need to do anything in that namespace, right?

```suggestion
  if output="$($TOP_LEVEL_COMMAND --namespace "$ROOK_CLUSTER_NAMESPACE" get deployment "$deployment_name")"; then
```

parth-gr: Yaa, it should be cluster_namespace I also think the same, 
But look here https://github.com/rook/kubectl-rook-ceph/blob/256ca021a47939dbe3cae4b954c4d934a61c4956/kubectl-rook-ceph.sh#L335

@subhamkrai your thoughts?

kubectl-rook-ceph.sh
```diff
@@ -263,6 +266,82 @@ function run_purge_osd() {
...
+  deployment_spec=$1
+  main_command='["sleep","infinity"]'
+  deployment_spec=$(jq .template.spec.containers[0].command="$main_command" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.containers[0].args=[] <<< $deployment_spec)
+}  
+
+function update_deployment_spec_image(){
+  deployment_spec=$1
+  alternalte_image=$2
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
```
parth-gr, line 283: @BlaineEXE can you please comment the example about you were saying to return the value by `echo`

parth-gr:
```
function fun1(){
  echo 34
}

function fun2(){
  local res=$(fun1)
  echo $res
}
```

This way it is working,
Thanks

BlaineEXE: https://github.com/rook/kubectl-rook-ceph/pull/36#discussion_r932425506

subhamkrai (changes requested):
please add the new command in the readme doc in the commands section(no need to add examples).

Also, check CI errors. Thanks!

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -74,7 +74,12 @@ jobs:
           kubectl --namespace rook-ceph scale deploy/rook-ceph-osd-0 --replicas=0
           # we need to sleep so the osd will be marked down before purging the osd
           sleep 5
-          kubectl-rook_ceph rook purge-osd 0 --force
+          kubectl rook_ceph rook purge-osd 0 --force
```
subhamkrai, line 77: similar changes require in another ci file `ci-for-diff-ns`

subhamkrai (changes requested):

.github/workflows/ci-for-default-ns.yaml
parth-gr: DOne

kubectl-rook-ceph.sh
subhamkrai: change this to cluster namespace since this is osd and mons deploy

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,94 @@ function run_purge_osd() {
...
+  deployment_spec=$(jq .template.spec.containers[0].env[0].value="$alternalte_image" <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
+}
+
+function verify_deployment_exist(){
+  if [[ $deployment_name != *"mon"* ]] && [[ $deployment_name != *"osd"* ]]; then
+    fail_error "only mon or osd deployment name can passed for starting debug mode"
+  fi
+  if output="$(KUBECTL_NS_CLUSTER get deployment "$deployment_name")"; then
+    echo "deployment $deployment_name exist, $output"
```
subhamkrai, line 304: I don't think we need to print in case of a positive case. This is a validation method it should be like
```
if deployment doesn't exists; then
 fail
fi

```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,94 @@ function run_purge_osd() {
...
+  deployment_spec=$1
+  alternalte_image=\"$2\"
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.containers[0].env[0].value="$alternalte_image" <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
+}
+
+function verify_deployment_exist(){
+  if [[ $deployment_name != *"mon"* ]] && [[ $deployment_name != *"osd"* ]]; then
+    fail_error "only mon or osd deployment name can passed for starting debug mode"
```
subhamkrai, line 301:
```suggestion
  if [[ $deployment_name != *"mon"* ]] && [[ $deployment_name != *"osd"* ]]; then
    fail_error "only mon or osd deployment name can passed for starting debug mode"
```
above should not be required here, you can count the an argument somewhere around L312, use `$#` to get  count of argument

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,94 @@ function run_purge_osd() {
...
+  fi
+}
+
+function run_start_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing mon or osd deployment name"
+  deployment_name="$1"
+  alternate_image_flag="${2:-""}"
+  alternate_image="${3:-""}"
+  verify_deployment_exist "$deployment_name"
+  echo "setting debug mode for \"$deployment_name\""
```
subhamkrai, line 316:
```suggestion
  echo "setting debug mode for \"$deployment_name\""
```
merge this echo with L326 and L335 to look something like
```
Setting debug mode for deployment <name> for container <name>
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,94 @@ function run_purge_osd() {
...
+        echo "setting debug image to $alternate_image"
+        deployment_spec=$(update_deployment_spec_image "$deployment_spec" "$alternate_image")
+      else
+        fail_error "if --alternate-image flag is passed then image argument is mandatory"   
+      fi 
+    else
+      fail_error "--alternate-image should be passed as flag"
+    fi
+  fi  
+  echo "setting debug command to main container"
```
subhamkrai, line 335:
```suggestion
  echo "setting debug command to main container"
```
can we get the name of `main/default` container?

subhamkrai (changes requested):
marking request changes for above

kubectl-rook-ceph.sh
subhamkrai: @parth-gr ^^

kubectl-rook-ceph.sh
parth-gr: Done,
In respect that I am calling a subcommand and which possible this way, so moved the `setting debug mode for "rook-ceph-mon-a"` from main function to here to it's echo and removed the validation command

kubectl-rook-ceph.sh
parth-gr: SO @travisn mentioned in one comment we also specifically need to check the deployment name passed should only be of mon and osd so anyhow would need this.

kubectl-rook-ceph.sh
parth-gr: There is in between image echo command too 
`echo "setting debug image to $alternate_image"`

So anyhow we need to keep it separately, right?

I did a lot's of echo so it can be easy to reproduce from the script which subtask it got failed

kubectl-rook-ceph.sh
parth-gr: name should be simply mon or osd should we need to print that, by fetching it?

kubectl-rook-ceph.sh
parth-gr: That means it should accept `rook-ceph-mon-b` and `rook-ceph-mon-b-debug` both?
@BlaineEXE

kubectl-rook-ceph.sh
```diff
@@ -263,6 +265,87 @@ function run_purge_osd() {
...
+  main_command="[\"sleep\",\"infinity\"]"
+  deployment_spec=$(jq .template.spec.containers[0].command="$main_command" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.initContainers[1].command="$main_command" <<< $deployment_spec)
+  deployment_spec=$(jq .template.spec.initContainers[0].command="$main_command" <<< $deployment_spec)
+}  
+
+function update_deployment_spec_image(){
+  deployment_spec=$1
+  alternalte_image=$2
+  deployment_spec=$(jq .template.spec.containers[0].env[0].value="$alternalte_image" <<< $deployment_spec)
```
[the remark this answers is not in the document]
parth-gr: Are we sure??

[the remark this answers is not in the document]
parth-gr: > Are we sure about the env, because I think it is needed for container env??

kubectl-rook-ceph.sh
parth-gr: sure done

kubectl-rook-ceph.sh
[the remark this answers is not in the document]
BlaineEXE: From looking at the Ceph code, I believe the var is just used to get the name of the image from debug logs. But since we are in a manual debug scenario, it shouldn't matter.

kubectl-rook-ceph.sh
BlaineEXE: That was my thought. Bash has built-in basic string manipulation. You can remove a suffix from a string pretty easily.

```sh
base_name="${input_deployment_name%'-debug'}" # strip '-debug' suffix
# the base_name is always rook-ceph-mon-a regardless of whether
# input_deployment_name has the -debug suffix or not
```

parth-gr: yaa thankx

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,97 @@ function run_purge_osd() {
...
+  else  
+    fail_error "deployment $deployment_name doesn't exist"
+  fi
+}
+
+function run_start_debug(){
+  [[ -z "${1:-""}" ]] && fail_error "Missing mon or osd deployment name"
+  deployment_name="$1"
+  alternate_image_flag="${2:-""}"
+  alternate_image="${3:-""}"
```
BlaineEXE, line 313: What if users want to issue the command like `kubectl rook-ceph start-debug <deploy> --image=my-image`? Or what if they do `kubectl rook-ceph start-debug --image my-image <deploy>`? This will fail in both cases. We should use the `parse_flags()` function for all flag parsing.

The commented-out `run_status_command()` example should help.

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,97 @@ function run_purge_osd() {
...
+
+function update_deployment_spec_image(){
+  deployment_spec=$1
+  alternalte_image=\"$2\"
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
+}
+
+function verify_deployment_exist(){
+  if [[ $deployment_name != *"mon"* ]] && [[ $deployment_name != *"osd"* ]]; then
```
BlaineEXE, line 299:
```suggestion
  if [[ $deployment_name != "rook-ceph-mon-"* ]] && [[ $deployment_name != "rook-ceph-osd-"* ]]; then
```

This could still allow debugging on an RGW deployment where the CephObjectStore is named `pokemon` (rook-ceph-rgw-pokemon-a). It is best to make these checks as specific as possible so we don't allow unsupported cases.

kubectl-rook-ceph.sh
```diff
@@ -45,6 +45,9 @@ function print_usage() {
   echo "    status all                              : print the phase and conditions of all CRs"
   echo "    status <CR>                             : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
   echo "    purge-osd <osd-id> [--force]            : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo " start-debug <deployment-name> 
+                        [--image <alternate-image>] : Create the OSD or mon in debug deployment"
+  echo " stop-debug <deployment-name>               : Remove the OSD or mon out of debug deployment"
```
BlaineEXE, line 50: What if we did `debug start` and `debug stop` instead of having two top-level commands? @travisn?

```suggestion
  echo " debug                                      : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only
  echo "   start  <deployment-name>                 : Start debugging a deployment with an optional container image
  echo "     [--image <alternate-image>]"
  echo "   stop  <deployment-name>                  : Stop debugging a deployment
```

IMO, this allows us a little more space to document what debug mode is for a deployment. Otherwise, users may not know what to expect.

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,97 @@ function run_purge_osd() {
...
+  verify_deployment_exist "$deployment_name"
+  echo "removing debug mode from \"$deployment_name\""
+
+  # delete the deployment debug pod
+  KUBECTL_NS_CLUSTER delete deployment "$deployment_name"
+  # remove the debug suffix from the deployment
+  deployment_name="${deployment_name%'-debug'}"
+  # scale the deployment to 1
+  KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=1
+}
```
BlaineEXE, line 368: I think the operator should be restarted afterwards so that it re-reconciles the deployments that may have been in debug mode. An OSD or mon may need to be updated, for example.

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,97 @@ function run_purge_osd() {
...
+  alternalte_image=\"$2\"
+  deployment_spec=$(jq .template.spec.containers[0].image="$alternalte_image" <<< $deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
+}
+
+function verify_deployment_exist(){
+  if [[ $deployment_name != *"mon"* ]] && [[ $deployment_name != *"osd"* ]]; then
+    fail_error "only mon or osd deployment name can passed for starting debug mode"
+  fi
+  if _="$(KUBECTL_NS_CLUSTER get deployment "$deployment_name")"; then
```
BlaineEXE, line 302: If the intent is just to swallow the console output, it is likely better to use `KUBECTL_NS_CLUSTER get deployment "$deployment_name" &>/dev/null`

```suggestion
  if KUBECTL_NS_CLUSTER get deployment "$deployment_name" &>/dev/null; then
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,97 @@ function run_purge_osd() {
...
+    echo "setting debug mode for \"$deployment_name\""
```
BlaineEXE, line 303: This message is untrue when `verify_deployment_exist` is used from `run_stop_debug()`. I would remove this from the function and allow the caller to print a "good" message after this function returns successfully.

```sh
if ! KUBECTL_NS_CLUSTER get deployment "$deployment_name"; then
  fail_error "deployment $deployment_name doesn't exist"
fi
```

kubectl-rook-ceph.sh
BlaineEXE, line 299: Why is the function `verify_deployment_exist` verifying that it is a mon/osd deployment? By the name of this function, I would expect it to return success or failure depending on if the deployment exists or not and should allow any deployment. Instead, I would rename this to `verify_debug_deployment()`.

kubectl-rook-ceph.sh
```diff
@@ -273,6 +276,97 @@ function run_purge_osd() {
...
+      if [ "$#" -eq 3 ]; then
+        echo "setting debug image to $alternate_image"
+        deployment_spec=$(update_deployment_spec_image "$deployment_spec" "$alternate_image")
+      else
+        fail_error "if --alternate-image flag is passed then image argument is mandatory"   
+      fi
+    else
+      fail_error "--alternate-image should be passed as flag"
+    fi
+  fi
```
BlaineEXE, line 332: If we use `parse_flags()`, then the flag parsing code will be a separate function and won't make the code that implements the logic longer and harder to read.

BlaineEXE (changes requested):

kubectl-rook-ceph.sh
parth-gr: After returning from the debug mode,

The support team might need to do some injections or corrections so the bug can resolve, in between if they reconcile then they might be gone to some different state maybe. 

I would say if it is needed we can support this as an extra flag, thoughts @BlaineEXE ?

BlaineEXE: @travisn might have thoughts as well, but I'm still of the opinion that once debug mode is stopped, the user should have the expectation that the resource should go back to being healthy and up-to-date with the rest of the cluster. I think a restart should be a given here.

kubectl-rook-ceph.sh
parth-gr: DOne, 
Anyway if I can print only false condition here,

Currently I need to go for if and else and print both it is present and not present

kubectl-rook-ceph.sh
parth-gr: I need to return some number value for it like 1-> there is no error, 2-> if there is error

The make a if else in main function as it is a shell function

I thought this will be a more complicated flow for reading, right?

Your thought?

kubectl-rook-ceph.sh
parth-gr: Probably parse_flags() also not help in position problem,

We might need to add `getopts` support for that

kubectl-rook-ceph.sh
travisn: Deleting a mon or osd deployment may already trigger a reconcile, let's test that first before an operator restart.

travisn (changes requested):

travisn:
In my testing with the operator skipping debug deployments (see https://github.com/rook/rook/pull/10585), we need the following changes:
- The mon and osd debug deployments are missing all the expected labels. The label `ceph.rook.io.debug: "true"` needs to be added to the list, instead of replacing all the labels. 
- The startupProbe needs to be removed from the debug deployments. After some time, I see them crashing because of the failing probe.

parth-gr:
> * The startupProbe needs to be removed from the debug deployments. After some time, I see them crashing because of the failing probe.
We need to also remove the liveness probe and readiness probe, right?

travisn:
> > * The startupProbe needs to be removed from the debug deployments. After some time, I see them crashing because of the failing probe.
> 
> We need to also remove the liveness probe and readiness probe, right? As liveness probe will again and again restart the application

Yes, all the probes need to be removed, thanks.

vumrao:
@parth-gr @travisn - I see we can define custom image ```start-debug <deployment-name> [--image <alternate-image>]` : Create the OSD or mon in debug deployment"``` - this will help a lot but I missed to mention another requirement and I am not sure if this can be implemented in this PR or a new commit in this same PR or new PR. 

But we also want to have a capability in these debugs PODs that they start with higher memory and CPU configuration as compared to the original PODs. For now, we achieve it with the following command:

```oc set resources deployment rook-ceph-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi```

The use case here is that the assigned resources to original PODs are sometimes not sufficient to run these offline ceph tools.

Thanks,
Vikhyat

**Comment by travisn** [comment]
Supporting `--limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` as args to the `debug start` command is simple enough. If those args are set, the script could run the command you mentioned.

**Comment by parth-gr** [comment]
> Supporting `--limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` as args to the `debug start` command is simple enough. If those args are set, the script could run the command you mentioned.

Just we need to manage the input correctly, So we should prior know what all different flags can be used with it.

File: .github/workflows/ci-for-default-ns.yaml [hunk 33]
```diff
@@ -76,7 +76,13 @@ jobs:
...
+
+          # for testing start-debug and stop-debug deployment 
+          kubectl rook_ceph start-debug rook-ceph-mon-a
+          sleep 5
```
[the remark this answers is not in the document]
parth-gr: https://github.com/rook/kubectl-rook-ceph/issues/42

Will fix by this

kubectl-rook-ceph.sh
```diff
@@ -273,6 +277,113 @@ function run_purge_osd() {
...
+  
+  # create debug deployment
+  cat <<EOF | $TOP_LEVEL_COMMAND create -f -
+    apiVersion: apps/v1
+    kind: Deployment
+    metadata:
+        name: "$deployment_name-debug"
+        namespace: "$ROOK_CLUSTER_NAMESPACE"
+        labels:
+            ceph.rook.io/debug: "true"
```
travisn, line 348: In https://github.com/rook/rook/pull/10585 the label is now updated to `ceph.rook.io/do-not-reconcile: "true"` instead of the `debug` label name.

travisn:
> > Supporting `--limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` as args to the `debug start` command is simple enough. If those args are set, the script could run the command you mentioned.
> 
> Just we need to manage the input correctly, So we should prior know what all different flags can be used with it.

To keep it simple, I don't think we need to validate the input for these params. If there is an invalid syntax that one command can just fail.

kubectl-rook-ceph.sh
parth-gr: Okay

parth-gr:
> > > Supporting `--limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` as args to the `debug start` command is simple enough. If those args are set, the script could run the command you mentioned.
> > 
> > 
> > Just we need to manage the input correctly, So we should prior know what all different flags can be used with it.
> 
> To keep it simple, I don't think we need to validate the input for these params. If there is an invalid syntax that one command can just fail.

Okay

parth-gr:
> --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi

@travisn  I have a question are these values need to to applied while creating the pod, 

If yes-> then we actually want to replace/change the values manually by code. And this needs to update a code and to be specific

If no-> then we can run the above `oc set resources deployment rook-ceph-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` what @vumrao shared.

travisn:
> > --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi
> 
> @travisn I have a question are these values need to to applied while creating the pod,
> 
> If yes-> then we actually want to replace/change the values manually by code. And this needs, to update the code and be specific
> 
> If no-> then we can run the above `oc set resources deployment rook-ceph-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` what @vumrao shared.

Actually, if it's anyway a separate command that would require another pod restart, it doesn't seem like we get much from integrating these params into the `debug start` command. @vumrao How common would you say it is to increase the memory when you need to do this? How about just continuing to use that separate command to set the memory on the deployment? If these params were as simple to apply to the spec as the image it would be no big deal, but it's a bit more complicated for resources.

vumrao:
> > > --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi
> > 
> > 
> > @travisn I have a question are these values need to to applied while creating the pod,
> > If yes-> then we actually want to replace/change the values manually by code. And this needs, to update the code and be specific
> > If no-> then we can run the above `oc set resources deployment rook-ceph-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi` what @vumrao shared.
> 
> Actually, if it's anyway a separate command that would require another pod restart, it doesn't seem like we get much from integrating these params into the `debug start` command. @vumrao How common would you say it is to increase the memory when you need to do this? How about just continuing to use that separate command to set the memory on the deployment? If these params were as simple to apply to the spec as the image it would be no big deal, but it's a bit more complicated for resources.

If I understand correctly, providing these resource options as a command line during maintenance pod creation needs a lot of changes? Hence we want to first create a maintenance POD and then run the following command to change the resources something like the following?

```oc set resources deployment rook-ceph-maintenace-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi```

Is my understanding correct?

**Comment by parth-gr** [comment]
> `oc set resources deployment rook-ceph-maintenace-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi`
> 
> Is my understanding correct?

yes this is what we are thinking

File: kubectl-rook-ceph.sh (hunk 26 above)
Comment (BlaineEXE) [reply kubectl-rook-ceph.sh:321]: You're right about the position. `parse_flags()` assumes it should stop when it reaches a non-flag argument. We could probably solve that by calling `parse_flags()` twice.

**Comment by parth-gr** [comment]
@travisn so after applying empty map to probes `{}` the error was coming from K8's that we can't specify it,

So with the help of `jq` I deleted the probe's key from the specs,

It working fine now,

Ig it now just requires final testing with rook PR so this PR can be merged.

File: kubectl-rook-ceph.sh [hunk 35]
```diff
@@ -45,6 +45,10 @@ function print_usage() {
   echo "    status all                              : print the phase and conditions of all CRs"
   echo "    status <CR>                             : print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc"
   echo "    purge-osd <osd-id> [--force]            : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
+  echo "  debug <subcommand>...                     : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
+  echo "    start  <deployment-name>                
+                     [--image <alternate-image>]    : Start debugging a deployment with an optional container image"
+  echo "  stop  <deployment-name>                   : Stop debugging a deployment"
```
travisn, line 51: indentation
```suggestion
  echo "    stop  <deployment-name>                   : Stop debugging a deployment"
```

README.md
```diff
@@ -40,6 +40,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs."
 
+- `debug <subcommand>...`           : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```
travisn, line 43: We really need more details documented about the behavior of this command.
- Describe the scenario when it could be used
- Explain that it will scale down the existing deployment, start a new debug deployment without probes and that sleeps forever so you can connect and run ceph commands

Perhaps we could put these details in a new Examples section below. We don't need to include the details here so we can keep this section simple.

travisn (changes requested):
Testing with https://github.com/rook/rook/pull/10585 looks good! Just a final suggestion on documentation.

README.md
parth-gr: Done

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -56,6 +56,18 @@ jobs:
           kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
           kubectl rook_ceph -o test-operator -n test-cluster health
 
+          # for testing start-debug and stop-debug deployment
```
travisn, line 59: How about checking that the deployments are created, the debug pods are running, and the original pods are stopped?

vumrao:
> > `oc set resources deployment rook-ceph-maintenace-osd-${osdid} --limits=cpu=8,memory=64Gi --requests=cpu=8,memory=64Gi`
> > Is my understanding correct?
> 
> yes this is what we are thinking

Sounds good! Can you please document this in the feature?

README.md
```diff
@@ -42,6 +42,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
 
+- `debug <subcommand>...`           : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```
travisn, line 45: A link could help find the details
```suggestion
- `[debug](#debug-mode) <subcommand>...`           : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```

README.md
```diff
@@ -121,6 +126,37 @@ kubectl rook-ceph ceph versions
 }
 ```
 
+### Debug Mode
+
+Debug Mode will be needed when cluster mon's or osd are been into error state or in the state when
```
travisn, line 131: For lines 131-136:
```suggestion
Debug mode can be useful when a mon or OSD needs advanced maintenance operations that require the daemon to be stopped. Ceph tools such as `ceph-objectstore-tool`,`ceph-bluestore-tool`, or `ceph-monstore-tool` are commonly used in these scenarios. Debug mode will set up the mon or OSD so that these commands can be run.

Debug mode will automate the following:
1. Scale down the existing mon or OSD deployment
2. Start a new debug deployment where operations can be performed directly against the mon or OSD without that daemon running
   a. The main container sleeps so you can connect and run the ceph commands
   b. Liveness and startup probes are removed
```

README.md
```diff
@@ -121,6 +126,37 @@ kubectl rook-ceph ceph versions
...
+### Debug Mode
+
+Debug Mode will be needed when cluster mon's or osd are been into error state or in the state when 
+the pods are not responding, So we use ceph tools like  ceph-objectstore-tool or ceph-bluestore-tool 
+or ceph-monstore-tool, etc  for debugging.
+Debug Mode will automate this process of using ceph tools for the opeartor. 
+
+Debug Mode will work like it will scale down the existing deployment, start a new debug deployment without probes and that container pod will sleeps forever so you can connect and run ceph tools commands.
+
+Starting Debug Mode:
```
travisn, line 138:
```suggestion
For example, start the debug pod for mon `b`:
```

README.md
```diff
@@ -121,6 +126,37 @@ kubectl rook-ceph ceph versions
...
+kubectl rook-ceph debug start rook-ceph-mon-b
+```
+```text
+setting debug mode for "rook-ceph-mon-b"
+setting debug command to main container
+deployment.apps/rook-ceph-mon-b scaled
+deployment.apps/rook-ceph-mon-b-debug created
+```
+
+Stopping Debug Mode:
```
travisn, line 149:
```suggestion
When finished, stop debug mode and restore the original daemon:
```

README.md
```diff
@@ -121,6 +126,37 @@ kubectl rook-ceph ceph versions
...
+```console
+kubectl rook-ceph debug start rook-ceph-mon-b
+```
+```text
+setting debug mode for "rook-ceph-mon-b"
+setting debug command to main container
+deployment.apps/rook-ceph-mon-b scaled
+deployment.apps/rook-ceph-mon-b-debug created
+```
+
```
travisn, line 148: Now connect to the daemon pod and perform operations:
```
kubectl exec <debug-pod> -- <ceph command>
```

travisn (changes requested):

.github/workflows/ci-for-diff-ns.yaml
parth-gr: Working on this https://github.com/rook/kubectl-rook-ceph/issues/42

Actually in the whole CI there are random sleep so need to update that too

kubectl-rook-ceph.sh
parth-gr: THat makes other input more complicated,

How about this input,

`kubectl rook_ceph xyz  --flag1 parmeter1 --flag2 parmater2 --flag3`

README.md
```diff
@@ -121,6 +126,59 @@ kubectl rook-ceph ceph versions
 }
 ```
 
+### Debug Mode
+
+Debug mode can be useful when a mon's or OSD needs advanced maintenance operations that require the daemon to be stopped. Ceph tools such as `ceph-objectstore-tool`,`ceph-bluestore-tool`, or `ceph-monstore-tool` are commonly used in these scenarios. Debug mode will set up the mon or OSD so that these commands can be run.
```
travisn, line 131:
```suggestion
Debug mode can be useful when a mon or OSD needs advanced maintenance operations that require the daemon to be stopped. Ceph tools such as `ceph-objectstore-tool`,`ceph-bluestore-tool`, or `ceph-monstore-tool` are commonly used in these scenarios. Debug mode will set up the mon or OSD so that these commands can be run.
```

README.md
```diff
@@ -42,6 +42,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
 
+- [debug](#debug-mode) `<subcommand>...`  : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```
travisn, line 45: The rendered link doesn't look quite right how I suggested it the first time. How about this?
```suggestion
- `debug <subcommand>...`  : [Debug a deployment](#debug-mode) by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```

README.md
```diff
@@ -121,6 +126,59 @@ kubectl rook-ceph ceph versions
...
+1. Scale down the existing mon or OSD deployment
+2. Start a new debug deployment where operations can be performed directly against the mon or OSD without that daemon running
+   
+   a. The main container sleeps so you can connect and run the ceph commands
+   
+   b. Liveness and startup probes are removed
+
+   c. If alternate Image is passed by --alternate-image flag then the new debug deployment container will be using alternate Image.
+
+Debug Mode will work like it will scale down the existing deployment, start a new debug deployment without probes and that container pod will sleeps forever so you can connect and run ceph tools commands.
```
travisn, line 143: We don't need this line anymore, right? It's covered by the bullets above

README.md
```diff
@@ -42,6 +42,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
 
+- [debug](#debug-mode) `<subcommand>...`  : Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
+  - `start  <deployment-name> ` 
+    `[--alternate-image <alternate-image>]` : Start debugging a deployment with an optional container image"
```
travisn, line 47:
```suggestion
    `[--alternate-image <alternate-image>]` : Start debugging a deployment with an alternative ceph container image"
```

README.md
```diff
@@ -121,6 +126,59 @@ kubectl rook-ceph ceph versions
...
+deployment.apps/rook-ceph-mon-b scaled
+deployment.apps/rook-ceph-mon-b-debug created
+```
+
+Now connect to the daemon pod and perform operations:
+```console
+kubectl exec <debug-pod> -- <ceph command>
+```
+
+You can also use alternate-image while starting the debug mode:
```
travisn, line 161: I'd suggest removing lines 161-164, the alternate-image flag seems self explanatory. Let's keep this example focused.

travisn (commented):
A few more small doc suggestions

README.md
```diff
@@ -121,6 +126,52 @@ kubectl rook-ceph ceph versions
...
+```
+```text
+setting debug mode for "rook-ceph-mon-b-debug"
+removing debug mode from "rook-ceph-mon-b-debug"
+deployment.apps "rook-ceph-mon-b-debug" deleted
+deployment.apps/rook-ceph-mon-b scaled
+```
+
+>Note: If you need to update the limits and request of the debug deployment that is created using debug command you can run:
+>```console
```
travisn, line 171: The next section is rendered incorrectly, I think this character needs to be removed.
```suggestion
```console
```

README.md
```diff
@@ -42,6 +42,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
 
+- `debug <subcommand>...`  : [Debug a deployment](#debug-mode)  by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```
travisn, line 45: I just realized the other commands don't mention subcommands in this doc. This seems consistent
```suggestion
- `debug`  : [Debug a deployment](#debug-mode)  by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```

travisn (approved):
approved with a couple nits

README.md
parth-gr: This is added intentionally,
And after this at cmd `> oc` also

So the console cmd can be the part of `Note` only

README.md
parth-gr: Yup

README.md
```diff
@@ -42,6 +42,11 @@ To install the plugin, run:
   - `status <CR>` : Print the phase and conditions of CRs of a specific type, such as `cephobjectstore`, `cephfilesystem`, etc
   - `purge-osd <osd-id> [--force]` : Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
 
+- `debug` : [Debug a deployment](#debug-mode)  by scaling it down and creating a debug copy. This is supported for mons and OSDs only"
```
travisn, line 45:
```suggestion
- `debug` : [Debug a deployment](#debug-mode)  by scaling it down and creating a debug copy. This is supported for mons and OSDs only
```

README.md
```diff
@@ -42,6 +42,11 @@ To install the plugin, run:
...
+  - `start  <deployment-name> ` 
+    `[--alternate-image <alternate-image>]` : Start debugging a deployment with an optional alternative ceph container image"
```
travisn, line 47:
```suggestion
    `[--alternate-image <alternate-image>]` : Start debugging a deployment with an optional alternative ceph container image
```

parth-gr: Thanks:)

subhamkrai (approved):

BlaineEXE (approved):

PR #38, opened 2022-07-19
ci: fix ci issues

subhamkrai:
some commands where failing but debugging in
runner it was pasing so increasing the wait
time for operator to be in ready state.


tests/github-action-helper.sh
```diff
@@ -85,7 +85,7 @@ EOF
 }
 
 wait_for_operator_pod_to_be_ready_state_default() {
```
subhamkrai, line 87: @BlaineEXE any idea why GitHub action does not respect this method when calling from action yaml?
it works when calling from cli directly

travisn: Where is this called? I don't see it.

BlaineEXE: It may be a race condition. It takes a little bit of time from issuing the restart command until kubernetes begins terminating the pod. If we check for "Ready" before then, it might be the original pod that is still Ready.

The most complete way to verify it would be to get the name of the operator pod, then issue the restart command. After that, wait until the old pod is no longer present. Then wait for a new operator pod to be ready.

BlaineEXE (approved):

PR #39, opened 2022-07-28
Add reusable functions for kubectl in namespaces

BlaineEXE:
Add KUBECTL_NS_OPERATOR and KUBECTL_NS_CLUSTER functions to replace
repeated calls of `TOP_LEVEL_COMMAND --namespace "${ROOK_*_NAMESPACE}"`.

Use all caps for the function names so they stand out more clearly. When
using `kubectl_ns_operator some-command`, it isn't as clear that it is a
special type of call.

Signed-off-by: Blaine Gardner <blaine.gardner@redhat.com>

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -149,13 +149,23 @@ function end_of_command_parsing() {
   fi
 }
 
+# run a kubectl command in the operator namespace
+function KUBECTL_NS_OPERATOR() {
+  $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" "$@"
+}
+
+# run a kubectl command in the cluster namespace
+function KUBECTL_NS_CLUSTER() {
+  $TOP_LEVEL_COMMAND --namespace "$ROOK_CLUSTER_NAMESPACE" "$@"
+}
+
 ####################################################################################################
 # 'kubectl rook-ceph ceph ...' command
 ####################################################################################################
 
 function run_ceph_command() {
   # do not call end_of_command_parsing here because all remaining input is passed directly to 'ceph'
-  $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- ceph "$@" --conf="$CEPH_CONF_PATH"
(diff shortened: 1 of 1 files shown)


kubectl-rook-ceph.sh
```diff
@@ -230,20 +240,20 @@ function rook_version() {
...
 function run_rook_version() {
   end_of_command_parsing "$@" # end of command tree
-  $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" exec deploy/rook-ceph-operator -- rook version
+  KUBECTL_NS_OPERATOR exec deploy/rook-ceph-operator -- rook version
 }
 
 function run_rook_cr_status() {
   if [ "$#" -eq 1 ] && [ "$1" = "all" ]; then
-    cr_list=$($TOP_LEVEL_COMMAND --namespace "$ROOK_CLUSTER_NAMESPACE" get crd | awk '{print $1}' | sed '1d')
+    cr_list=$(KUBECTL_NS_CLUSTER get crd | awk '{print $1}' | sed '1d')
```
BlaineEXE, line 248: @parth-gr and @subhamkrai this is also an example of how a function can be used in the same way as a command.

If this were my project, I would probably have a lot of these smaller functions that repeat commonly-repeated commands for kubectl. For example, I could also have a function like below for executing commands in the operator.

```sh
function exec_in_operator() {
  KUBECTL_NS_OPERATOR exec deploy/rook-ceph-operator -- "$@"
}
```

Some reasons to do this:
- because code is reused, if I want to change how commands are issued, I can change fewer lines of code
- code might be easier to read
- the important parts of code lines (the command being executed) is less likely to be pushed off to the right side of the screen (easier to see)

Some reasons not to do this:
- while code could be easier to read, that is subjective; it could be harder to read if the function name isn't very clear

Do you think the benefits of this PR outweigh the drawbacks? Is this a pattern you would like to see used in the project? Since I see @subhamkrai as the primary maintainer of this utility, I think this is a good thing for him to decide. What do you think?

parth-gr: It looks good to me, by following this approach.

subhamkrai (approved):

kubectl-rook-ceph.sh
subhamkrai: Looks more readable and cleaner to me. Thanks!

Issue #32, opened 2022-05-24
Add command to analyze cluster health and warn about common health problems


travisn:
Users frequently have questions about getting their cluster setup properly. Some of the issues new users are repetitive and it would be amazing to provide a tool that gives guidance on how to fix some of these common issues. 

I'd propose a new command such as "kubectl rook-ceph health", which performs the following health checks and prints whether they are each healthy or if the checks failed:
- Three mon pods on different nodes
- All mons in quorum
- At least three OSDs on different nodes
- All rook pods in Running state, and none in CLBO
- PGs are all active+clean
- At least one mgr pod is running
- (advanced) If any pools have a failureDomain other than `host` or `osd`, confirm that the CRUSH map will support replica 3 across that topology

There are many other checks we can surely add, but would be great to start with a simple tool and add more checks over time.

PR #40, opened 2022-08-01
core: add health command to print cluster health

subhamkrai:
This commit adds new command `kubectl rook-ceph health`,
which check cluster health, by looking at
1. minimum 3 mons on different nodes
2. minimum 3 osds on different nodes
3. all mons are in quorum
4. all pods state
5. pg active+clean
6. atleast one mgr pod running

Closes: #32 

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -35,6 +35,7 @@ function print_usage() {
   echo "COMMANDS"
   echo "  ceph <args>                               : call a 'ceph' CLI command with arbitrary args"
   echo "  rbd <args>                                : call a 'rbd' CLI command with arbitrary args"
+  echo "  health                                    : check health of the cluster and common configuration issues"
   echo "  operator <subcommand>..."
   echo "    restart                                 : restart the Rook-Ceph operator"
   echo "    set <property> <value>                  : Set the property in the rook-ceph-operator-config configmap."
@@ -50,7 +51,7 @@ function print_usage() {
 
 function fail_error() {
   print_usage >&2
-  echo "ERROR: $*" >&2
+  error_msg " $*" >&2
   exit 1
 }
 
@@ -149,6 +150,18 @@ function end_of_command_parsing() {
   fi
 }
 
+function info_msg() {
+  echo -e "${INFO_PREFIX}" "$@"
+}
+
+function warn_msg() {
+  echo -e "${WARNING_PREFIX}" "$@"
+}
+
+function error_msg() {
+  echo -e "${ERROR_PREFIX}" "$@"
+}
+
 # run a kubectl command in the operator namespace
 function KUBECTL_NS_OPERATOR() {
   $TOP_LEVEL_COMMAND --namespace "$ROOK_OPERATOR_NAMESPACE" "$@"
@@ -219,7 +232,7 @@ function fetch_mon_endpoints() {
 ####################################################################################################
 
 function rook_version() {
-  [[ -z "${1:-""}" ]] && fail_error "Missing 'version' subcommand"
+  [[ -z "${1:-""}" ]] && fail_error "Missing subcommand"
   subcommand="$1"
   shift # remove the subcommand from the front of the arg list
   case "$subcommand" in
@@ -246,9 +259,9 @@ function run_rook_version() {
 function run_rook_cr_status() {
   if [ "$#" -eq 1 ] && [ "$1" = "all" ]; then
     cr_list=$(KUBECTL_NS_CLUSTER get crd | awk '{print $1}' | sed '1d')
-    echo "CR status"
+    info_msg " CR status"
     for cr in $cr_list; do
-      echo "$cr": "$(KUBECTL_NS_CLUSTER get "$cr" -ojson | jq --monochrome-output '.items[].status')"
+      echo -e "$cr": "$(KUBECTL_NS_CLUSTER get "$cr" -ojson | jq --monochrome-output '.items[].status')"
     done
   elif [[ "$#" -eq 1 ]]; then
     KUBECTL_NS_CLUSTER get "$1" -ojson | jq --monochrome-output '.items[].status'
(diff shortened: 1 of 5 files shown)


subhamkrai:
#32

subhamkrai:
you notice `${GREEN}, $ {RED}, ${YELLOW}`, these are just to add color in the command output to look fancier or also kind of give the user an impression of how good/bad cluster status is.

let me know if coloring makes sense. @BlaineEXE @travisn 

Screenshot from 2022-08-01 11-23-03

BlaineEXE:
I like the idea of coloring. We might want an option to disable it in the future, but let's leave that.

One thought I have is that we might want to specifically code outputs for users in case they want to script around this. Output 0 if everything is good, 1 if there are warnings, and 2 if there are errors.

subhamkrai:
> I like the idea of coloring. We might want an option to disable it in the future, but let's leave that.
> 
> One thought I have is that we might want to specifically code outputs for users in case they want to script around this. Output 0 if everything is good, 1 if there are warnings, and 2 if there are errors.

I think with new changes, we might not require, we have small coloring like `Info, Warning, Error` not the completed sentence as color, something like what kubectl has.

FYI, this coloring is not added for all the commands.

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,91 @@ function run_purge_osd() {
...
+  echo
+  check_all_pods_status
+  echo
+  check_pg_are_active_clean
+  echo
+  check_mgr_pods_status_and_count
+}
+
+function check_mon_pods_nodes() {
+  echo -e "${INFO}Checking if atleast three mon pods are running on different nodes"
```
travisn, line 294: "at least" should be two words throughout
```suggestion
  echo -e "${INFO}Checking if at least three mon pods are running on different nodes"
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,91 @@ function run_purge_osd() {
...
+function check_all_pods_status() {
+  echo -e "${INFO}Pods that are in 'Running' status"
+  KUBECTL_NS_OPERATOR get pod --field-selector status.phase=Running
+  if [ "$ROOK_OPERATOR_NAMESPACE" != "$ROOK_CLUSTER_NAMESPACE" ]; then
+    KUBECTL_NS_CLUSTER get pod --field-selector status.phase=Running
+  fi
+
+  echo
+  echo -e "${WARNING}Pods that are 'Not' in 'Running' status"
+  set +e # if there are no pod that are not running, then do no exit
```
travisn, line 332: Perhaps we `set +e` for all commands that are printing status (as part of `cluster_health()` function above)? If a cluster is unhealthy, we want to print as much status as possible even if some commands fail.

kubectl-rook-ceph.sh
```diff
@@ -344,6 +433,10 @@ function run_main_command() {
 : "${ROOK_CLUSTER_NAMESPACE:=rook-ceph}"
 : "${ROOK_OPERATOR_NAMESPACE:=$ROOK_CLUSTER_NAMESPACE}"
 : "${TOP_LEVEL_COMMAND:=kubectl}"
+: "${RESET:=\033[0m}"                    # For no color
+: "${ERROR:=\033[1;31mError $RESET}"     # \033[1;31m for Red color
```
travisn, line 437: How does it look with a colon? Or maybe it's not necessary
```suggestion
: "${ERROR:=\033[1;31mError: $RESET}"     # \033[1;31m for Red color
```

kubectl-rook-ceph.sh
```diff
@@ -344,6 +433,10 @@ function run_main_command() {
...
+: "${INFO:=\033[1;32mInfo $RESET}"       # \033[1;32m for Green color
+: "${WARNING:=\033[1;33mWarning $RESET}" # \033[1;33m for Yellow color
```
travisn, line 439: How about we name them ERROR_PREFIX, INFO_PREFIX, and WARNING_PREFIX? 
```suggestion
: "${WARNING_PREFIX:=\033[1;33mWarning $RESET}" # \033[1;33m for Yellow color
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,91 @@ function run_purge_osd() {
...
+  else
+    echo "${WARNING}$pg_state"
+  fi
+}
+
+function check_mgr_pods_status_and_count() {
+  echo -e "${INFO}checking if atleast one mgr pod is running"
+  mgr_pod_count=$(KUBECTL_NS_CLUSTER get pod -o=custom-columns=NAME:.metadata.name,STATUS:.status.phase | grep mgr | awk '{print $2}' | sort | uniq | wc -l)
+  if [ "$mgr_pod_count" -lt 1 ]; then
+    echo -e "${ERROR}Atleast one mgr pods should running"
```
travisn, line 357:
```suggestion
    echo -e "${ERROR}At least one mgr pod should running"
```

travisn (changes requested):
Looks great, just a few small suggestions. Let's also discuss some of the error conditions that could be tested manually.

kubectl-rook-ceph.sh
BlaineEXE: It might be good to make info blue instead of green. Green usually indicates a "good" or "correct" state.

kubectl-rook-ceph.sh
```diff
@@ -331,11 +417,14 @@ function run_main_command() {
   rook)
     rook_version "$@"
     ;;
+  health)
+    cluster_health "$@"
```
BlaineEXE, line 421: When I wrote the skeleton, I intended functions that are running commands to start with `run_`. You don't have to follow this convention, but it can help visually identify core functions versus helpers as the file grows large.

```suggestion
    run_cluster_health "$@"
```

kubectl-rook-ceph.sh
BlaineEXE: `set +e` should set it for the whole script as well. We can do `set +e` in the `run_cluster_health()` function and then `set -e` at the end to turn off error-ing for all helper funcions.

kubectl-rook-ceph.sh
```diff
@@ -35,6 +35,7 @@ function print_usage() {
   echo "COMMANDS"
   echo "  ceph <args>                               : call a 'ceph' CLI command with arbitrary args"
   echo "  rbd <args>                                : call a 'rbd' CLI command with arbitrary args"
+  echo "  health                                    : print cluster health status"
```
BlaineEXE, line 38: IMO, a more descriptive help message would be useful.

For example: check health of the cluster and common configuration issues

kubectl-rook-ceph.sh
```diff
@@ -123,19 +124,19 @@ function parse_flags() {
       break
       ;;
     esac
-    is_flag "$VAL" && fail_error "Flag '$FLAG' value '$VAL' looks like another flag"
+    is_flag "$VAL" && fail_error "${ERROR}Flag '$FLAG' value '$VAL' looks like another flag"
```
BlaineEXE, line 127: Why not change `fail_error` to print the error format instead of having to add `${ERROR}` to everything?

BlaineEXE (changes requested):

kubectl-rook-ceph.sh
subhamkrai: changed

kubectl-rook-ceph.sh
subhamkrai: done

kubectl-rook-ceph.sh
```diff
@@ -344,6 +432,10 @@ function run_main_command() {
 : "${ROOK_CLUSTER_NAMESPACE:=rook-ceph}"
 : "${ROOK_OPERATOR_NAMESPACE:=$ROOK_CLUSTER_NAMESPACE}"
 : "${TOP_LEVEL_COMMAND:=kubectl}"
+: "${RESET:=\033[0m}"                            # For no color
+: "${ERROR_PREFIX:=\033[1;31mError: $RESET}"     # \033[1;31m for Red color
```
travisn, line 436: nit: I think the script will be more easily readable if the space is not here, but is instead added where the prefix is used. (similar change for warning and info)
```suggestion
: "${ERROR_PREFIX:=\033[1;31mError:$RESET}"     # \033[1;31m for Red color
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,90 @@ function run_purge_osd() {
...
+  else
+    echo -e "${WARNING_PREFIX}$pg_state"
+  fi
+}
+
+function check_mgr_pods_status_and_count() {
+  echo -e "${INFO_PREFIX}checking if at least one mgr pod is running"
+  mgr_pod_count=$(KUBECTL_NS_CLUSTER get pod -o=custom-columns=NAME:.metadata.name,STATUS:.status.phase | grep mgr | awk '{print $2}' | sort | uniq | wc -l)
+  if [ "$mgr_pod_count" -lt 1 ]; then
+    echo -e "${ERROR_PREFIX}At least one mgr pod should running"
```
travisn, line 356: For readability, to go with the other space suggestion:
```suggestion
    echo -e "${ERROR_PREFIX} At least one mgr pod should running"
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,90 @@ function run_purge_osd() {
...
+    echo -e "$ceph_health_details"
+  elif [[ "$ceph_health_details" =~ "HEALTH_WARN" ]]; then
+    echo -e "${WARNING_PREFIX}$ceph_health_details"
+  elif [[ "$ceph_health_details" =~ "HEALTH_ERR" ]]; then
+    echo -e "${ERROR_PREFIX}$ceph_health_details"
+  fi
+}
+
+function check_osd_pod_count_and_nodes() {
+  echo -e "${INFO_PREFIX}Checking if atleast three osd pods are running on different nodes"
```
travisn, line 316:
```suggestion
  echo -e "${INFO_PREFIX}Checking if at least three osd pods are running on different nodes"
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,90 @@ function run_purge_osd() {
...
+  osd_unique_node_count=$(KUBECTL_NS_CLUSTER get pod -o=custom-columns=NAME:.metadata.name,NODE:.spec.nodeName | grep osd | grep -v prepare | awk '{print $2}' | sort | uniq | wc -l)
+  if [ "$osd_unique_node_count" -lt 3 ]; then
+    echo -e "${WARNING_PREFIX}Atleast three osd pods should running on different node"
```
travisn, line 319:
```suggestion
    echo -e "${WARNING_PREFIX}At least three osd pods should running on different nodes"
```

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,90 @@ function run_purge_osd() {
...
+  echo
+  check_mgr_pods_status_and_count
+  set -e
+}
+
+function check_mon_pods_nodes() {
+  echo -e "${INFO_PREFIX}Checking if at least three mon pods are running on different nodes"
+  mon_unique_node_count=$(KUBECTL_NS_CLUSTER get pod -o=custom-columns=NAME:.metadata.name,NODE:.spec.nodeName | grep mon | grep -v canary | awk '{print $2}' | sort | uniq | wc -l)
+  if [ "$mon_unique_node_count" -lt 3 ]; then
+    echo -e "${WARNING_PREFIX}At least three mon pods should running on different node"
```
travisn, line 298:
```suggestion
    echo -e "${WARNING_PREFIX}At least three mon pods should running on different nodes"
```

travisn (changes requested):
A few nits...

kubectl-rook-ceph.sh
subhamkrai: agreed

tests/github-action-helper.sh
```diff
@@ -32,6 +32,7 @@ deploy_rook() {
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/operator.yaml
   curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/cluster-test.yaml -o cluster-test.yaml
   sed -i "s|#deviceFilter:|deviceFilter: $(lsblk | awk '/14G/ {print $1}' | head -1)|g" cluster-test.yaml
+  sed -i '34s/1/3/' cluster-test.yaml
```
travisn, line 35: The line number could change in the yaml. Instead could you replace the first instance of `count: 1`?

travisn (approved):

kubectl-rook-ceph.sh
```diff
@@ -273,6 +274,90 @@ function run_purge_osd() {
...
+
+function check_all_pods_status() {
+  echo -e "${INFO_PREFIX} Pods that are in 'Running' status"
+  KUBECTL_NS_OPERATOR get pod --field-selector status.phase=Running
+  if [ "$ROOK_OPERATOR_NAMESPACE" != "$ROOK_CLUSTER_NAMESPACE" ]; then
+    KUBECTL_NS_CLUSTER get pod --field-selector status.phase=Running
+  fi
+
+  echo
+  echo -e "${WARNING_PREFIX} Pods that are 'Not' in 'Running' status"
```
BlaineEXE, line 332: I see a lot of this `echo -e "${SOME_PREFIX} text".

Generally, I have used helpers for things like this

```sh
function info_msg() {
  echo -e "${INFO_PREFIX}""$@"
}

function warn_msg() {
  echo -e "${WARN_PREFIX}""$@"
}

function error_msg() {
  echo -e "${ERROR_PREFIX}""$@"
}
```


Then these messages are called like

```sh
info_msg "checking placement group status"

warn_msg "$pg_state"

error_msg "$pg_state"
```

subhamkrai: done

subhamkrai: thanks for the suggestion

kubectl-rook-ceph.sh
```diff
@@ -246,9 +247,9 @@ function run_rook_version() {
 function run_rook_cr_status() {
   if [ "$#" -eq 1 ] && [ "$1" = "all" ]; then
     cr_list=$(KUBECTL_NS_CLUSTER get crd | awk '{print $1}' | sed '1d')
-    echo "CR status"
+    echo -e "${INFO_PREFIX}CR status"
```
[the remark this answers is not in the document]
subhamkrai: done

BlaineEXE (approved):

PR #46, opened 2022-08-19
ci: run command according to krew latest releases

subhamkrai:

### DIFF
--- a/.github/workflows/ci-for-default-ns.yaml
+++ b/.github/workflows/ci-for-default-ns.yaml
@@ -28,10 +28,44 @@ jobs:
         run: |
           export PATH="${PATH}:${HOME}/.krew/bin"
           # run ceph commands with the krew plugin
-          kubectl rook-ceph ceph status
+           kubectl rook-ceph ceph status
           kubectl rook-ceph ceph status -f json
           kubectl rook-ceph ceph status --format json-pretty
-          kubectl rook-ceph ceph mon stat
+          POD=$(kubectl -n rook-ceph get pod -l app=rook-ceph-operator -o jsonpath="{.items[0].metadata.name}")
+          kubectl rook-ceph operator restart
+
+          # let's wait for operator pod to be restart
+          kubectl -n rook-ceph wait --for=delete pod/$POD --timeout=100s
+          tests/github-action-helper.sh wait_for_operator_pod_to_be_ready_state_default
+          kubectl rook-ceph operator set ROOK_LOG_LEVEL DEBUG
+          kubectl rook-ceph --context=$(kubectl config current-context) mons
+          kubectl rook-ceph rook version
+          kubectl rook-ceph rook status
+          kubectl rook-ceph rook status all
+          kubectl rook-ceph rook status cephobjectstores
+          # to allow time for reconcile, sleep before listing the pools
+          sleep 10
+          kubectl rook-ceph rbd ls replicapool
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-for-default-ns.yaml
```diff
@@ -28,10 +28,44 @@ jobs:
...
           kubectl rook-ceph ceph status --format json-pretty
-          kubectl rook-ceph ceph mon stat
+          POD=$(kubectl -n rook-ceph get pod -l app=rook-ceph-operator -o jsonpath="{.items[0].metadata.name}")
+          kubectl rook-ceph operator restart
+
+          # let's wait for operator pod to be restart
+          kubectl -n rook-ceph wait --for=delete pod/$POD --timeout=100s
+          tests/github-action-helper.sh wait_for_operator_pod_to_be_ready_state_default
+          kubectl rook-ceph operator set ROOK_LOG_LEVEL DEBUG
+          kubectl rook-ceph --context=$(kubectl config current-context) mons
```
travisn, line 41: Why do we need the kubectl context here and not for other commands?

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -25,48 +23,73 @@ jobs:
           op-ns: "test-operator"
           cluster-ns: "test-cluster"
 
-      - name: install script
-        run: sudo install kubectl-rook-ceph.sh /usr/local/bin/kubectl-rook_ceph
+      - name: install krew
+        run: tests/github-action-helper.sh install_krew
 
-      - name: Test Plugin
+      - name: Test pluging with ceph commands
```
travisn, line 29:
```suggestion
      - name: Test plugin with ceph commands
```

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -25,48 +23,73 @@ jobs:
...
           # osd
-          kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-osd-0
+          kubectl rook-ceph -o test-operator -n test-cluster debug start rook-ceph-osd-0
           # sleep till the debug pod get running
           sleep 5
-          kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
+          kubectl rook-ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
+  # This test is required to test latest changes or the changes that not present
+  # with current version of rook-ceph krew plugin
+  with-pr-changes-in-custom-namespace:
```
travisn, line 72: This is a difficult pattern to maintain if every new feature is added to this section, then it merges, then included in a release, then we need a separate PR to move it to the other section. What if we just leave everything in the CI in the non-official krew commands using `rook_ceph`, and then no need to update the CI later after new features are added? Or is there a benefit of converting the CI like this other than using the exact syntax as the docs?

travisn (changes requested):

.github/workflows/ci-for-diff-ns.yaml
subhamkrai: no, I don't think we have a benefit here. So, should we close this pr or any other changes you are suggesting?

travisn: Yeah how about if we close this PR and leave the tests with their current pattern

PR #47, opened 2022-08-19
ci: add codespell in ci

subhamkrai:

### DIFF
--- a/.github/workflows/codespell.yaml
+++ b/.github/workflows/codespell.yaml
@@ -1,0 +1,29 @@
+name: Codespell
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - release-*
+  pull_request:
+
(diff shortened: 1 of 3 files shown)


travisn (approved):

.github/workflows/codespell.yaml
```diff
@@ -0,0 +1,35 @@
+name: Codespell
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - release-*
+  pull_request:
```
travisn, line 9: Shouldn't the codespell action show up on this PR? I don't see it

travisn (changes requested):

.github/workflows/codespell.yaml
subhamkrai: running

travisn (approved):

Issue #42, opened 2022-08-03
Instead of random wait time in CI, read the resource details using Timeout


parth-gr:
```
+ kubectl rook_ceph -o test-operator -n test-cluster rbd ls replicapool
rbd: error opening pool 'replicapool': (2) No such file or directory
rbd: listing images failed: (2) No such file or directory
command terminated with exit code 2
Error: Process completed with exit code 2.
```

The CI failed because the wait time was not enough to reconcile the operator completely,

Instead of adding random wait time, we can read the resource until it becomes ready using Timeout,

LIke : https://github.com/rook/rook/blob/ec0c9a7963357fcd9023fc8d7c28720fe6f3c810/.github/workflows/canary-integration-test.yml#L59

subhamkrai:
after this fix #38 I have not seen a CI issue, before this, I faced but not once was this merged.
try rebasing.

parth-gr:
@subhamkrai You can see my PR is already on top of it https://github.com/rook/kubectl-rook-ceph/pull/36

Still, the CI fails with this error,

And anyhow using Timeout would be a good practice in time, when the CI would be growing more with commands:)

subhamkrai:
right, checking for resource is better
Feel free to raise the pr

Issue #41, opened 2022-08-01
Not indented correctly


parth-gr:
https://github.com/rook/kubectl-rook-ceph/blob/ef787842a5e73ecac7687cb7706254e574b44d9e/.github/workflows/ci-for-diff-ns.yaml#L56

This should be `kubectl rook_ceph ...`

parth-gr:
ALso need to investigate in the CI why it was passing even with this bug.

parth-gr:
Also somewhere there should be mentioned that `flags` should be only used before the arguments and commands.
For example:           kubectl rook_ceph -o test-operator -n test-cluster ceph status

subhamkrai:
> ALso need to investigate in the CI why it was passing even with this bug.

this is not a bug, kubectl-rook_ceph or kubectl rook_ceph behaves the same since we install kubectl-rook_ceph in the $PATH
and kubectl treat it as plugin

parth-gr:
> > ALso need to investigate in the CI why it was passing even with this bug.
> 
> this is not a bug, kubectl-rook_ceph or kubectl rook_ceph behaves the same since we install kubectl-rook_ceph in the $PATH and kubectl treat it as plugin

Any specific reason to install this too kubectl-rook_ceph

subhamkrai:
> > > ALso need to investigate in the CI why it was passing even with this bug.
> > 
> > 
> > this is not a bug, kubectl-rook_ceph or kubectl rook_ceph behaves the same since we install kubectl-rook_ceph in the $PATH and kubectl treat it as plugin
> 
> Any specific reason to install this too kubectl-rook_ceph

yes, to accept the tool as kubectl plugin it should have prefix of `kubectl-`

subhamkrai:
@parth-gr can we close this?

PR #48, opened 2022-08-22
ci: added timeout instead of sleep commands

parth-gr:
Instead of random sleep time, added timeout
so we will wait till the pods get started

Closes: https://github.com/rook/kubectl-rook-ceph/issues/42

### DIFF
--- a/.github/workflows/ci-for-default-ns.yaml
+++ b/.github/workflows/ci-for-default-ns.yaml
@@ -75,15 +75,15 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          tkubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug rook-ceph
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug rook-ceph
           kubectl rook_ceph debug stop rook-ceph-mon-a
           # osd
           kubectl rook_ceph debug start rook-ceph-osd-0
-          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug rook-ceph
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug rook-ceph
           kubectl rook_ceph debug stop rook-ceph-osd-0
 
            # check if osd 0 is started
-          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0 rook-ceph
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0 rook-ceph
 
           # for testing osd purge scale the osd deplyment
           kubectl --namespace rook-ceph scale deploy/rook-ceph-osd-0 --replicas=0
--- a/.github/workflows/ci-for-default-ns.yaml
+++ b/.github/workflows/ci-for-default-ns.yaml
@@ -75,15 +75,15 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          tkubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug rook-ceph
           kubectl rook_ceph debug stop rook-ceph-mon-a
           # osd
           kubectl rook_ceph debug start rook-ceph-osd-0
-          sleep 5
+          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug rook-ceph
           kubectl rook_ceph debug stop rook-ceph-osd-0
 
-          # sleep again so OSD can start
-          sleep 5
+           # check if osd 0 is started
+          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0 rook-ceph
 
           # for testing osd purge scale the osd deplyment
           kubectl --namespace rook-ceph scale deploy/rook-ceph-osd-0 --replicas=0
--- a/.github/workflows/ci-for-diff-ns.yaml
+++ b/.github/workflows/ci-for-diff-ns.yaml
@@ -57,6 +57,7 @@ jobs:
           tests/github-action-helper.sh wait_for_three_mons test-cluster
           kubectl -n test-cluster wait deployment rook-ceph-mon-d --for condition=Available=True --timeout=90s
           kubectl -n test-cluster wait deployment rook-ceph-mon-e --for condition=Available=True --timeout=90s
+<<<<<<< HEAD
 
           # for testing osd purge scale the osd deplyment
           kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
@@ -64,18 +65,29 @@ jobs:
           sleep 5
           kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
           kubectl rook_ceph -o test-operator -n test-cluster health
+=======
+>>>>>>> 3007d8d (ci: added timeout instead of sleep commands)
 
           # for testing start-debug and stop-debug deployment
           # mon
           kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-mon-a
-          # sleep till the debug pod get running
-          sleep 5
+          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug test-cluster
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-mon-a
           # osd
           kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-osd-0
-          # sleep till the debug pod get running
-          sleep 5
+          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug test-cluster
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
+
+           # check if osd 0 is started
+          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0 test-cluster
+        
+          # for testing osd purge scale the osd deplyment
+          kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
+          # we need to sleep so the osd will be marked down before purging the osd
+          sleep 5
+          kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
+          kubectl rook_ceph -o test-operator -n test-cluster health
+
 
       - name: setup tmate session for debugging when event is PR
         if: failure() && github.event_name == 'pull_request'
--- a/.github/workflows/ci-for-diff-ns.yaml
+++ b/.github/workflows/ci-for-diff-ns.yaml
@@ -57,29 +57,19 @@ jobs:
           tests/github-action-helper.sh wait_for_three_mons test-cluster
           kubectl -n test-cluster wait deployment rook-ceph-mon-d --for condition=Available=True --timeout=90s
           kubectl -n test-cluster wait deployment rook-ceph-mon-e --for condition=Available=True --timeout=90s
-<<<<<<< HEAD
-
-          # for testing osd purge scale the osd deplyment
-          kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
-          # we need to sleep so the osd will be marked down before purging the osd
-          sleep 5
-          kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
-          kubectl rook_ceph -o test-operator -n test-cluster health
-=======
->>>>>>> 3007d8d (ci: added timeout instead of sleep commands)
 
           # for testing start-debug and stop-debug deployment
           # mon
           kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-mon-a
-          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug test-cluster
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug test-cluster
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-mon-a
           # osd
           kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-osd-0
-          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug test-cluster
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug test-cluster
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
 
            # check if osd 0 is started
-          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0 test-cluster
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0 test-cluster
         
           # for testing osd purge scale the osd deplyment
           kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -244,7 +244,7 @@ function fetch_mon_endpoints() {
   KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o json | jq --monochrome-output '.data.data' | tr -d '"' | tr -d '=' | sed 's/[A-Za-z]*//g'
 }
 
-wait_for_deployment_to_be_running() {
+function wait_for_deployment_to_be_running() {
   deployment=$1
   info_msg "Waiting for the pod from deployment \"$deployment\" to be running"
   KUBECTL_NS_CLUSTER wait deployment $deployment --for condition=Available=True --timeout=90s
--- a/tests/github-action-helper.sh
+++ b/tests/github-action-helper.sh
@@ -118,6 +118,13 @@ wait_for_three_mons() {
 EOF
 }
 
+wait_for_deployment_to_be_running() {
+  deployment=$1
+  namespace=$2
+  echo "Waiting for the pod from deployment \"$deployment\" to be running"
+  kubectl -n "$namespace" wait deployment "$deployment" --for condition=Available=True --timeout=90s
+}
+
 ########
 # MAIN #
 ########


subhamkrai:
> @subhamkrai by this
> 
> https://github.com/rook/kubectl-rook-ceph/blob/30cf64691b7104dbbe48e108629e5302ad92866c/.github/workflows/ci-for-diff-ns.yaml#L49
> 
> what we are exactly waiting for
> So I can add this in timeout

added sleep to so that blocks pools are ready before listing them

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -49,24 +49,49 @@ jobs:
...
+            if [[ $val == *"debug"* ]];
+            then
+              debug_osd=$val
+            fi  
+          done
+          timeout 60 sh -c 'until [ "$(kubectl get pod $debug_osd -n test-cluster -o jsonpath='{.status.phase}'| grep -c "Running")" -eq 1 ]; do echo "waiting for pod rook-ceph-osd-0-debug to get created " && sleep 1; done'
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
+          
+          # check if osd 0 is started
+          osds=$(kubectl get pod -l app=rook-ceph-osd -n test-cluster -o jsonpath='{.items[*].metadata.name}'| grep 0)
```
travisn, line 79: There is a label for the osd id you can query, but you better confirm the label name, i'm not sure if this is correct
```suggestion
          osds=$(kubectl get pod -l app=rook-ceph-osd,osd-id=0 -n test-cluster -o jsonpath='{.items[*].metadata.name}'| grep 0)
```

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,16 +75,23 @@ jobs:
...
+            then
+              debug_mon=$val
+            fi  
+          done
+          timeout 60 sh -c 'until [ "$(kubectl get pod $debug_mon -n rook-ceph -o jsonpath='{.status.phase}'| grep -c "Running")" -eq 1 ]; do echo "waiting for pod rook-ceph-mon-a-debug to get created " && sleep 1; done'
           kubectl rook_ceph debug stop rook-ceph-mon-a
           # osd  
           kubectl rook_ceph debug start rook-ceph-osd-0
-          sleep 5
+          tests/github-action-helper.sh wait_for_operator_pod_to_be_ready_state_default
```
travisn, line 89: This helper waits for the operator pod, but don't we need to wait for the osd debug pod?

travisn (changes requested):

.github/workflows/ci-for-default-ns.yaml
parth-gr: Yes added

.github/workflows/ci-for-diff-ns.yaml
parth-gr: Okay, thanks

travisn (commented):
LGTM, there is just a merge conflict to resolve

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,21 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          sleep 10
+          debug_mon=$(kubectl get pod -l app=rook-ceph-mon,app.kubernetes.io/instance=a -n rook-ceph -o jsonpath='{.items[*].metadata.name}')
+          timeout 1000 sh -c 'until [ "$(kubectl get pod $debug_mon -n rook-ceph -o jsonpath='{.status.phase}'| grep -c "Running")" -eq 1 ]; do echo "waiting for pod rook-ceph-mon-a-debug to get created " && sleep 1; done'
```
parth-gr, line 80: @travisn the CI is again and again failing here,

I am not sure why this is happeening, 
I increased the timeout to 1000s still it didn't came into ready state

travisn: Did you confirm the pod is correctly retrieved on line 79? Maybe the pod is not being found. For debugging i'd suggest breaking these long lines apart and print intermediate results.

parth-gr: Logs

```
deployment.apps/rook-ceph-mon-a-debug created
+ sleep 10
++ kubectl get pod -l app=rook-ceph-mon,app.kubernetes.io/instance=a -n rook-ceph -o 'jsonpath={.items[*].metadata.name}'
+ debug_mon=rook-ceph-mon-a-debug-58774f6fb6-xjnp6
+ timeout 1000 sh -c 'until [ "$(kubectl get pod $debug_mon -n rook-ceph -o jsonpath={.status.phase}| grep -c "Running")" -eq 1 ]; do echo "waiting for pod rook-ceph-mon-a-debug to get created " && sleep 1; done'
waiting for pod rook-ceph-mon-a-debug to get created 
waiting for pod rook-ceph-mon-a-debug to get created 
```


Yes it got created and fetched properly,

In timeout I runn the same command it run fine

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,21 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          sleep 10
+          debug_mon=$(kubectl get pod -l app=rook-ceph-mon,app.kubernetes.io/instance=a -n rook-ceph -o jsonpath='{.items[*].metadata.name}')
```
BlaineEXE, line 79: What's to stop this line from returning the old pod if it doesn't terminate immediately? Often, pods take a while to terminate. Instead, can we select just the debug pod? The name of the deployment is deterministic (`rook-ceph-mon-a-debug` IIRC) also, so we should be able to use that to our advantage.

.github/workflows/ci-for-default-ns.yaml
BlaineEXE, line 80: Why run `sh` instead of `bash`?

.github/workflows/ci-for-default-ns.yaml
BlaineEXE: A reusable function for waiting for a deployment's pod to be running is probably a good idea. I see this overall structure 3 times in this PR.

I use this script for my development environment tooling.

`wait_for`:
```sh
#!/usr/bin/env bash
set -eEuo pipefail

REASON="$1"  # reason
TIMEOUT="$2"  # in seconds
CMD="$3"

BASH_CMD="${BASH_CMD:=bash}" # might be 'bash -x', for example

msg="Waiting ${TIMEOUT} seconds for ${REASON} ..."
echo "${msg} countdown: ${TIMEOUT}"
start_time=$SECONDS
until ${BASH_CMD} -c "${CMD}" 2>&1 ; do
  echo ""
  if (( SECONDS - start_time >= TIMEOUT )); then
    echo "${msg} timed out! (output above)"
    exit 1
  fi
  echo "${msg} countdown: $((TIMEOUT - SECONDS + start_time)) "
  sleep 5
done
echo "${msg} done after $((SECONDS - start_time)) seconds."
```

You might use the function I gave like this:
```sh
# allow shorthand wait_for instead of full path
function wait_for() { ./script/dir/wait_for "$@"; }

wait_for "rook-ceph-mon-a-deploy pod to be ready" 60 "kubectl -n rook-ceph get pod | grep `rook-ceph-mon-a-deploy.*Ready'"
```

Or with a conditional like below, but I don't think you'll need this form:
```sh
./script/dir/wait_for "OSD provisioning to be complete" 900 "[[ \$(kubectl get -n rook-ceph pods 2>&1 | grep -c 'rook-ceph-osd-prepare.*Completed') -eq $num_osd_nodes ]]"
```

.github/workflows/ci-for-default-ns.yaml
parth-gr: Changed

.github/workflows/ci-for-default-ns.yaml
parth-gr: Changed grep `rook-ceph-mon-a-debug` directly

parth-gr: > A reusable function for waiting for a deployment's pod to be running is probably a good idea. I see this overall structure 3 times in this PR.
> 
> I use this script for my development environment tooling.
> 
> `wait_for`:
> 
> ```shell
> #!/usr/bin/env bash
> set -eEuo pipefail
> 
> REASON="$1"  # reason
> TIMEOUT="$2"  # in seconds
> CMD="$3"
> 
> BASH_CMD="${BASH_CMD:=bash}" # might be 'bash -x', for example
> 
> msg="Waiting ${TIMEOUT} seconds for ${REASON} ..."
> echo "${msg} countdown: ${TIMEOUT}"
> start_time=$SECONDS
> until ${BASH_CMD} -c "${CMD}" 2>&1 ; do
>   echo ""
>   if (( SECONDS - start_time >= TIMEOUT )); then
>     echo "${msg} timed out! (output above)"
>     exit 1
>   fi
>   echo "${msg} countdown: $((TIMEOUT - SECONDS + start_time)) "
>   sleep 5
> done
> echo "${msg} done after $((SECONDS - start_time)) seconds."
> ```
> 
> You might use the function I gave like this:
> 
> ```shell
> # allow shorthand wait_for instead of full path
> function wait_for() { ./script/dir/wait_for "$@"; }
> 
> wait_for "rook-ceph-mon-a-deploy pod to be ready" 60 \
>   "kubectl -n rook-ceph get pod | grep `rook-ceph-mon-a-deploy.*Ready'"
> ```
> 
> Or with a conditional like below, but I don't think you'll need this form:
> 
> ```shell
> ./script/dir/wait_for "OSD provisioning to be complete" 900 \
>   "[[ \$(kubectl get -n rook-ceph pods 2>&1 | grep -c 'rook-ceph-osd-prepare.*Completed') -eq $num_osd_nodes ]]"
> ```

There is only one-liner 
`   timeout 10 bash -c 'until [ "$(kubectl get pod $debug_osd -n test-cluster -o jsonpath='{.status.phase}'| grep -c "Running")" -eq 1 ]; do echo "waiting for pod rook-ceph-osd-0-debug to get created " && sleep 1; done'
`

The Above script gives a lot of information I would propose to add it in the rook CI as it gives us a lot of information

For a CI like Krew which is simpler and easy to go, I don't think we need such information for debugging.

Anyway, If you say I would be happy to add in Krew CI too. @BlaineEXE @subhamkrai thoughts?

BlaineEXE: That's a pretty complicated one-liner. IMO, a reusable function is still good. When reading the code, you can see the intent without getting distracted by the boilerplate, and there will be less chance for errors in that boilerplate.

BlaineEXE: > What's to stop this line from returning the old pod if it doesn't terminate immediately? Often, pods take a while to terminate. Instead, can we select just the debug pod? The name of the deployment is deterministic (`rook-ceph-mon-a-debug` IIRC) also, so we should be able to use that to our advantage.

Regarding this, it also makes me think that maybe we should wait for the old pod to be gone before creating the debug deployment. While a pod is terminating, we have to assume the OSD is still running, and if a user quickly starts into the debug pod and starts performing operations on an OSD while the OSD process is running, it could corrupt the disk. @travisn thoughts?

subhamkrai: I agree, we should first wait for older pod to remove before waiting for new one to start as pod takes some time to terminating and running again.

.github/workflows/a.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+echo "Running" | grep -c "Running"
```
travisn, line 3: Is this script used, or did you intend to revert it?

parth-gr: Added for testing, not needed.

wait_for.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
subhamkrai, line 1: @parth-gr can we move this `tests/github_action_helper` file as a function?

parth-gr: DOne

travisn:
Looks like the CI is stuck waiting

parth-gr:
> Looks like the CI is stuck waiting

There still random timeout for osd purge which stuck sometimes.

tests/github-action-helper.sh
```diff
@@ -104,6 +104,28 @@ wait_for_operator_pod_to_be_ready_state_custom() {
 EOF
 }
 
+wait_for_command_to_get_success() {
```
BlaineEXE, line 107: wait_for_command_to_get_success seems unnecessarily long to me. IMO, wait_for_condition is still clear and not as long.

```suggestion
wait_for_condition() {
```

parth-gr: yup

parth-gr:
@BlaineEXE @travisn @subhamkrai ^^

subhamkrai (commented):
@parth-gr can you try restarting CI multiple times and see how quickly they or there are getting stuck? since I started CI it is trying to run the commands for 8 min.

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,24 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          sleep 10
```
subhamkrai, line 78:
```suggestion
```
why use sleep again, the goal of the pr is to remove the sleep?

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,24 @@ jobs:
...
+          debug_mon=$(kubectl get pods -nrook-ceph | grep rook-ceph-mon-a-debug | awk '{ print $1  }')          
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-mon-a-deploy deployment pod to be ready" 10 \
+            "kubectl -n rook-ceph get pod $debug_mon | grep 'rook-ceph-mon-a-debug-.*Running'"
           kubectl rook_ceph debug stop rook-ceph-mon-a
           # osd
           kubectl rook_ceph debug start rook-ceph-osd-0
-          sleep 5
+          sleep 10
```
subhamkrai, line 85:
```suggestion
```

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,24 @@ jobs:
...
+          debug_osd=$(kubectl get pod -n rook-ceph | grep rook-ceph-osd-0-debug | awk '{ print $1  }')
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-osd-0-deploy deployment pod to be ready" 10 \
+            "kubectl -n rook-ceph get pod $debug_osd | grep 'rook-ceph-osd-0-debug-.*Running'"
           kubectl rook_ceph debug stop rook-ceph-osd-0
 
-          # sleep again so OSD can start
-          sleep 5
+          # check if osd 0 is started
+          sleep 10
```
subhamkrai, line 92:
```suggestion
```

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -49,24 +49,35 @@ jobs:
...
-          sleep 5
-          kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
-          kubectl rook_ceph -o test-operator -n test-cluster health
-
           # for testing start-debug and stop-debug deployment
           # mon
           kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-mon-a
-          # sleep till the debug pod get running
-          sleep 5
+          sleep 10
```
subhamkrai, line 55:
```suggestion
```

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -49,24 +49,35 @@ jobs:
...
+          debug_mon=$(kubectl get pods -ntest-cluster | grep rook-ceph-mon-a-debug | awk '{ print $1  }') 
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-mon-a-deploy deployment pod to be ready" 10 \
+            "kubectl -n test-cluster get pod $debug_mon | grep 'rook-ceph-mon-a-debug-.*Running'"
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-mon-a
           # osd
           kubectl rook_ceph -o test-operator -n test-cluster debug start rook-ceph-osd-0
-          # sleep till the debug pod get running
-          sleep 5
+          sleep 10
```
subhamkrai, line 62:
```suggestion
```

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -49,24 +49,35 @@ jobs:
...
+          debug_osd=$(kubectl get pod -n test-cluster | grep rook-ceph-osd-0-debug | awk '{ print $1  }')
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-osd-0-deploy deployment pod to be ready" 10 \
+            "kubectl -n test-cluster get pod $debug_osd | grep 'rook-ceph-osd-0-debug-.*Running'"
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
+          
+          # check if osd 0 is started
+          sleep 10
```
subhamkrai, line 69:
```suggestion
```

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -49,24 +49,35 @@ jobs:
...
+          osd=$(kubectl get pod -n test-cluster | grep rook-ceph-osd-0 | awk '!/rook-ceph-osd-0-debug/' | awk '{ print $1  }')
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-osd-0 pod to be ready" 10 \
+            "kubectl -n test-cluster get pod $osd | grep 'rook-ceph-osd-0.*Running'"
+
+          # for testing osd purge scale the osd deplyment
+          kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
+          # we need to sleep so the osd will be marked down before purging the osd
+          sleep 5
```
subhamkrai, line 77:
```suggestion
```

subhamkrai (changes requested):
In general, can we use `kubectl wait` commands if possible? It will make the script smaller and easier to read.  Sorry for late suggestion 😅

parth-gr:
> In general, can we use `kubectl wait` commands if possible? It will make the script smaller and easier to read. Sorry for late suggestion sweat_smile

@subhamkrai We have been using script that Blaine provided to us instead of wait or timeout cmd, which will be more robust and informatic, as decided here https://github.com/rook/kubectl-rook-ceph/pull/48#discussion_r960734689

.github/workflows/ci-for-default-ns.yaml
parth-gr: These timeout's are there for certain reasons,

Like Subham this one is for whenever we run the  ` kubectl rook_ceph debug start rook-ceph-mon-a`
we have to wait till the debug pod should triggered and in initialising phase as sometimes we just trigger the pod to create and if api response is not quick we won't get any pod in that namespace that is been starting

.github/workflows/ci-for-default-ns.yaml
parth-gr: Same as above

parth-gr:
> @parth-gr can you try restarting CI multiple times and see how quickly they or there are getting stuck? since I started CI it is trying to run the commands for 8 min.

TOtal it took around, 5- 6 mins,  4 mins for setup and 1.5mins for running

.github/workflows/ci-for-default-ns.yaml
travisn: So if there is no pod created yet, the script fails to wait for the condition? Could we change the script to handle the case where the pod doesn't exist?

subhamkrai (approved):

.github/workflows/ci-for-default-ns.yaml
parth-gr: Thanks, Addressed them :)

travisn: I was thinking about this a bit differently... As part of the loop inside `wait_for_condition`, it should just wait for the pod to exist before also checking that the pod is running. Then we don't need to have the sleep and check for a pod duplicated everywhere.

travisn (changes requested):

.github/workflows/ci-for-default-ns.yaml
parth-gr: Okay addressed it

tests/github-action-helper.sh
```diff
@@ -104,6 +104,28 @@ wait_for_operator_pod_to_be_ready_state_custom() {
 EOF
 }
 
+wait_for_condition() {
```
travisn, line 107: This function is always called twice, once for pending, then for running state. How about if inside this function it checked for pending then running automatically? It really will simplify all the places this function is called.
```suggestion
wait_for_running_pod() {
```

parth-gr: Updated it, But according to this above discussion it will again become a complicated one-liner https://github.com/rook/kubectl-rook-ceph/pull/48#discussion_r960734689

BlaineEXE: Do we need to check for pending at all, or can we just wait for running?

travisn (changes requested):

tests/github-action-helper.sh
parth-gr: It may not be required, removed it

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,18 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-mon-a-debug deployment pod to start" 10 \
```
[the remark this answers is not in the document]
parth-gr: Okay
Removed the starting wit time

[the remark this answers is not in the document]
travisn: No, but why do we need three lines for the caller? Lines 78-80 should **only** be one line:
```
tests/github-action-helper.sh wait_for_running_pod rook-ceph-mon-a-debug
```
 If the caller is a single line instead of three, it is much simpler. I don't mean to create a very long line out of three, I mean that the one line is very simple and everything else for the implementation can be in the `wait_for_running_pod` function.

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,18 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          debug_mon=$(kubectl get pods -nrook-ceph | grep rook-ceph-mon-a-debug | awk '{ print $1  }')
+          tests/github-action-helper.sh wait_for_condition "rook-ceph-mon-a-debug deployment pod to start" 10 "kubectl -n rook-ceph get pod $debug_mon | grep 'rook-ceph-mon-a-debug.*Running'"
```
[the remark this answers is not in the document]
parth-gr: This looks good to me calling a simpler one-liner,

But by this, we might be restricting the script implementation to only wait for a pod, which can do a lot of operation waiting in overall, Like waiting for `cephblockpool` etc, I know we don't have many complications in the CI know but in future, it might more help us.....

Just a add-on suggestion otherwise I can go with the above flow:)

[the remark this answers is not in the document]
travisn: Currently does the function need to wait for anything except a pod? In this PR that's the only place I see it used. So I'd suggest we optimize for this PR. If a future PR needs to wait for a different resource, a new function could be added to wait for that resource. It will keep the test script a lot cleaner, even if we end up with multiple helper functions for different types of resources.

.github/workflows/ci-for-default-ns.yaml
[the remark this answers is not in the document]
parth-gr: `osd=$(kubectl get pod -n test-cluster | grep rook-ceph-osd-0 | awk '!/rook-ceph-osd-0-debug/' | awk '{ print $1  }')` 

One command was this, so do not pick the debug pod that is stopping,

Should I use if else to manage this in the function?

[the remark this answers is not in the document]
travisn: Perhaps we need a second argument to the function such as this? Then the function could handle it with if/else
```
tests/github-action-helper.sh wait_for_pod rook-ceph-mon-a-debug [running | deleted]
```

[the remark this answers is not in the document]
parth-gr: The 3rd argument, 2nd I already used for namespace XD

.github/workflows/ci-for-default-ns.yaml
```diff
@@ -75,15 +75,15 @@ jobs:
...
           # osd
           kubectl rook_ceph debug start rook-ceph-osd-0
-          sleep 5
+          tests/github-action-helper.sh wait_for_pod_running rook-ceph-osd-0-debug rook-ceph running
           kubectl rook_ceph debug stop rook-ceph-osd-0
 
-          # sleep again so OSD can start
-          sleep 5
+          # check if osd 0 is started
+          tests/github-action-helper.sh wait_for_pod_running rook-ceph-osd-0 rook-ceph deleted
```
travisn, line 86: Shouldn't we wait for the osd pod to be running again? It's the debug pod that should be deleted.
```suggestion
          tests/github-action-helper.sh wait_for_pod_running rook-ceph-osd-0 rook-ceph running
          tests/github-action-helper.sh wait_for_pod_running rook-ceph-osd-0-debug rook-ceph deleted
```

parth-gr: So the flow is like this 
```
if [[ $STATUS == "deleted" ]]
  then
    debug_pod=$(kubectl get pods -n ${NAMESPACE} | grep ${DEPLOYMENT} | awk '!/rook-ceph-osd-0-debug/' | awk '{ print $1  }')
  else
    debug_pod=$(kubectl get pods -n ${NAMESPACE} | grep ${DEPLOYMENT} | awk '{ print $1  }')
```

If it is `deleted` it will skip picking `rook-ceph-osd-0-debug`

tests/github-action-helper.sh
```diff
@@ -104,6 +104,38 @@ wait_for_operator_pod_to_be_ready_state_custom() {
 EOF
 }
 
+wait_for_pod_running() {
```
travisn, line 107: Since we wait for either running or deleted, how about a more generic name again :) 
```suggestion
wait_for_pod() {
```

subhamkrai:
@parth-gr update on this?

parth-gr:
> @parth-gr update on this?

Will discuss today

.github/workflows/ci-for-default-ns.yaml
parth-gr: Actually, I had some changes which would avoid `running` and `deleting` paramter,

It will just be hardcoded for "rook-ceph-osd-0"
```
if [[ $DEPLOYMENT == "rook-ceph-osd-0" ]]
  then
    debug_pod=$(kubectl get pods -n ${NAMESPACE} | grep ${DEPLOYMENT} | awk '!/rook-ceph-osd-0-debug/' | awk '{ print $1  }')
```

tests/github-action-helper.sh
```diff
@@ -104,6 +104,37 @@ wait_for_operator_pod_to_be_ready_state_custom() {
...
+
+  BASH_CMD="${BASH_CMD:=bash}" # might be 'bash -x', for example
+
+  msg="Waiting ${TIMEOUT} seconds for pod from deployment ${DEPLOYMENT} to be running..."
+  echo "${msg} countdown: ${TIMEOUT}"
+  start_time=$SECONDS
+  debug_pod
+  if [[ $DEPLOYMENT == "rook-ceph-osd-0" ]]
+  then
+    debug_pod=$(kubectl get pods -n ${NAMESPACE} | grep ${DEPLOYMENT} | awk '!/rook-ceph-osd-0-debug/' | awk '{ print $1  }')
```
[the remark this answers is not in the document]
travisn: How can this be changed so the pod name isn't hard coded in the helper?

tests/github-action-helper.sh
```diff
@@ -104,6 +104,37 @@ wait_for_operator_pod_to_be_ready_state_custom() {
...
+  TIMEOUT="10"  # in seconds
+
+  BASH_CMD="${BASH_CMD:=bash}" # might be 'bash -x', for example
+
+  msg="Waiting ${TIMEOUT} seconds for pod from deployment ${DEPLOYMENT} to be running..."
+  echo "${msg} countdown: ${TIMEOUT}"
+  start_time=$SECONDS
+  debug_pod
+  if [[ $DEPLOYMENT == "rook-ceph-osd-0" ]]
+  then
```
BlaineEXE, line 119:
```suggestion
  if [[ $DEPLOYMENT == "rook-ceph-osd-0" ]]; then
```

tests/github-action-helper.sh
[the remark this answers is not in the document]
parth-gr: @travisn So a special case because,
The debug-osd-0 pod still runs in the cluster, and if we go and fetch the osd-0 pod from the deployment name,
It returns the:
debug-osd-0 pod and osd-0-pod name, 
So to distinguish between them we need to have this special condition.

Another thing we can do is to watch for debug-osd-0 pod to be deleted, But I am not sure if we can do it for the CI as we don't know the debug+pod name, and if go and fetch the name using the deployment name the above problem mentioned needs to handle.

There is one thing that @BlaineEXE mentioned we can do,
For checking that the debug pod is deleted successfully before scaling back the original deployment in the actual code,
BUt this would add an additional wait to the application which is not required.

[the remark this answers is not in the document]
travisn: Actually, I think we can replace this function with a `kubectl wait` condition. In #61 I have a helper to [wait_for_pod_of_deployment_to_be_running](https://github.com/rook/kubectl-rook-ceph/pull/61/files#diff-8545dbb7aa4b5f9f70a1fc5bfb63a6ea985db01f32b89ec751eda93824d03651R247-R251). I believe it's the same thing you're trying to accomplish with this wait. Can you try that instead?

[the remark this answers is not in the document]
parth-gr: I thought the `rook-ceph-osd` pod is even not required so removed it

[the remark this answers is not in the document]
travisn: Did you look if we could use the wait_for_pod_of_deployment_to_be_running in this case as well? This function seems like it has the same purpose.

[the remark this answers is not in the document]
parth-gr: Yaa we can use that function too for this case.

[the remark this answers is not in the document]
parth-gr: But this current function provide more details,
SHould we go ahead with that as it looks simpler in terms of code...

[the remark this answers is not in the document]
parth-gr: @travisn they end up doing the same task,

But your function looks more robust becasue it directly checks the deployment resource without fetching the pod name and checking,
So updated to it.

tests/github-action-helper.sh
travisn: Since #61 is merged now, how about updating to call `wait_for_deployment_to_be_running`? Then I think this PR will be ready for merge.

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -58,25 +58,27 @@ jobs:
...
           kubectl rook_ceph -o test-operator -n test-cluster debug stop rook-ceph-osd-0
 
+           # check if osd 0 is started
+          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-osd-0 test-cluster
+        
+          # for testing osd purge scale the osd deplyment
+          kubectl --namespace test-cluster scale deploy/rook-ceph-osd-0 --replicas=0
+          # we need to sleep so the osd will be marked down before purging the osd
+          sleep 5
+          kubectl-rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
```
travisn, line 78: typo?
```suggestion
          kubectl rook_ceph -o test-operator -n test-cluster rook purge-osd 0 --force
```

parth-gr: probably it would be fine,
as I discussed this in an issue https://github.com/rook/kubectl-rook-ceph/issues/41#issuecomment-1201185823

Still trying to look more deep into this call.

parth-gr:
@travisn I see the `kubectl-rook-ceph.sh` is not in the CI build,
So that means I need to have a separate function in the github-action-helper.sh file

Or otherwise, I need to include that file in the CI
```
runner@fv-az183-529:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ ls
LICENSE  README.md  cluster-test.yaml  common.yaml  docs  kubectl-rook-ceph.sh  media  operator.yaml  storageclass-test.yaml  tests  toolbox.yaml
runner@fv-az183-529:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ 

```

parth-gr:
> > @travisn I see the `kubectl-rook-ceph.sh` is not in the CI build, So that means I need to have a separate function in the github-action-helper.sh file
> > Or otherwise, I need to include that file in the CI
> > ```
> > runner@fv-az183-529:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ ls
> > LICENSE  README.md  cluster-test.yaml  common.yaml  docs  kubectl-rook-ceph.sh  media  operator.yaml  storageclass-test.yaml  tests  toolbox.yaml
> > runner@fv-az183-529:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ 
> > ```
> 
> it will be present under `tests` folder

I only see 
```
runner@fv-az183-529:~/work/kubectl-rook-ceph/kubectl-rook-ceph/tests$ ls
github-action-helper.sh
```

parth-gr:
I was not able to call the function inside the `kubectl-rook-ceph.sh` as because of it's structure,

So moved the function to the github-action-helper so it can be called from the CI and the file

kubectl-rook-ceph.sh
```diff
@@ -244,12 +244,6 @@ function fetch_mon_endpoints() {
   KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o json | jq --monochrome-output '.data.data' | tr -d '"' | tr -d '=' | sed 's/[A-Za-z]*//g'
 }
 
-wait_for_deployment_to_be_running() {
```
travisn, line 247: This function is needed by the restore-quorum command, so if we can't call this helper from the CI, it's a simple one to duplicate.

parth-gr: But we can call this function from a github helper file,
```tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-tools "$ROOK_CLUSTER_NAMESPACE"```

So with the krew plugin you mean to say it won't be possible to call this way and it will fail as it might not have this directory with it??

Comment (travisn) [reply kubectl-rook-ceph.sh:247]: Correct, the krew plugin only has this one .sh file installed, so the helper script is not available.

File: .github/workflows/ci-for-default-ns.yaml [hunk 24]
```diff
@@ -75,15 +75,15 @@ jobs:
           # for testing start-debug and stop-debug deployment
           # mons
           kubectl rook_ceph debug start rook-ceph-mon-a
-          sleep 5
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug rook-ceph
```
travisn, line 78: Did you try this? Seems like it should work to call a function in that plugin script.
```suggestion
          kubectl-rook-ceph.sh wait_for_deployment_to_be_running rook-ceph-mon-a-debug rook-ceph
```

parth-gr: Yaa, I tried using it,
But it won't work as in because of it's structure, As when we called ` kubectl-rook-ceph.sh ` it will look for the following toolbox cmds let say, but won't directly call any function, 
For ex: `kubectl-rook-ceph.sh debug start` and won't like directly calling a function ` kubectl-rook-ceph.sh wait_for_deployment_to_be_running `,
But the `tests/github-action-helper.sh` entry point has this check to call a certain function.

travisn: Ok thanks for confirming

travisn (approved):

Issue #44, opened 2022-08-10
Flag position should not matter


parth-gr:
What if users want to issue the command like `kubectl rook-ceph start-debug <deploy> --image=my-image`? Or what if they do `kubectl rook-ceph start-debug --image my-image <deploy>`? This will fail in both cases. We should use the `parse_flags()` function for all flag parsing.

The commented-out `run_status_command()` example should help.

_Originally posted by @BlaineEXE in https://github.com/rook/kubectl-rook-ceph/pull/36#discussion_r935005719_

BlaineEXE:
I think we have to be careful here. Flag positioning is really only a concern for commands that take an argument that is not a flag. At least for now, that case is limited to the last sub-command in a chain. I think it's sufficient to use `parse_flags()` twice to support this case without requiring further changes to `parse_flags()` which would complicate that function further.

parth-gr:
@BlaineEXE do we also need to consider this case `kubectl rook-ceph  --image=my-image start-debug <deploy>`

BlaineEXE:
I would say no. `--image` is a flag for `start-debug` and not a top-level flag. IIRC, kubectl has some "global" flags like `--namespace`, but many flags are only associated with a given subcommand and won't work before the subcommand.

parth-gr:
Okay thanks

PR #49, opened 2022-08-30
core: make alternate image flag unpositioned for debug mode

parth-gr:
Closes:  https://github.com/rook/kubectl-rook-ceph/issues/44

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -95,6 +95,7 @@ function flag_no_value() {
 # When a non-flag arg is reached, stop parsing and return the remaining args in REMAINING_ARGS.
 REMAINING_ARGS=()
 function parse_flags() {
+  REMAINING_ARGS=()
   local set_value_function="$1"
   shift # pop set_value_function arg from the arg list
   while (($#)); do
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -411,11 +411,18 @@ function verify_debug_deployment() {
   fi
 }
 
-function run_start_debug() {
-  [[ -z "${1:-""}" ]] && fail_error "Missing mon or osd deployment name"
-  deployment_name="$1"
-  alternate_image_flag="${2:-""}"
-  alternate_image="${3:-""}"
+function run_start_debug(){
+  # debug start can be used in different ways 
+  # 1) debug start deploymentName --alternate-image imageName 
+  # 2) debug start --alternate-image imageName deploymentName
+  # 3) debug start deploymentName
+  parse_flags parse_image_flag "$@" # parse flags before the deployment name
+  [[ -z "${REMAINING_ARGS[0]:-""}" ]] && fail_error "Missing mon or osd deployment name"
+  deployment_name="${REMAINING_ARGS[0]}" # get deployment name
+  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}") # remove deploy name from remaining args
+  parse_flags parse_image_flag "${REMAINING_ARGS[@]}" # parse flags after the deployment name
+  end_of_command_parsing "${REMAINING_ARGS[@]}"
+
   verify_debug_deployment "$deployment_name"
 
   # copy the deployment spec before scaling it down
@@ -427,17 +434,9 @@ function run_start_debug() {
   # remove probes from the deployment
   deployment_spec=$(remove_probe_from_deployment "$deployment_spec")
   # update the deployment_spec with new image if alternate-image is passed
-  if [ -n "$alternate_image_flag" ]; then
-    if [ "$alternate_image_flag" == "--alternate-image" ]; then
-      if [ "$#" -eq 3 ]; then
-        echo "setting debug image to $alternate_image"
-        deployment_spec=$(update_deployment_spec_image "$deployment_spec" "$alternate_image")
-      else
-        fail_error "if --alternate-image flag is passed then image argument is mandatory"
-      fi
-    else
-      fail_error "--alternate-image should be passed as flag"
-    fi
+  if [ -n "$ALTERNATE_IMAGE" ];then
+    echo "setting debug image to \"$ALTERNATE_IMAGE\""
+    deployment_spec=$(update_deployment_spec_image "$deployment_spec" "$ALTERNATE_IMAGE")
   fi
   # update the deployment_spec main container to be only a placeholder,
   echo "setting debug command to main container"
@@ -477,7 +476,21 @@ function run_stop_debug() {
   KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=1
 }
 
-function run_debug() {
+function parse_image_flag() {
+  local flag="$1"
+  local val="$2"
+  case "$flag" in
+  "--alternate-image")
+    val_exists "$val" || return 1 # val should exist
+    ALTERNATE_IMAGE="${val}"
+    ;;
+  *)
+    fail_error "Flag $flag is not supported"
+    ;;
+  esac
+}
+
+function run_debug(){
   [[ -z "${1:-""}" ]] && fail_error "Missing 'debug' subcommand"
   subcommand="$1"
   shift # remove the subcommand from the front of the arg list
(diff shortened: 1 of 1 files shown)


kubectl-rook-ceph.sh
```diff
@@ -568,6 +581,7 @@ function run_main_command() {
 
 # Default values
 : "${ROOK_CLUSTER_NAMESPACE:=rook-ceph}"
+: "${ALTERNATE_IMAGE=0}"
```
subhamkrai, line 585:
```suggestion
: "${ALTERNATE_IMAGE=}"
```

kubectl-rook-ceph.sh
```diff
@@ -427,17 +434,9 @@ function run_start_debug(){
...
-      if [ "$#" -eq 3 ]; then
-        echo "setting debug image to $alternate_image"
-        deployment_spec=$(update_deployment_spec_image "$deployment_spec" "$alternate_image")
-      else
-        fail_error "if --alternate-image flag is passed then image argument is mandatory"   
-      fi
-    else
-      fail_error "--alternate-image should be passed as flag"
-    fi
+  if [[ $ALTERNATE_IMAGE != 0 ]];then
```
subhamkrai, line 438:
```suggestion
  if [ -z $ALTERNATE_IMAGE ]; then
```

subhamkrai (changes requested):

subhamkrai (changes requested):
@parth-gr could you use vsCode extention like `shell-format` `shellcheck` for bash formatting? I see space and other lint.

subhamkrai:
Also, I'll let @BlaineEXE review the flags changes.

kubectl-rook-ceph.sh
```diff
@@ -411,11 +412,18 @@ function verify_debug_deployment() {
...
 }
 
-function run_start_debug() {
-  [[ -z "${1:-""}" ]] && fail_error "Missing mon or osd deployment name"
-  deployment_name="$1"
-  alternate_image_flag="${2:-""}"
-  alternate_image="${3:-""}"
+function run_start_debug(){
+  parse_flags parse_image_flag "$@"
+  if [ -n "$ALTERNATE_IMAGE" ];then
```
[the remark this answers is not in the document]
parth-gr: @BlaineEXE 

The current code works completely fine
```
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh debug start --alternate-image imagename deploymentName 
Deployment Name: deploymentName
Alternate Image Name:imagename
[rider@localhost kubectl-rook-ceph]$ 
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh debug start deploymentName --alternate-image imagename
Deployment Name: deploymentName
Alternate Image Name:imagename
[rider@localhost kubectl-rook-ceph]$ 
[rider@localhost kubectl-rook-ceph]$ 
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh debug start deploymentName
Deployment Name: deploymentName
Alternate Image Name:
[rider@localhost kubectl-rook-ceph]$ 
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh debug start --alternate-image imagename
Error:  Missing mon or osd deployment name
[rider@localhost kubectl-rook-ceph]$ ./kubectl-rook-ceph.sh debug start deploymentName --alternate-image
./kubectl-rook-ceph.sh: line 136: $1: unbound variable
```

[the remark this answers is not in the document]
parth-gr: The above code and `end_of_command_parsing` cannot be used, the command needs to be shifted when deployment name is at starting as the "${REMAINING_ARGS[@]}" will still contain the deployment name, so it need array shifting and all
And also for `end_of_command_parsing` we need shift if the deployment name is at the last as the "${REMAINING_ARGS[@]} will still contain the deployment name

 And also it needs another Debug_deploymernt global variable to make use

If you just want to shift this code into a function I can do it.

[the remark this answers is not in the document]
BlaineEXE: We should be using `end_of_command_parsing` to flag issues where users give too many commands.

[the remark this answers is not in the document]
BlaineEXE: Also, the script should never error because of an unbound variable.

Even if this code works, I am having trouble reading the code's intent, which usually means maintaining the code in the future will be more difficult. I think breaking it into a function and treating DEPLOYMENT_NAME as an argument (which it is) will make things more clear. And on that note as well, comments to explain what is happening will help too.

kubectl-rook-ceph.sh
```diff
@@ -411,11 +412,18 @@ function verify_debug_deployment() {
...
+    [[ -z "${REMAINING_ARGS:-""}" ]] && fail_error "Missing mon or osd deployment name"
```
BlaineEXE, line 418: I'm not sure if this is what you intend here, but `REMAINING_ARGS` is an array. `"${REMAINING_ARGS}"` will implicitly return only the first item of the array. Generally, implicit behavior makes code harder to read, and I suggest using `"${REMAINING_ARGS[0]:-""}"` to make it more explicit that the intent is to check only the first array item.

kubectl-rook-ceph.sh
```diff
@@ -95,6 +95,7 @@ function flag_no_value() {
 # When a non-flag arg is reached, stop parsing and return the remaining args in REMAINING_ARGS.
 REMAINING_ARGS=()
 function parse_flags() {
+  REMAINING_ARGS=()
```
BlaineEXE, line 98: This changes the functionality of `parse_flags()` which affects all other commands. This was intentionally left out of the function so that it doesn't destroy the value of `REMAINING_ARGS` until it successfully ends.

BlaineEXE (changes requested):

kubectl-rook-ceph.sh
parth-gr: Done

kubectl-rook-ceph.sh
parth-gr: Let's take a case in which we don't pass the deployment name by mistake, and if we don't refresh the REMAINING_ARGS

`./kubectl.sh  debug start --alternate-image imag1`

If we use REMAINING_ARGS outside parse_flags() ,

It will update   Deployment_Name to `debug ` as `REMAINING_ARGS` still contains `debug start --alternate-image 2` when it was called at the very starting.

So whenever there is no non flag argument it will not update it


Maybe I can discuss this with you in a call

BlaineEXE: Hmm... let me pull it and take a look. Maybe I have something mixed up.

kubectl-rook-ceph.sh
```diff
@@ -566,6 +580,16 @@ function run_main_command() {
   esac
 }
 
+# Default values
+: "${ROOK_CLUSTER_NAMESPACE:=rook-ceph}"
+: "${ALTERNATE_IMAGE=}"
```
BlaineEXE, line 585: This isn't a global value. IMO, this should be moved to the section where debug commands are parsed.

kubectl-rook-ceph.sh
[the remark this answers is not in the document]
BlaineEXE: I played with this a bit and found that using a function does seem confusing. I have a further suggestion to help keep the code readable. It doesn't rely on conditionals of side-effects, so it is more straightforward to understand, and the comments should help as well. Plus, it uses end_of_command_parsing to accurately determine if a user has supplied too many commands.

```sh
ALTERNATE_IMAGE=""

function run_start_debug(){
  parse_flags parse_image_flag "$@" # parse flags before the deployment name
  deployment_name="${REMAINING_ARGS[0]}" # get deployment name
  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}") # remove deploy name from remaining args
  parse_flags parse_image_flag "${REMAINING_ARGS[@]}" # parse flags after the deployment name
  end_of_command_parsing "${REMAINING_ARGS[@]}"

  verify_debug_deployment "$deployment_name"

  # copy the deployment spec before scaling it down
  deployment_spec=$(KUBECTL_NS_CLUSTER get deployments "$deployment_name" -o json | jq -r ".spec")
  # ... etc. ...
```

This will also make it easier to add new flags in the future. Using conditionals on the assumption that there is only one flag will break if we add another flag to make this tool more configurable in the future.

kubectl-rook-ceph.sh
```diff
@@ -477,7 +477,21 @@ function run_stop_debug() {
...
 }
 
-function run_debug() {
+function parse_image_flag() {
+  local flag="$1"
+  local val="$2"
+  case "$flag" in
+  "--alternate-image")
+    val_exists "$val" || return 1 # val should exist
+    ALTERNATE_IMAGE="${val}"
```
parth-gr, line 486: Actually `ALTERNATE_IMAGE` var needs to pass with this function `parse_image_flag` and this function is already a sub-function call from `  parse_flags parse_image_flag "$@"`  so I was not sure how we can also pass the alternate_image var, So that's why kept it Global

kubectl-rook-ceph.sh
subhamkrai: I agree `ALTERNATE_IMAGE` this var is not a global variable.

parth-gr: @subhamkrai can you look into this https://github.com/rook/kubectl-rook-ceph/pull/49#discussion_r966895863 as a reason to make it GLobal

subhamkrai (approved):
approving my changes has been addressed

kubectl-rook-ceph.sh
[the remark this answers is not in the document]
parth-gr: Sorry, I missed this comment, 
addressed it

BlaineEXE (approved):

PR #52, opened 2022-09-02
core: default values should be set after reading flag

subhamkrai:
core: default values should be set after reading flag

we should read flags before setting the default values
otherwise, if someone provides a ceph cluster namespace, the
operator namespace will still point to default rook-ceph

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -566,15 +566,6 @@ function run_main_command() {
   esac
 }
 
-# Default values
-: "${ROOK_CLUSTER_NAMESPACE:=rook-ceph}"
-: "${ROOK_OPERATOR_NAMESPACE:=$ROOK_CLUSTER_NAMESPACE}"
-: "${TOP_LEVEL_COMMAND:=kubectl}"
-: "${RESET:=\033[0m}"                           # For no color
-: "${ERROR_PREFIX:=\033[1;31mError:$RESET}"     # \033[1;31m for Red color
-: "${INFO_PREFIX:=\033[1;34mInfo:$RESET}"       # \033[1;34m for Blue color
-: "${WARNING_PREFIX:=\033[1;33mWarning:$RESET}" # \033[1;33m for Yellow color
-
 ####################################################################################################
 # MAIN: PARSE MAIN ARGS AND CALL MAIN COMMAND HANDLER
 ####################################################################################################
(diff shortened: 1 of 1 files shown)


kubectl-rook-ceph.sh
```diff
@@ -378,50 +378,50 @@ function check_mgr_pods_status_and_count() {
...
+  deployment_spec=$(jq 'del(.template.spec.containers[0].startupProbe)' <<<$deployment_spec)
+  echo "$deployment_spec" # for returning the updated deployment_spec
 }
 
-function update_deployment_spec_command(){
+function update_deployment_spec_command() {
   deployment_spec=$1
   main_command='["sleep","infinity"]'
-  deployment_spec=$(jq .template.spec.containers[0].command="$main_command" <<< $deployment_spec)
-  deployment_spec=$(jq .template.spec.containers[0].args=[] <<< $deployment_spec)
```
travisn, line 392: The editor automatically changed all of these? Seems like this space is good to keep, unless the editor really removes it.

kubectl-rook-ceph.sh
```diff
@@ -610,6 +601,15 @@ function parse_main_flag() {
 REMAINING_ARGS=()
 parse_flags 'parse_main_flag' "$@"
 
+# Default values
```
travisn, line 604: We have a CI test in a namespace other that rook-ceph, right? If so, I'm curious how we missed this.

kubectl-rook-ceph.sh
subhamkrai: yes, it was the editor

kubectl-rook-ceph.sh
subhamkrai: so in the CI or generally when I did testing, we are either not passing the namespace or passing both the namespace cluster and operator. That's why it is missed in CI.

The current behavior is
1. default is rook-ceph
2.  if only the cluster namespace passed then the operator will also pick the namespace as the cluster
3. we can pass both cluster and operator namespace

The only other case I can see that I don't think is important is when the user is only passing the operator namespace, in this case, cluster namespace will rook-ceph only which I think make sense if user don't pass namespace we stick with `rook-ceph`

travisn: Agreed on that behavior, thanks for clarifying.

travisn (approved):

PR #61, opened 2022-10-10
core: Add command to reset mon quorum

travisn:
When quorum is lost, restoring quorum to a single mon is currently a complex manual process. Now with this krew command the admin can with less risk reset the mon quorum and restore the cluster again in disaster scenarios.

There are still some TODOs marked in the script to harden it some more.


README.md
```diff
@@ -58,6 +58,7 @@ These are args currently supported:
 - `rbd <args>` : Call a 'rbd' CLI command with arbitrary args
 
 - `mons` : Print mon endpoints
+  - `reset-quorum <mon-name>` : Reset the mon quorum to a single mon since quorum was lost with the other mons
```
subhamkrai, line 61:
```suggestion
  - `restore-quorum <mon-name>` : Reset the mon quorum to a single mon since quorum was lost with the other mons
```
how about `restore` instead of `reset`

docs/mons.md
```diff
@@ -1,4 +1,6 @@
-# Mons
+# Mon Commands
```
subhamkrai, line 1: this change requires to update in the Readme doc where we are calling mon-endpoints

docs/mons.md
```diff
@@ -1,4 +1,6 @@
...
+
+## Print Mons
```
subhamkrai, line 3:
```suggestion
## Print Mons Endpoints
```
?

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,176 @@ function path_cm_rook_ceph_operator_config() {
...
 function fetch_mon_endpoints() {
   end_of_command_parsing "$@" # end of command tree
   KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o json | jq --monochrome-output '.data.data' | tr -d '"' | tr -d '=' | sed 's/[A-Za-z]*//g'
 }
 
+
+wait_for_pod_of_deployment_to_be_running() {
+  echo -e "\nTODO: waiting for the pod from deployment \"rook-ceph-mon-$good_mon-debug\" to be running"
+  sleep 5
+}
```
subhamkrai, line 251:
```suggestion
wait_for_pod_of_deployment_to_be_running() {
  echo -e "\nTODO: waiting for the pod from deployment \"rook-ceph-mon-$good_mon-debug\" to be running"
  sleep 5
}
```

instead, can we do something like 
```
kubectl wait deployment -n <ns> <deployment-name> --for condition=Available=True --timeout=90s
```

subhamkrai (changes requested):
added some initial reviews

README.md
travisn: I'm going back and forth between them. When I started I actually called it restore, but then reset seemed more accurate.

kubectl-rook-ceph.sh
travisn: Yeah, the sleep was just a hack :)

README.md
subhamkrai: hm, restore seems better to me but yeah let's have a small discussion and get  others feedback 😅

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+  shift # remove the healthy mon from the front of the arg list
+  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}")           # remove mon name from remaining args
+  end_of_command_parsing "$@" # end of command tree
+
+  # Parse the endpoints configmap for the mon endpoints
+  bad_mons=()
+  mon_endpoints=$(KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o jsonpath='{.data.data}')
+  for single_mon in ${mon_endpoints//,/ } ; do
+    mon_name=$(echo "${single_mon/=/ }" | awk '{print $1}')
+    mon_endpoint=$(echo "${single_mon/=/ }" | awk '{print $2}')
```
subhamkrai, line 266: few questions?
`  for single_mon in ${mon_endpoints//,/ } ; do` replace every `,` with nothing?

` mon_name=$(echo "${single_mon/=/ }" | awk '{print $1}')`  this replaces first `=` ?

    `mon_endpoint=$(echo "${single_mon/=/ }" | awk '{print $2}')` this replaces first `=` ?

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+}
+
 function fetch_mon_endpoints() {
   end_of_command_parsing "$@" # end of command tree
   KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o json | jq --monochrome-output '.data.data' | tr -d '"' | tr -d '=' | sed 's/[A-Za-z]*//g'
 }
 
+wait_for_pod_of_deployment_to_be_running() {
+  deployment=$1
+  echo -e "\nWaiting for the pod from deployment \"$deployment\" to be running"
```
subhamkrai, line 249:
```suggestion
  info_msg " Waiting for the pod from deployment \"$deployment\" to be running"
```

in other places, for logging, use this 
```
function info_msg() {
  echo -e "${INFO_PREFIX}" "$@"
}

function warn_msg() {
  echo -e "${WARNING_PREFIX}" "$@"
}

function error_msg() {
  echo -e "${ERROR_PREFIX}" "$@"
}
```

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+  parse_flags parse_image_flag "$@" # parse flags before the good mon name
+  [[ -z "${REMAINING_ARGS[0]:-""}" ]] && fail_error "Missing healthy mon name"
+  good_mon="${REMAINING_ARGS[0]}"              # get the good mon being used to reset quorum
+  shift # remove the healthy mon from the front of the arg list
+  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}")           # remove mon name from remaining args
+  end_of_command_parsing "$@" # end of command tree
+
+  # Parse the endpoints configmap for the mon endpoints
+  bad_mons=()
+  mon_endpoints=$(KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o jsonpath='{.data.data}')
```
subhamkrai, line 263: we already have a method to print mon endpoints, can we use that `fetch_mon_endpoints`

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+  echo "The mons to discard are: ${bad_mons[*]}"
+  echo "The cluster fsid is $ceph_fsid"
+  echo "If you want to continue resetting the mon quorum to $good_mon, type \"yes\""
+  read CONTINUE_SCRIPT
+  if [ "$CONTINUE_SCRIPT" != "yes" ]; then
+    exit 1
+  fi
+
+  # scale the operator deployment down
+  KUBECTL_NS_CLUSTER scale deployment rook-ceph-operator --replicas=0
```
subhamkrai, line 300:
```suggestion
  KUBECTL_NS_OPERATOR scale deployment rook-ceph-operator --replicas=0
```

for operator namespace use `KUBECTL_NS_OPERATOR`

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+
+  # scale down all the mon pods
+  KUBECTL_NS_CLUSTER scale deployment -l app=rook-ceph-mon --replicas=0
+
+  # wait for the operator and mons to all stop
+  echo -e "\nWaiting for operator and mon pods to stop"
+  KUBECTL_NS_CLUSTER wait pod -l app=rook-ceph-operator --for=delete --timeout=90s
```
subhamkrai, line 307:
```suggestion
  KUBECTL_NS_OPERATOR wait pod -l app=rook-ceph-operator --for=delete --timeout=90s
```

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+    --foreground \
+    --public-addr=$good_mon_public_ip \
+    --setuser-match-path=/var/lib/ceph/mon/ceph-$good_mon/store.db \
+    --public-bind-addr=$ROOK_POD_IP \
+    --inject-monmap=$monmap_path
+  echo -e "\nFinished updating the monmap!"
+  set -eu
+
+  echo -e "\nPrinting final monmap"
+  KUBECTL_NS_OPERATOR exec deploy/rook-ceph-mon-$good_mon-debug -c mon -- monmaptool --print $monmap_path
```
subhamkrai, line 370:
```suggestion
  KUBECTL_NS_CLUSTER exec deploy/rook-ceph-mon-$good_mon-debug -c mon -- monmaptool --print $monmap_path
```
`KUBECTL_NS_OPERATOR` is this right or it will be `KUBECTL_NS_CLUSTER`

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,202 @@ function path_cm_rook_ceph_operator_config() {
...
+  i=0
+  max_retries=20
+  sleep_time=5
+
+  exit_status=1
+  while [[ $exit_status != 0 ]]
+  do
+    # Don't fail the script if the ceph command fails
+    set +e
+    KUBECTL_NS_OPERATOR exec deploy/rook-ceph-tools -- ceph status --connect-timeout=3
```
subhamkrai, line 409:
```suggestion
    KUBECTL_NS_CLUSTER exec deploy/rook-ceph-tools -- ceph status --connect-timeout=3
```
also we have method to print ceph command can we use that?

subhamkrai (changes requested):

kubectl-rook-ceph.sh
```diff
@@ -423,15 +617,13 @@ function run_start_debug() {
   [[ -z "${REMAINING_ARGS[0]:-""}" ]] && fail_error "Missing mon or osd deployment name"
   deployment_name="${REMAINING_ARGS[0]}"              # get deployment name
   REMAINING_ARGS=("${REMAINING_ARGS[@]:1}")           # remove deploy name from remaining args
-  parse_flags parse_image_flag "${REMAINING_ARGS[@]}" # parse flags after the deployment name
-  end_of_command_parsing "${REMAINING_ARGS[@]}"
```
travisn, line 427: @parth-gr @subhamkrai  These two lines to `parse_flags` and `end_of_command_parsing` cause an error if run when the restore-quorum command is run. But if removed like this, they don't cause a problem for the debug start command or the restore-quorum command. Am I missing anything by removing these two lines?

subhamkrai: `parse_flags` is to read flag args and `end_of_command_parsing` is to check if no other args is passed.

@parth-gr can give more details here.

parth-gr: `end_of_command_parsing` can be removed as it just says there should not be any args after it, but which can be for this case.

parse_flags parse_image_flag "${REMAINING_ARGS[@]}" # parse flags after the deployment name
This would check if there is any alternate_image flag after deployment name,
What is the error which come if this line is kept?

kubectl-rook-ceph.sh
```diff
@@ -445,13 +651,20 @@ function run_start_debug() {
...
   deployment_pod=$(KUBECTL_NS_CLUSTER get pod | grep "$deployment_name" | awk '{ print $1  }')
-  # scale the deployment to 0
-  KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=0
-
-  # wait for the deployment pod to be deleted
-  echo "waiting for the deployment pod \"$deployment_pod\" to be deleted"
-  KUBECTL_NS_CLUSTER wait --for=delete pod/"$deployment_pod" --timeout=60s
+  set -e
+  if [ "$deployment_pod" != "" ]; then
+    # scale the deployment to 0
```
parth-gr, line 660: scaling deployment to 0 can be still be out of `if` statement

kubectl-rook-ceph.sh
```diff
@@ -465,6 +678,8 @@ function run_start_debug() {
     spec:
         $deployment_spec
 EOF
+    echo "ensure the debug deployment $deployment_name is scaled up"
+    KUBECTL_NS_CLUSTER scale deployments "$deployment_name-debug" --replicas=1
```
parth-gr, line 682: Is this needed for any case?

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,221 @@ function path_cm_rook_ceph_operator_config() {
...
+    exit 1
+  fi
+  if [ "$ceph_fsid" = "" ]; then
+    error_msg "ceph cluster fsid not found"
+    exit 1
+  fi
+
+  # Check for the existence of the toolbox
+  info_msg "Start the toolbox if it is not yet running"
+  wait_for_pod_of_deployment_to_be_running rook-ceph-tools
```
parth-gr, line 290: maybe if we just say
`wait_for_pod_to_run`

travisn: `rook-ceph-tools` is the name of a deployment, not a pod. How about `wait_for_deployment_to_run`?

kubectl-rook-ceph.sh
travisn: If there is no pod, is there any reason to still scale it to 0? Are you just saying it's a precaution to still scale it down?

kubectl-rook-ceph.sh
travisn: This is needed because if the original daemon was scaled down, the debug pod needs to be scaled up. In the case of the mon restore, I'm scaling down the mon deployments before starting the debug pod, so I had to scale it back up at least for this scenario. But in other scenarios, this would just be a no-op.

kubectl-rook-ceph.sh
travisn: If I keep that line to parse the flags:
```
parse_flags parse_image_flag "${REMAINING_ARGS[@]}"
```

I get the error:
```
./kubectl-rook-ceph.sh: line 634: REMAINING_ARGS[@]: unbound variable
```

I'll take a closer look to make sure the `alternate_image` is still supported.

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,220 @@ function path_cm_rook_ceph_operator_config() {
...
+  for bad_mon in "${bad_mons[@]}"
+  do
+    info_msg "purging old mon: $bad_mon"
+    KUBECTL_NS_CLUSTER delete deploy rook-ceph-mon-$bad_mon
+    KUBECTL_NS_CLUSTER delete svc rook-ceph-mon-$bad_mon
+  done
+
+  info_msg "Mon quorum was successfully restored to mon $good_mon"
+
+  prompt_to_continue_or_cancel "Start up the operator and expand to full mon quorum again?" "yes"
```
subhamkrai, line 391:
```suggestion
  prompt_to_continue_or_cancel "Start up the operator and expand to full mon quorum again?" "yes"
```
since Mon are successfully restored(as we are logging above) why we are asking for starting the operator up again, we should be doing every time once mons are up?

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,220 @@ function path_cm_rook_ceph_operator_config() {
...
+  proceed_answer=$2
+  info_msg "$proceed_question If so, enter: $proceed_answer"
+  set +u
+  if [[ "$ROOK_PLUGIN_SKIP_PROMPTS" != "true" ]]; then
+    read INPUT_VAR
+    if [ "$INPUT_VAR" = "$proceed_answer" ]; then
+      info_msg "proceeding"
+    else
+      warn_msg "cancelled"
+      exit 1
```
subhamkrai, line 408:
```suggestion
    if [ "$INPUT_VAR" = "$proceed_answer" ]; then
      info_msg "proceeding"
    else
      warn_msg "cancelled"
      exit 1
```

I have a question about this maybe I thinking to much, but since checking for strict `"$INPUT_VAR" = "$proceed_answer"` and if somehow user has a typo or just does enter( which somewhere means `default value` or n in y/n` case) then we are doing `exit 1`  which will exit the script in the middle of the process or some step. is this okay? maybe we are doing the steps in debug mode pod so we are good? 

Also, let's say the user didn't give any input in the prompt or wrong input that will lead to exit then should we delete the debug pod deployment in that case?

These are negative case questions and we can fix in follow-up too according to user feedback?

kubectl-rook-ceph.sh
```diff
@@ -226,14 +227,220 @@ function path_cm_rook_ceph_operator_config() {
...
+    fi
+  else
+    warn_msg "skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true"
+  fi
```
subhamkrai, line 412:
```suggestion
  else
    warn_msg "skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true"
  fi
```

is this set to handle for CI as I think it is good to have user input everytime?

subhamkrai (changes requested):

kubectl-rook-ceph.sh
parth-gr: Okay I see it is because the deployment spec which is copied, should have the replicas count zero

kubectl-rook-ceph.sh
parth-gr: Yes for a precaution,
A pod can be not present for certain interval because of some restarting, etc
But the deployment may exist

kubectl-rook-ceph.sh
parth-gr: Looks good

kubectl-rook-ceph.sh
```diff
@@ -423,15 +630,16 @@ function run_start_debug() {
   [[ -z "${REMAINING_ARGS[0]:-""}" ]] && fail_error "Missing mon or osd deployment name"
   deployment_name="${REMAINING_ARGS[0]}"              # get deployment name
   REMAINING_ARGS=("${REMAINING_ARGS[@]:1}")           # remove deploy name from remaining args
+  set +u
```
parth-gr, line 633: Does plus +u is set as it turns off the `unset variables as an error` 
SO to avoid failing

docs/mons.md
```diff
@@ -7,3 +9,27 @@ kubectl rook-ceph mons
...
+the cluster will be down. If the majority of mons are also lost permanently,
+the quorum will need to be restore to a remaining good mon in order to bring
+the Ceph cluster up again.
+
+To restore the quorum in this disaster scenario:
+1. Identify that mon quorum is lost. Some indications include:
+   - The Rook operator log shows timeout errors and continuously fails to reconcile
+   - All commands in the toolbox are unresponsive
+   - Multiple mon pods are likely down
+2. Identify which mon has good state (TBD)
```
gauravsitlani, line 25: To identify which mon has good state, a user can exec into the mon pod and check its status using the admin socket. For example:
```
1. Locate admin socket file: # ls /var/run/ceph/
2. Check `mon_status`  using : # ceph daemon /var/run/ceph/<mon_admin_socket_file> mon_status
```

travisn: Good point, thanks

kubectl-rook-ceph.sh
travisn: It's a good question, it just felt like a good thing to pause and say "I did my job to restore quorum to a single mon, are you sure you're ready to start the operator up again?"

gauravsitlani: if i understand correctly at this point we have a single monitor running right ? If the user cancels this prompt, they'll know that they're taking a risk to run with a single mon, maybe we should add some warning here if at a point they have a cluster running with a single monitor. 
And they would need to start and scale up the operator manually right in such situation.

kubectl-rook-ceph.sh
travisn: There are currently prompts in two places and I think it's fine if they exit the script:
1. All the info is gathered, and we prompt if the user wants to proceed. No debug pod has been started yet.
2. The mon restore is completed and we just have to scale up the operator. They could scale the operator manually if they exit. 
The second one could be annoying that you have to scale up the operator. Perhaps a better approach for the 2nd one is to "press any key to continue when you're ready to scale up the operator".

kubectl-rook-ceph.sh
travisn: Yes we need it for CI, and could be useful if someone automates it

kubectl-rook-ceph.sh
travisn: Correct. I wonder if this was only necessary on mac. I was hitting the same error for the main `debug start` command even though it was working in the CI.

kubectl-rook-ceph.sh
travisn: Yeah we don't want them to risk staying with a single mon. I think the "press any key to continue" will solve this.

travisn:
Here is the output of a test run in minikube, with verbose rocksdb output removed. See attached for the full output.
[mon-restore-full-output.txt](https://github.com/rook/kubectl-rook-ceph/files/9814314/mon-restore-full-output.txt)

```
~/src/go/src/github.com/rook/kubectl-rook-ceph$ ./kubectl-rook-ceph.sh mons restore-quorum ab
mon=ab, endpoint=10.103.13.182:6789
mon=ae, endpoint=10.102.163.160:6789
mon=af, endpoint=10.101.243.83:6789
Info: Start the toolbox if it is not yet running
Info: Waiting for the pod from deployment "rook-ceph-tools" to be running
deployment.apps/rook-ceph-tools condition met

Warning: Restoring mon quorum to mon ab (10.103.13.182)
Info: The mons to discard are: ae af
Info: The cluster fsid is 7a2d8bf6-670e-41a8-9ae2-8a61c6732241
Info: Are you sure you want to restore the quorum to mon "ab"? If so, enter: yes-really-restore
yes-really-restore
Info: proceeding
deployment.apps/rook-ceph-operator scaled
deployment.apps/rook-ceph-mon-ab scaled
deployment.apps/rook-ceph-mon-ae scaled
deployment.apps/rook-ceph-mon-af scaled
Info: Waiting for operator and mon pods to stop
pod/rook-ceph-operator-b5c96c99b-hvvlk condition met
pod/rook-ceph-mon-ae-58c8b486d6-4gcr5 condition met
pod/rook-ceph-mon-af-7c99ff79c8-6pj87 condition met
setting debug mode for "rook-ceph-mon-ab"
setting debug command to main container
get pod for deployment rook-ceph-mon-ab
deployment.apps/rook-ceph-mon-ab-debug created
ensure the debug deployment rook-ceph-mon-ab is scaled up
deployment.apps/rook-ceph-mon-ab-debug scaled
Info: Waiting for the pod from deployment "rook-ceph-mon-ab-debug" to be running
deployment.apps/rook-ceph-mon-ab-debug condition met
Info: Started debug pod, restoring the mon quorum in the debug pod
Info: Extracting the monmap
parse error setting 'public_bind_addr' to ''

REMOVED: Verbose rocksdb output

debug2022-10-18T19:03:54.603+0000 7f9d3be0e880 -1 wrote monmap to /tmp/monmap
Info: Printing monmap
monmaptool: monmap file /tmp/monmap
epoch 48
fsid 7a2d8bf6-670e-41a8-9ae2-8a61c6732241
last_changed 2022-10-18T19:03:09.461446+0000
created 2022-10-14T15:12:40.925913+0000
min_mon_release 17 (quincy)
election_strategy: 1
0: [v2:10.103.13.182:3300/0,v1:10.103.13.182:6789/0] mon.ab
1: [v2:10.102.163.160:3300/0,v1:10.102.163.160:6789/0] mon.ae
2: [v2:10.101.243.83:3300/0,v1:10.101.243.83:6789/0] mon.af
Info: Removing mon ae
monmaptool: monmap file /tmp/monmap
monmaptool: removing ae
monmaptool: writing epoch 48 to /tmp/monmap (2 monitors)
Info: Removing mon af
monmaptool: monmap file /tmp/monmap
monmaptool: removing af
monmaptool: writing epoch 48 to /tmp/monmap (1 monitors)
Info: Injecting the monmap
parse error setting 'public_bind_addr' to ''

REMOVED: Verbose rocksdb output

Info: Finished updating the monmap!
Info: Printing final monmap
monmaptool: monmap file /tmp/monmap
epoch 48
fsid 7a2d8bf6-670e-41a8-9ae2-8a61c6732241
last_changed 2022-10-18T19:03:09.461446+0000
created 2022-10-14T15:12:40.925913+0000
min_mon_release 17 (quincy)
election_strategy: 1
0: [v2:10.103.13.182:3300/0,v1:10.103.13.182:6789/0] mon.ab
Info: Restoring the mons in the rook-ceph-mon-endpoints configmap to the good mon
configmap/rook-ceph-mon-endpoints patched
Info: Stopping the debug pod for mon ab
setting debug mode for "rook-ceph-mon-ab-debug"
removing debug mode from "rook-ceph-mon-ab-debug"
deployment.apps "rook-ceph-mon-ab-debug" deleted
deployment.apps/rook-ceph-mon-ab scaled
Info: Check that the restored mon is responding
timed out
command terminated with exit code 1
Info: 0: waiting for ceph status to confirm single mon quorum
Info: sleeping 5
timed out
command terminated with exit code 1
Info: 1: waiting for ceph status to confirm single mon quorum
Info: sleeping 5
timed out
command terminated with exit code 1
Info: 2: waiting for ceph status to confirm single mon quorum
Info: sleeping 5
  cluster:
    id:     7a2d8bf6-670e-41a8-9ae2-8a61c6732241
    health: HEALTH_WARN
            OSD count 0 < osd_pool_default_size 1

  services:
    mon: 1 daemons, quorum ab (age 22s)
    mgr: a(active, since 40m)
    osd: 0 osds: 0 up, 0 in

  data:
    pools:   0 pools, 0 pgs
    objects: 0 objects, 0 B
    usage:   0 B used, 0 B / 0 B avail
    pgs:

Info: finished waiting for ceph status
Info: Purging the bad mons: ae af
Info: purging old mon: ae
deployment.apps "rook-ceph-mon-ae" deleted
service "rook-ceph-mon-ae" deleted
Info: purging old mon: af
deployment.apps "rook-ceph-mon-af" deleted
service "rook-ceph-mon-af" deleted
Info: Mon quorum was successfully restored to mon ab
Info: Only a single mon is currently running
Info: Press Enter to start the operator and expand to full mon quorum again

Info: continuing
deployment.apps/rook-ceph-operator scaled
```

README.md
```diff
@@ -58,6 +58,7 @@ These are args currently supported:
 - `rbd <args>` : Call a 'rbd' CLI command with arbitrary args
 
 - `mons` : Print mon endpoints
+  - `restore-quorum <mon-name>` : Restore the mon quorum to a single mon since quorum was lost with the other mons
```
subhamkrai, line 61:
```suggestion
  - `restore-quorum <mon-name>` : Restore the mon quorum to a single mon since quorum was lost with the other mons
```
is this right? `to a single mon`

kubectl-rook-ceph.sh
```diff
@@ -445,13 +661,20 @@ function run_start_debug() {
   echo "setting debug command to main container"
   deployment_spec=$(update_deployment_spec_command "$deployment_spec")
 
+  # scale down the daemon pod if it's running
+  set +e
+  echo "get pod for deployment $deployment_name"
```
subhamkrai, line 666:
```suggestion
  info_msg "get pod for deployment $deployment_name"
```
?

kubectl-rook-ceph.sh
```diff
@@ -445,13 +661,20 @@ function run_start_debug() {
...
-  # scale the deployment to 0
-  KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=0
-
-  # wait for the deployment pod to be deleted
-  echo "waiting for the deployment pod \"$deployment_pod\" to be deleted"
-  KUBECTL_NS_CLUSTER wait --for=delete pod/"$deployment_pod" --timeout=60s
+  set -e
+  if [ "$deployment_pod" != "" ]; then
+    # scale the deployment to 0
+    echo "scale down the deployment $deployment_name"
```
subhamkrai, line 671:
```suggestion
    info_msg "scale down the deployment $deployment_name"
```

kubectl-rook-ceph.sh
```diff
@@ -445,13 +661,20 @@ function run_start_debug() {
...
+    KUBECTL_NS_CLUSTER scale deployments "$deployment_name" --replicas=0
+
+    # wait for the deployment pod to be deleted
+    echo "waiting for the deployment pod \"$deployment_pod\" to be deleted"
```
subhamkrai, line 675:
```suggestion
    info_msg "waiting for the deployment pod \"$deployment_pod\" to be deleted"
```

subhamkrai (approved):
LGTM, just a few small nits

kubectl-rook-ceph.sh
parth-gr: Ohkay got it

.github/workflows/ci-for-diff-ns.yaml
```diff
@@ -49,6 +49,15 @@ jobs:
           sleep 5
           kubectl rook_ceph -o test-operator -n test-cluster rbd ls replicapool
 
+          # test the mon restore to restore to mon a, delete mons b and c, then add d and e
+          export ROOK_PLUGIN_SKIP_PROMPTS=true
```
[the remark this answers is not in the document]
travisn: Correct it's for the CI. Or if someone else wanted to avoid the prompts they could set it, though I wouldn't expect it for normal use.

README.md
travisn: I'll rephrase it for clarity

gauravsitlani (approved):
LGTM

travisn:
Also tested on openshift with three nodes and PVC and non-PVC

PR #65, opened 2022-10-20
docs: Add example output for restoring quorum

travisn:
Restoring the mon quorum can be a risky operation and the admin may want to see example output to give more confidence that they will be prompted to continue after they are sure they want to restore the quorum.

### DIFF
--- a/docs/mons.md
+++ b/docs/mons.md
@@ -38,3 +38,105 @@ In this example, quorum is restore to mon **a**.
 ```bash
 kubectl rook-ceph mons restore-quorum a
 ```
+
+Before the restore proceeds, you will be prompted if you want to continue.
+For example, here is the output in a test cluster up to the point of starting the restore.
+
+```console
+$ kubectl rook-ceph mons restore-quorum c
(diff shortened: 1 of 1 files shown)


docs/mons.md
```diff
@@ -38,3 +38,20 @@ In this example, quorum is restore to mon **a**.
...
+mon=b, endpoint=192.168.64.168:6789
+mon=c, endpoint=192.168.64.167:6789
+mon=a, endpoint=192.168.64.169:6789
+Info: Check for the running toolbox
+Info: Waiting for the pod from deployment "rook-ceph-tools" to be running
+deployment.apps/rook-ceph-tools condition met
+
+Warning: Restoring mon quorum to mon c (192.168.64.167)
+Info: The mons to discard are: b a
+Info: Are you sure you want to restore the quorum to mon "c"? If so, enter: yes-really-restore
```
subhamkrai, line 56: can you also add last few line of the output that shows successful message?

subhamkrai (approved):

PR #66, opened 2022-10-20
docs: Fix rendering of restore mons doc

travisn:
The restore quorum doc had some extra output and also need consistent formatting between the blocks of
output.

My bad for not viewing the rendered output before merging #65.


subhamkrai (approved):
I didn't look at the output, I should have checked 😄

subhamkrai (commented):
would like to add a line here too, to press enter to continue ?
https://github.com/rook/kubectl-rook-ceph/pull/66/files#diff-435fa82258cbc13d1341a8b3083eb5147d9be44afa7d12cb6a733352c11a38b2R133

I see we have added earlier

subhamkrai (approved):

Issue #69, opened 2022-10-26
Add krew command to debugging csi


Madhu-1:
- [ ]  Verify network connectivity from csi pods and ceph cluster

This can be done by doing curl/ping or executing ceph commands from from cephfs/rbd-polugin container of provisioner and daemonset pods.

- [ ]  Command to get dmesg logs from the node based on the pod name or pvc name

If the PVC is not attaching to a give pod, we can identify to which node the pod is scheduled and on that node/rbdplugin pod we can run dmesg and print logs which helps for debuggin

- [ ] Command to check any stale map or mount commands in csi plugin pods

More details about what command to run and where to run is documented [here](https://rook.io/docs/rook/latest/Troubleshooting/ceph-csi-common-issues/#rbd-stale-operations) and [here](https://rook.io/docs/rook/latest/Troubleshooting/ceph-csi-common-issues/#cephfs-stale-operations)

- [ ]  Command to pull required logs from the leader provisioner and sidecar container based on the pod name or the pvc name

We might have two provisioner pods running, sometime for new comers its hard to find leader pod and which container to pull the logs from , we can make simple helper command which helps in pulling the required logs

* For pvc create/delete issue pull logs from csi-provisioner and csi-rbdplugin/csi-cephfsplugin container
* For  snapshot create/delete issue pull logs from csi-snapshotter and csi-rbdplugin/csi-cephfsplugin container
* For  resize issue pull logs from csi-resizer and csi-rbdplugin/csi-cephfsplugin container
etc

- [ ] Command to run ceph commands from provisioner or nodeplugin containers

This helps sometime where csi is not getting expected results but still admin wants to manually mount the rbd image or umount the image or run rados commands or ceph fs commands

- [ ] Command to identify and cleanup stale resources (rbd images/subvolumes/omap)

This is a big topic and need lot of automation but this is really helpful and also at the same time its a dangerous command. i will provide more details when we start working on this one.

- [ ] Print kernel version of all the plugin pod nodes

Printing the kernel version from the nodes where cephfs/rbd plugin pod runs, this helps in some debug cases

- [ ] Recover from Node lost cases

some details [here](https://rook.io/docs/rook/latest/Troubleshooting/ceph-csi-common-issues/#node-loss) we might also need some command to remove watcher also.

travisn:
Before we implement these commands, we need a design about what will be most helpful for troubleshooting csi issues. For example, if the tool could help with questions such as:
- Why is my PVC unbound?
[...]

PR #72, opened 2022-11-10
csi: adding new command csi-debug

subhamkrai:
csi: adding new command csi-debug

command `csi-debug dmesg <node_name>` will help printing
the dmesg logs where pvc mounting is failing or the
csi-rbdplugin container of the csi-rbdplugin-xxxx pod on that node.

ci/docs: adds docs and ci test for csi-debug dmesg


subhamkrai:
part of #69

kubectl-rook-ceph.sh
```diff
@@ -795,6 +794,50 @@ function run_dr_health() {
...
+
+  case $1 in
+  rbd) ;;
+  cephfs) ;;
+  *)
+    fail_error "$1 not a valid csi driver name, valid names rbd/cephfs"
+    ;;
+  esac
+
+  csi_rbdplugin_pod=$(KUBECTL_NS_OPERATOR get pods --no-headers -o custom-columns=":metadata.name" --field-selector spec.nodeName=minikube | grep "$csi_driver" | grep -v "provisioner")
```
Madhu-1, line 835: hardcoded minikube value.

subhamkrai: changed 😅

Madhu-1:
Looks like i forgot to submit my last comment :dagger: 

My idea was to keep it simple for this user as below. Let me know what do you think about it.

```bash
kubectl rook_ceph csi-debug dmesg <pod-name> <pod-namespace>
```

* Get the pod details and extract Node where it is scheduled
* Get the PVC details from the PVC Name in the pod spec and check its rbd or cephfs
* Now from the NodeID in the pod spec get to the exact csi driver nodeplugin
* Exec into the nodeplugin pod and run required command , In this case its `dmesg`

The above three steps will remain the same for pod debugging, and they can be reused. Good to have it as a helper function and only the last command will change most of the cases.

subhamkrai:
> Looks like i forgot to submit my last comment dagger
> 
> My idea was to keep it simple for this user as below. Let me know what do you think about it.
> 
> ```shell
> kubectl rook_ceph csi-debug dmesg <pod-name> <pod-namespace>
> ```
> 
> * Get the pod details and extract Node where it is scheduled
> * Get the PVC details from the PVC Name in the pod spec and check its rbd or cephfs
> * Now from the NodeID in the pod spec get to the exact csi driver nodeplugin
> * Exec into the nodeplugin pod and run required command , In this case its `dmesg`
> 
> The above three steps will remain the same for pod debugging, and they can be reused. Good to have it as a helper function and only the last command will change most of the cases.

make sense, if we can keep the 1st and 2nd as helper method for most of the csi debug command and last one changes, will do the changes

subhamkrai:
will add ci and doc for this command

travisn:
The user may not even know about csi. What about a more generic command such as getting the volume health, which will go look for various issues that could contribute to the volume not being provisioned or mounted? 
```
kubectl rook-ceph volume health <pod-name> <pod-namespace>
```

Running dmesg seems like a very specific command and could have a lot of output. What if instead we print some health info about the cluster, the csi pods, and then print the name of the pod where we suggest they run dmesg? Or maybe it is useful to run dmesg for them, I'm just looking for how we can make this as useful as possible for users. Some troubleshooting commands will need to be more advannced, and some should be simpler as well. Maybe dmseg just seems like the complex place to start.

Madhu-1:
> The user may not even know about csi. What about a more generic command such as getting the volume health, which will go look for various issues that could contribute to the volume not being provisioned or mounted?
> 
> ```
> kubectl rook-ceph volume health <pod-name> <pod-namespace>
> ```
> 
> Running dmesg seems like a very specific command and could have a lot of output. What if instead we print some health info about the cluster, the csi pods, and then print the name of the pod where we suggest they run dmesg? Or maybe it is useful to run dmesg for them, I'm just looking for how we can make this as useful as possible for users. Some troubleshooting commands will need to be more advannced, and some should be simpler as well. Maybe dmseg just seems like the complex place to start.

sounds good. we can add this command looks much  simpler but we need to change the argument from pod name to PVC name as the PVC is created by not attached to a pod.

subhamkrai:

PR #73, opened 2022-11-14
csi-debug: command to check network conn b/w csi

subhamkrai:
with new command `csi-debug test-connection` we can verify the connection between csi-cephfs/rbd and ceph cluster.

### DIFF
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -564,6 +564,42 @@ function run_dr_health() {
   set -e
 }
 
+function run_csi_command() {
+  [[ -z "${1:-""}" ]] && fail_error "Missing csi-debug subcommand"
+  subcommand="$1"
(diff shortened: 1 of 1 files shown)


subhamkrai:
part of #69

subhamkrai:
sample output

```console
csi-debug test-connection 
Info: checking 'csi-cephfsplugin-provisioner' network connection
Info: Network connection between cluster and csi-cephfs is: Connected

Info: checking 'csi-rbdplugin-provisioner' network connection
Info: Network connection between cluster and csi-rbd is: Connected
```

travisn (changes requested):
What about making this part of a "csi health" command?
```
kubectl rook-ceph csi health
```

Then we can add more checks over time to this same command that show different parts of the health of the csi driver.

subhamkrai:

PR #79, opened 2023-01-04
core: move project to golang

subhamkrai:
starting migration from bash to go, starting with
basic ceph commands and getting mons endpoint


pkg/mons/kubectl.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"os/exec"
+
+	"github.com/spf13/cobra"
+)
+
+var script = `
+kubectl -n %s get cm rook-ceph-mon-endpoints -o json | jq --monochrome-output '.data.data' | tr -d '"' | tr -d '=' | sed 's/[A-Za-z]*//g'
```
travisn, line 26: I wonder if we could covert commands like this to use apis instead of kubectl, instead of delaying the conversion to later. It should be pretty simple in go to read the configmap then parse the data.

pkg/kube-client.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	log "github.com/sirupsen/logrus"
+
+	k8s "k8s.io/client-go/kubernetes"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func GetKubeClinet() (*k8s.Clientset, *corev1client.CoreV1Client) {
```
travisn, line 26:
```suggestion
func GetKubeClient() (*k8s.Clientset, *corev1client.CoreV1Client) {
```

pkg/ceph/ceph-commands.go
```diff
@@ -0,0 +1,147 @@
...
+	fmt.Printf(msg)
+	os.Exit(1)
+}
+
+var (
+	OperatorNamespace    string // operator namespae
+	CephClusterNamespace string // Cephcluster namespace
+)
+
+func CephCommands(cmd *cobra.Command, args []string, operatorNamespace, clusterNamespace string) {
```
travisn, line 54: How about if the method name indicates where to run
```suggestion
func RunCephCommandInOperatorPod(cmd *cobra.Command, args []string, operatorNamespace, clusterNamespace string) {
```

kubectl-rook-ceph.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import "github.com/kubectl-rook-ceph/cmd"
+
+func main() {
```
travisn, line 19: Do we need this in the root dir, or can we follow the pattern of the rook repo, where there are no go files in the root? The main go file could be in the cmd dir.

travisn (commented):
A few questions to get started

pkg/mons/kubectl.go
subhamkrai: yeah, we can do, I'll make the changes

kubectl-rook-ceph.go
subhamkrai: I'll check on this how we can do that.

pkg/ceph/ceph-commands.go
subhamkrai: done

pkg/kube-client.go
subhamkrai: done

kubectl-rook-ceph.go
subhamkrai: done

pkg/mons/kubectl.go
subhamkrai: still need to do this change

subhamkrai: done

subhamkrai:
btw this how ci is executing commands,  not sure why

```
kubectl rook-ceph ceph status
ceph status --connect-timeout=10 --conf=/var/lib/rook/rook-ceph/rook-ceph.config

+ kubectl rook-ceph mons
a=10.99.115.54:67[8](https://github.com/rook/kubectl-rook-ceph/actions/runs/3870952267/jobs/6598314558#step:5:9)[9](https://github.com/rook/kubectl-rook-ceph/actions/runs/3870952267/jobs/6598314558#step:5:10),b=[10](https://github.com/rook/kubectl-rook-ceph/actions/runs/3870952267/jobs/6598314558#step:5:11).107.55.17:6789,c=10.96.247.121:6789
```

pkg/cephrbd/ceph_rbd_commands.go
```diff
@@ -0,0 +1,151 @@
...
+		log.Fatal(err)
+	}
+
+	// Get Kubernetes Client
+	_, client := pkg.GetKubeClient()
+
+	opts := metav1.ListOptions{LabelSelector: fmt.Sprintf("app=%s", "rook-ceph-operator")}
+	list, err := client.Pods(operatorNamespace).List(context.TODO(), opts)
+	if err != nil || len(list.Items) == 0 {
+		log.Error("failed to get pod to exec into container")
```
travisn, line 80:
```suggestion
		log.Error("failed to get rook operator pod where the command could be executed")
```

pkg/cephrbd/ceph_rbd_commands.go
```diff
@@ -0,0 +1,151 @@
...
+		log.Fatal(err)
+	}
+
+	get, err := client.Pods(operatorNamespace).Get(context.TODO(), list.Items[0].Name, metav1.GetOptions{})
```
travisn, line 84: The Pods.List() on line 78 would already retrieve the pod. Why do we need the Pod.Get()?

pkg/cephrbd/ceph_rbd_commands.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	cmd = append(cmd, args...)
+
+	if len(flagName) != 0 {
+		for i := range flagName {
+			cmd = append(cmd, fmt.Sprint(flagName[i]), flagValue[i])
+		}
+	}
+	if cmd[0] == "ceph" {
+		cmd = append(cmd, "--connect-timeout=10", "--conf=/var/lib/rook/rook-ceph/rook-ceph.config")
```
travisn, line 111: The config path will be different if in another namespace

pkg/cephrbd/ceph_rbd_commands.go
```diff
@@ -0,0 +1,151 @@
...
+	} else {
+		cmd = append(cmd, "--conf=/var/lib/rook/rook-ceph/rook-ceph.config")
```
travisn, line 113: Instead of duplicating this `--conf` option, can we factor out the `--connect-timeout` to a different condition?

pkg/cephrbd/ceph_rbd_commands.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil || len(get.Status.ContainerStatuses) == 0 {
+		log.Error("failed to get container")
+		log.Fatal(err)
+	}
+
+	ExecCmd(cmd, rest, get.Name, get.Namespace, get.Status.ContainerStatuses[0].Name, args, flagName, flagValue)
+}
+
+// ExecCmd exec command on specific pod and wait the command's output.
+func ExecCmd(command *cobra.Command, restconfig *restclient.Config, podName, podNameSpace, containerName string, args, flagName, flagValue []string) {
```
travisn, line 94:
```suggestion
func ExecCmd(command *cobra.Command, restconfig *restclient.Config, podName, podNamespace, containerName string, args, flagName, flagValue []string) {
```

pkg/mons/mon-endpoints.go
```diff
@@ -0,0 +1,40 @@
...
+	// Get Kubernetes Client
+	_, client := pkg.GetKubeClient()
+
+	monCm, err := client.ConfigMaps(clusterNamespace).Get(context.TODO(), monConfigMap, v1.GetOptions{})
+	if err != nil {
+		log.Errorf("failed to get mon configmap %s %v", monConfigMap, err)
+	}
+
+	monData := monCm.Data["data"]
+	fmt.Println(monData)
```
travisn, line 39: To match the current output of the mons command, do we need to do any parsing of the endpoints?

travisn (changes requested):

pkg/mons/mon-endpoints.go
subhamkrai: no we are getting the output as expected `a=10.96.52.53:6789`

pkg/cephrbd/ceph_rbd_commands.go
subhamkrai: done

pkg/cephrbd/ceph_rbd_commands.go
subhamkrai: done

pkg/cephrbd/ceph_rbd_commands.go
subhamkrai: done

pkg/cephrbd/ceph_rbd_commands.go
subhamkrai: done

pkg/cephrbd/ceph_rbd_commands.go
subhamkrai: done

cmd/root/root.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	cobra.CheckErr(RootCmd.Execute())
+}
+
+func init() {
+
+	//rootCmd.PersistentFlags().StringVar(cephCmd, "", "", "")
```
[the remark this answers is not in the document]
subhamkrai: removed

cmd/mon/mons.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright © 2023 kubectl-rook-ceph
```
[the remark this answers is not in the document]
subhamkrai: done

cmd/ceph/ceph.go
```diff
@@ -0,0 +1,35 @@
...
+	ceph "github.com/kubectl-rook-ceph/pkg/cephrbd"
+
+	"github.com/spf13/cobra"
+)
+
+// CephCmd represents the ceph command
+var CephCmd = &cobra.Command{
+	Use:                "ceph",
+	Short:              "call a 'ceph' CLI command with arbitrary args",
+	Long:               `call a 'ceph' CLI command with arbitrary args`,
```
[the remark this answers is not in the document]
subhamkrai: removed the Long

cmd/ceph/ceph.go
```diff
@@ -0,0 +1,35 @@
+/*
```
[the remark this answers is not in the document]
subhamkrai: I kept main.go under cmd folder and all other commands inside `commands` folder otherwise everything go file will be under main package and now current structure seems okay to me

.github/workflows/go-test.yaml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: setup cluster
+        uses: ./.github/workflows/cluster-setup
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: install script
+        run: sudo cp kubectl-rook-ceph /usr/local/bin/kubectl-rook_ceph
```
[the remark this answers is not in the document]
subhamkrai: okay building binary in ci now

pkg/mons/mon-endpoints.go
travisn: Looks like the mon name is stripped out: 
```
$ kubectl rook-ceph mons
10.97.4.37:6789,10.108.120.61:6789,10.100.156.174:6789
```

cmd/commands/ceph.go
```diff
@@ -0,0 +1,34 @@
+/*
```
travisn, line 1: What about `cmd/ceph.go`? The `commands` subfolder seems redundant.

cmd/commands/root.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	cobra.CheckErr(RootCmd.Execute())
+}
+
+func init() {
+
+	// Here you will define your flags and configuration settings.
```
travisn, line 44: nit: remove the wording that sounds like a template
```suggestion
	// Define your flags and configuration settings.
```

pkg/cephrbd/ceph_rbd_commands.go
```diff
@@ -0,0 +1,147 @@
...
+	fmt.Printf(msg)
+	os.Exit(1)
+}
+
+var (
+	OperatorNamespace    string // operator namespae
+	CephClusterNamespace string // Cephcluster namespace
+)
+
+func RunCommandInOperatorPod(cmd *cobra.Command, args []string, operatorNamespace, clusterNamespace string) {
```
travisn, line 55: These functions to run a command in the operator pod don't seem specific to rbd. How about moving them to a separate package? For example, `pkg/k8sutil`.

pkg/mons/mon-endpoints.go
travisn: Correct, it was in the current shell script. We just need to maintain compatibility with the script output for the same commands. Perhaps in the future we could add an option that would keep the mon names in the output.

travisn (changes requested):

pkg/cephrbd/ceph_rbd_commands.go
subhamkrai: moved to `k8sutil`

cmd/commands/root.go
subhamkrai: done

cmd/commands/ceph.go
subhamkrai: the commands where not working if I had main.go and all other command files. so keeping as it for now.

pkg/mons/mon-endpoints.go
subhamkrai: done

cmd/commands/root.go
```diff
@@ -0,0 +1,56 @@
...
+	OperatorNamespace    string
+	CephClusterNamespace string
+)
+
+// rookCmd represents the rook command
+var RootCmd = &cobra.Command{
+	Use:     "rook-ceph",
+	Short:   "kubectl rook-ceph provides common management and troubleshooting tools for Ceph.",
+	Args:    cobra.MinimumNArgs(1),
+	Version: "v0.5.0",
```
travisn, line 33: Let's not put the version in the go code. Krew can keep track of the version at its level

.github/workflows/go-test.yaml
```diff
@@ -0,0 +1,39 @@
...
+          go-version: 1.19
+
+      - name: setup cluster
+        uses: ./.github/workflows/cluster-setup
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: build the binary
+        run: |
+          go build -o kubectl-rook-ceph cmd/main.go
```
travisn, line 31: Instead of running go build in this action, we need a makefile. Then developers can easily and consistently build locally during development, and the CI can also call `make`. But let's try to make it much simpler than the rook makefile. :)

travisn (changes requested):

cmd/commands/root.go
subhamkrai: okay removed

.github/workflows/go-test.yaml
subhamkrai: created makefile, I think this is one of the simplest makefile you will come across 😅

travisn: Simple makefiles are the best kind!

pkg/k8sutil/ceph_rbd_commands.go
```diff
@@ -0,0 +1,145 @@
+/*
```
travisn, line 1: Is there something specific to rbd in this file? Could we name it exec.go?

pkg/k8sutil/ceph_rbd_commands.go
```diff
@@ -0,0 +1,145 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	//k8s "k8s.io/client-go/kubernetes"
```
travisn, line 30: Do we have gofmt enabled? I thought it would remove the commented imports.

pkg/k8sutil/ceph_rbd_commands.go
```diff
@@ -0,0 +1,145 @@
...
+	fmt.Printf(msg)
+	os.Exit(1)
+}
+
+var (
+	OperatorNamespace    string // operator namespae
+	CephClusterNamespace string // Cephcluster namespace
+)
+
+func RunCommandInOperatorPod(cmd *cobra.Command, args []string, operatorNamespace, clusterNamespace string) {
```
travisn, line 53: If possible, let's keep the usage of cobra just in the `cmd` packages. So this method could be simpler if it is just a string command and the `[]string` args.

pkg/mons/mon-endpoints.go
```diff
@@ -0,0 +1,58 @@
...
+	_, client := k8sutil.GetKubeClient()
+
+	monCm, err := client.ConfigMaps(clusterNamespace).Get(context.TODO(), monConfigMap, v1.GetOptions{})
+	if err != nil {
+		log.Errorf("failed to get mon configmap %s %v", monConfigMap, err)
+	}
+
+	monData := monCm.Data["data"]
+
+	c := exec.Command("/bin/bash",
```
travisn, line 45: Instead of bringing in the bash parsing, could we using go methods to parse instead? Methods like `strings.Split()` are fairly simple and then could be easily unit tested as well.

Makefile
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+# Linux doesn't guarantee file ordering, so sort the files to make sure order is deterministic.
+# And in order to handle file paths with spaces, it's easiest to read the file names into an array.
+# Set locale `LC_ALL=C` because different OSes have different sort behavior;
+# `C` sorting order is based on the byte values,
+# Reference: https://blog.zhimingwang.org/macos-lc_collate-hunt
+
+
+build:
```
travisn, line 22: Let's also add gofmt to make sure formatting stays clean.

travisn (changes requested):

pkg/k8sutil/ceph_rbd_commands.go
subhamkrai: done

pkg/k8sutil/ceph_rbd_commands.go
subhamkrai: added gofmt in make build but it is not removing commented imports

pkg/k8sutil/ceph_rbd_commands.go
subhamkrai: done

pkg/mons/mon-endpoints.go
subhamkrai: done, using regex

Makefile
subhamkrai: added

cmd/commands/root.go
```diff
@@ -0,0 +1,55 @@
...
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	cobra.CheckErr(RootCmd.Execute())
+}
+
+func init() {
+
+	// Define your flags and configuration settings.
+
+	// Cobra supports Persistent Flags which will work for this command
```
travisn, line 45: How about removing lines 45-51? No need for the template comments

pkg/k8sutil/exec.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	//k8s "k8s.io/client-go/kubernetes"
```
travisn, line 28: nit: remove the commented imports?

pkg/mons/mon-endpoints.go
```diff
@@ -0,0 +1,46 @@
...
+	"regexp"
+
+	"github.com/kubectl-rook-ceph/pkg/k8sutil"
+	log "github.com/sirupsen/logrus"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const monConfigMap = "rook-ceph-mon-endpoints"
+
+func GetMonEndpoint(clusterNamespace string) {
```
travisn, line 30: Let's add a unit test for this method so we can establish the unit tests from the start. You'll just need to have the kube client as a parameter to the method.

travisn (changes requested):

cmd/commands/ceph.go
```diff
@@ -0,0 +1,34 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package command
+
+import (
+	k8sutil "github.com/kubectl-rook-ceph/pkg/k8sutil"
```
Madhu-1, line 20: this should be `github.com/rook/kubectl-rook-ceph/pkg/k8sutil` right?

Makefile
```diff
@@ -0,0 +1,26 @@
+# Copyright 2016 The Rook Authors. All rights reserved.
```
Madhu-1, line 1: 2023?

pkg/k8sutil/exec.go
```diff
@@ -0,0 +1,127 @@
...
+	cmd = append(cmd, args...)
+
+	if cmd[0] == "ceph" {
+		cmd = append(cmd, "--connect-timeout=10")
+	}
+
+	cmd = append(cmd, fmt.Sprintf("--conf=/var/lib/rook/%s/%s.config", clusterNamespace, clusterNamespace))
+
+	fmt.Println(strings.Join(cmd, " "))
+	fmt.Println()
```
Madhu-1, line 93: Why one more extra line? use `fmt.Printf("\n\n") if you want.

cmd/commands/root.go
subhamkrai: removed

pkg/k8sutil/exec.go
subhamkrai: removed

pkg/mons/mon-endpoints.go
subhamkrai: this method is not returning anything, how we should write the unit test?

cmd/commands/ceph.go
subhamkrai: right, that's good catch, updated

go.mod
```diff
@@ -0,0 +1,52 @@
+module github.com/kubectl-rook-ceph
```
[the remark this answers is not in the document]
subhamkrai: right,updated

Makefile
subhamkrai: done

pkg/k8sutil/exec.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil || len(list.Items) == 0 {
+		log.Error("failed to get rook operator pod where the command could be executed")
+		log.Fatal(err)
+	}
+
+	ExecCmd(cmd, rest, list.Items[0].Name, list.Items[0].Namespace, clusterNamespace, args)
+}
+
+// ExecCmd exec command on specific pod and wait the command's output.
+func ExecCmd(command string, restconfig *restclient.Config, podName, podNamespace, clusterNamespace string, args []string) {
```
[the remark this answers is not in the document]
subhamkrai: make sense, changed

pkg/k8sutil/exec.go
```diff
@@ -0,0 +1,127 @@
...
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	//k8s "k8s.io/client-go/kubernetes"
+
+	"k8s.io/client-go/kubernetes/scheme"
+	corev1client "k8s.io/client-go/kubernetes/typed/core/v1"
+
```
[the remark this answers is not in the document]
subhamkrai: removed

pkg/k8sutil/exec.go
subhamkrai: updated

cmd/commands/mons.go
```diff
@@ -0,0 +1,34 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package command
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/mons"
+	"github.com/spf13/cobra"
```
Madhu-1, line 21: new line between different packages is required.

pkg/k8sutil/exec.go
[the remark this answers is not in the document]
Madhu-1: Need to remove lines 27, 30, 32

pkg/k8sutil/exec.go
```diff
@@ -0,0 +1,123 @@
...
+	restclient "k8s.io/client-go/rest"
+
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/remotecommand"
+)
+
+func fail(msg string) {
+	fmt.Printf(msg)
+	os.Exit(1)
```
Madhu-1, line 40: you can replace this with log.Fatal i think

pkg/k8sutil/kube-client.go
```diff
@@ -0,0 +1,50 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package k8sutil
+
+import (
+	log "github.com/sirupsen/logrus"
+
```
Madhu-1, line 21: newline is not required here as all the imports are third party

pkg/mons/mon-endpoints.go
```diff
@@ -0,0 +1,46 @@
...
+	// Get Kubernetes Client
+	_, client := k8sutil.GetKubeClient()
+
+	monCm, err := client.ConfigMaps(clusterNamespace).Get(context.TODO(), monConfigMap, v1.GetOptions{})
+	if err != nil {
+		log.Errorf("failed to get mon configmap %s %v", monConfigMap, err)
```
Madhu-1, line 36: Return here as or use Fatal its an error?

Madhu-1 (commented):
Some nits. LGTM

pkg/k8sutil/exec.go
subhamkrai: actually above function is not used now, so removed

pkg/k8sutil/kube-client.go
subhamkrai: removed

pkg/mons/mon-endpoints.go
subhamkrai: used log.Fatalf

Madhu-1 (approved):

cmd/commands/root.go
```diff
@@ -0,0 +1,48 @@
...
+	cobra.CheckErr(RootCmd.Execute())
+}
+
+func init() {
+
+	// Define your flags and configuration settings.
+
+	RootCmd.PersistentFlags().StringVar(&KubeConfig, "kubeconfig", "", "kubernetes config path")
+	RootCmd.PersistentFlags().StringVar(&OperatorNamespace, "opeator-namespace", "rook-ceph", "Kubernetes namespace where rook operator is running")
+	RootCmd.PersistentFlags().StringVar(&CephClusterNamespace, "namespace", "rook-ceph", "Kubernetes namespace where ceph cluster is created")
```
parth-gr, line 47: How the krew cmd will run can you give a doc example for ceph health,

I can't how you might be calling the status, does it like calling the k8s client and then calling ceph cmd through it?

And one more question/doubt What if we just create and call the cmds from toolbox and wrap them into a go package?

pkg/k8sutil/exec.go
```diff
@@ -0,0 +1,115 @@
...
+	CephClusterNamespace string // Cephcluster namespace
+)
+
+func RunCommandInOperatorPod(cmd string, args []string, operatorNamespace, clusterNamespace string) {
+	// 1. Create Kubernetes Client
+	kubeconfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		clientcmd.NewDefaultClientConfigLoadingRules(),
+		&clientcmd.ConfigOverrides{},
+	)
+	rest, err := kubeconfig.ClientConfig()
```
travisn, line 46: Seems like these lines 42-49 are duplicated from what is in GetKubeClient(). Instead of duplicating, how about a different helper method instead of GetKubeClient() that gives back the `rest` and `client`? Or some other refactoring.

pkg/mons/mon-endpoints.go
travisn: As discussed, how about refactoring lines 40-44 to a separate method that can be unit tested.

pkg/k8sutil/exec.go
travisn: In the future we may also want to unit test commands where we can mock K8s and the exec commands, like we do in the Rook core. So I'd suggest that this method should take the kubeconfig as a parameter, then the core implementation takes parameters for the context that can be mocked. The kubeconfig would likely be initialized at the top layer under the cmd package.

travisn (changes requested):

cmd/commands/root.go
subhamkrai: Currently, we are using operator pod for running ceph commands and toolbox and operator pod are almost same. So we prefer rook operator pod as rook/rook also uses operator pod for running ceph command internally

pkg/k8sutil/exec.go
subhamkrai: now, I'm returning the rest and client from `GetKubeClient` only.
I'm not sure about second part could you refer some example

pkg/mons/mon-endpoints.go
subhamkrai: added a unit-test

pkg/mons/mon-endpoints_test.go
```diff
@@ -0,0 +1,53 @@
+/*
```
travisn, line 1: nit: How about naming them mon_endpoints.go and mon_endpoints_test.go? At least in rook we have this pattern of underscores instead of dashes for go filenames.

travisn (changes requested):

.github/workflows/go-test.yaml
```diff
@@ -0,0 +1,39 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: build the binary
+        run: |
+          make build
+          sudo cp kubectl-rook-ceph /usr/local/bin/kubectl-rook_ceph
+
+      - name: Test plugin with ceph commands
+        run: |
+          sleep 60s
```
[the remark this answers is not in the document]
subhamkrai: actually, sleep is not required as we are already waiting for the pod to be running

subhamkrai:
> > > In the future we may also want to unit test commands where we can mock K8s and the exec commands, like we do in the Rook core. So I'd suggest that this method should take the kubeconfig as a parameter, then the core implementation takes parameters for the context that can be mocked. The kubeconfig would likely be initialized at the top layer under the cmd package.
> 
> > now, I'm returning the rest and client from GetKubeClient only.
> > I'm not sure about second part could you refer some example
> 
> Rook has a method to [create the context](https://github.com/rook/rook/blob/master/cmd/rook/ceph/ceph.go#L65-L70), which initializes most of the context [here](https://github.com/rook/rook/blob/master/cmd/rook/rook/rook.go#L89). By default, it will create types that will exec real commands and expect a real k8s cluster. But then the unit tests can mock the exec and k8s methods by passing a different context, for example as created [here](https://github.com/rook/rook/blob/master/pkg/operator/test/client.go#L42) for the K8s tests, or search for `MockExecutor` for mocking the commands. This can all be done in a separate PR though.

I think I'm not very much confident about how that's works, may be discussed in call will make it clear.

pkg/mons/mon-endpoints_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func TestParseMonEndpoint(t *testing.T) {
```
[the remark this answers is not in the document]
subhamkrai: added a make command it was simple

pkg/mons/mon-endpoints_test.go
subhamkrai: done

cmd/commands/root.go
```diff
@@ -0,0 +1,48 @@
...
+func Execute() {
+	cobra.CheckErr(RootCmd.Execute())
+}
+
+func init() {
+
+	// Define your flags and configuration settings.
+
+	RootCmd.PersistentFlags().StringVar(&KubeConfig, "kubeconfig", "", "kubernetes config path")
+	RootCmd.PersistentFlags().StringVar(&OperatorNamespace, "opeator-namespace", "rook-ceph", "Kubernetes namespace where rook operator is running")
```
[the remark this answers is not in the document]
subhamkrai: done

Makefile
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+build:
+	gofmt -w $(shell find . -type f -name '*.go')
+	@echo
+	env GOOS=$(shell go env GOOS) GOARCH=$(shell go env GOARCH) go build -o kubectl-rook-ceph  cmd/main.go
```
travisn, line 19: After a build, the binary shows up in the local repo as a new file, and could be accidentally committed. How about if we output the binary to a `bin/` subfolder, then add `bin` to a `.gitignore`?

travisn (approved):
Approved with one more suggestion, thanks

Makefile
subhamkrai: done PTAL

travisn:
LGTM!

PR #80, opened 2023-01-04
ci: update ubuntu to 20.04

subhamkrai:


travisn (commented):
Was ubuntu-latest causing an issue?
Looks like the CI is not completing.

subhamkrai:
> Was ubuntu-latest causing an issue? Looks like the CI is not completing.

yes, if you see #79 ci's

travisn (approved):

PR #82, opened 2023-01-23
core: add operator commands in go

subhamkrai:

### DIFF
--- a/pkg/k8sutil/operator.go
+++ b/pkg/k8sutil/operator.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


pkg/k8sutil/operator.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func RestartDeployment(namespace, deploymentName string) {
+	// Get Kubernetes Client
+	_, clientset, _ := GetKubeClient()
+
+	deploymentsClient := clientset.AppsV1().Deployments(namespace)
+	data := fmt.Sprintf(`{"spec": {"template": {"metadata": {"annotations": {"kubectl.kubernetes.io/restartedAt": "%s"}}}}}`, time.Now().String())
+	_, err := deploymentsClient.Patch(context.TODO(), deploymentName, types.StrategicMergePatchType, []byte(data), v1.PatchOptions{})
```
travisn, line 35: To restart the operator, I thought we could delete the pod, and using labels to match the pod. Patching this deployment actually restarts the operator?

subhamkrai: yes, in my local testing it was restarting, I share the output today. 
We have 3ways to restart

1. using scale --replica=0
2. patch
3. delete

travisn (approved):

pkg/k8sutil/operator.go
subhamkrai: Screenshot from 2023-01-25 11-37-54

subhamkrai: we can see that operator has restarted using the above command. Merging now.

PR #83, opened 2023-01-23
core: add command rook and its sub-command

subhamkrai:
core: add command rook and its sub-command

### DIFF
--- a/cmd/commands/rook.go
+++ b/cmd/commands/rook.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package command
+
+import (
+	"fmt"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/rook"
+	"github.com/spf13/cobra"
+)
+
+// RookCmd represents the rook commands
+var RookCmd = &cobra.Command{
+	Use:   "rook",
+	Short: "Calls subcommands like `version`, `purge-osd, status` and etc.",
+	Args:  cobra.ExactArgs(1),
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Prints rook version",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		fmt.Println(exec.RunCommandInOperatorPod(context, "rook", []string{cmd.Use}, OperatorNamespace, CephClusterNamespace))
+	},
+}
+
+var purgeCmd = &cobra.Command{
+	Use:   "purge-osd",
+	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		forceflagValue := cmd.Flag("force").Value.String()
+		osdID := args[0]
+		rook.PurgeOsd(context, OperatorNamespace, CephClusterNamespace, osdID, forceflagValue)
+	},
+}
+
+var statusCmd = &cobra.Command{
+	Use:   "status",
+	Short: "Print the phase and conditions of the CephCluster CR",
+	Run: func(cmd *cobra.Command, args []string) {
+		rook.PrintCustomResourceStatus(CephClusterNamespace, args)
+	},
+}
+
+func init() {
+	RookCmd.AddCommand(versionCmd)
+	RookCmd.AddCommand(statusCmd)
+	RookCmd.AddCommand(purgeCmd)
+	purgeCmd.PersistentFlags().Bool("force", false, "force deletion of an OSD if the OSD still contains data")
+}
--- a/pkg/k8sutil/operator.go
+++ b/pkg/k8sutil/operator.go
@@ -19,9 +19,11 @@ package k8sutil
 import (
 	"context"
 	"fmt"
-	"log"
 	"time"
 
+	log "github.com/sirupsen/logrus"
+
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
@@ -35,6 +37,22 @@ func RestartDeployment(ctx *Context, namespace, deploymentName string) {
 	}
 
 	fmt.Printf("deployment.apps/%s restarted\n", deploymentName)
+}
+
+func WaitForOperatorPod(ctx *Context, operatorNamespace string) (corev1.Pod, error) {
+	opts := v1.ListOptions{LabelSelector: fmt.Sprintf("app=%s", "rook-ceph-operator")}
+	for i := 0; i < 60; i++ {
+		pod, _ := ctx.Clientset.CoreV1().Pods(operatorNamespace).List(context.TODO(), opts)
+		if pod.Items[0].Status.Phase == corev1.PodRunning && pod.Items[0].DeletionTimestamp.IsZero() {
+			return pod.Items[0], nil
+		}
+
+		fmt.Println("waiting for pod to be running")
+		time.Sleep(time.Second * 5)
+	}
+
+	return corev1.Pod{}, fmt.Errorf("failed to get rook operator pod where the command could be executed")
+
 }
 
 func UpdateConfigMap(ctx *Context, namespace, configMapName, key, value string) {
--- a/pkg/mons/mon_endpoints_test.go
+++ b/pkg/mons/mon_endpoints_test.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"testing"
 
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -30,23 +31,24 @@ func TestParseMonEndpoint(t *testing.T) {
 
 	newClient := fake.NewSimpleClientset
 	k8s := newClient()
+	context := k8sutil.Context{
+		Clientset: k8s,
+	}
 	ns := "rook-ceph"
 
 	cm := &v1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      monConfigMap,
+			Name:      MonConfigMap,
 			Namespace: ns,
 		},
 		Data: map[string]string{
 			"data": "10.96.52.53:6789",
 		},
 	}
-	_, err := k8s.CoreV1().ConfigMaps(ns).Create(ctx, cm, metav1.CreateOptions{})
-	assert.NoError(t, err)
-	c, err := k8s.CoreV1().ConfigMaps(ns).Get(ctx, monConfigMap, metav1.GetOptions{})
+	_, err := context.Clientset.CoreV1().ConfigMaps(ns).Create(ctx, cm, metav1.CreateOptions{})
 	assert.NoError(t, err)
 
-	monData := parseMonEndpoint(c.Data["data"])
+	monData := GetMonEndpoint(&context, ns)
 	assert.Equal(t, "10.96.52.53:6789", monData)
 	assert.NotEqual(t, "10.96.52.54:6789", monData)
 
--- a/pkg/rook/purge_osd.go
+++ b/pkg/rook/purge_osd.go
@@ -1,0 +1,49 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rook
+
+import (
+	ctx "context"
+	"fmt"
+	"log"
+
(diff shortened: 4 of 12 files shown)


pkg/mons/mon_endpoints_test.go
```diff
@@ -30,23 +31,24 @@ func TestParseMonEndpoint(t *testing.T) {
 
 	newClient := fake.NewSimpleClientset
 	k8s := newClient()
+	context := k8sutil.Context{
```
subhamkrai, line 34: unit test using fake client https://github.com/rook/kubectl-rook-ceph/pull/84#discussion_r1095905259

cmd/commands/rook.go
```diff
@@ -0,0 +1,70 @@
...
+	},
+}
+
+var purgeCmd = &cobra.Command{
+	Use:   "purge-osd",
+	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		flag := cmd.Flag("force").Value.String()
```
travisn, line 51: Does this flag works? This is used differently than I have seen flags used elsewhere in cobra, but maybe it's just a different pattern. In that case, maybe just a more specific name.
```suggestion
		forceFlag := cmd.Flag("force").Value.String()
```

pkg/execBashCommand/exec_bash_cmd.go
```diff
@@ -0,0 +1,39 @@
+/*
```
travisn, line 1: How about renaming this file: `pkg/exec/bash.go`?

pkg/execBashCommand/exec_bash_cmd.go
```diff
@@ -0,0 +1,39 @@
...
+
+package execbashcommand
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func ExecuteBashCommand(command string) string {
```
travisn, line 25: We need the context so it can be mocked.

pkg/k8sutil/exec.go
```diff
@@ -17,10 +17,11 @@ limitations under the License.
 package k8sutil
```
travisn, line 17: maybe move this file to `pkg/exec/exec.go`

pkg/rook/cr_status.go
```diff
@@ -0,0 +1,53 @@
+/*
```
travisn, line 1: How about naming the file `status.go`? It's always nice in go to have a single word filename if possible IMO.

pkg/rook/cr_status.go
```diff
@@ -0,0 +1,53 @@
...
+
+var scriptPrintSpecificCRStatus = `
+kubectl -n %s get %s -ojson | jq --monochrome-output '.items[].status'
+`
+
+var getCrdList = `
+kubectl -n  %s get crd | awk '{print $1}' | sed '1d'
+`
+
+func PrintCrStatus(_ *k8sutil.Context, clusterNamespace string, arg []string) {
```
travisn, line 35: nit
```suggestion
func PrintCustomResourceStatus(_ *k8sutil.Context, clusterNamespace string, arg []string) {
```

pkg/rook/purge_osd.go
```diff
@@ -0,0 +1,49 @@
...
+
+	monEndPoint := monCm.Data["data"]
+	cephArgs := []string{
+		"auth", "print-key", "client.admin",
+	}
+	adminKey := k8sutil.RunCommandInOperatorPod(context, "ceph", cephArgs, operatorNamespace, clusterNamespace)
+
+	cmd := "/bin/sh"
+	args := []string{
+		"-c",
```
travisn, line 44: Doesn't the `RunCommandInOperatorPod()` method already take care of the `/bin/sh` and `-c` arg? Or it would be nice if it did, so this caller only has to worry about its own command and arguments

travisn (changes requested):

cmd/commands/rook.go
subhamkrai: done

pkg/rook/cr_status.go
subhamkrai: ack, done

pkg/rook/purge_osd.go
subhamkrai: I don't we can since we run ceph/rbd commands too and if we want to take care of `/bin/sh -c` it will fail since it wants other commands in quotes.
like `/bin/sh -c  "ceph -s"` not sure who we can do  it here

travisn: How about a helper such as `RunShellCommandInOperatorPod()`? It might also be nice to take the env vars as a separate parameter. It would just be nice to push the details of command execution into the `exec` package.

pkg/rook/purge_osd.go
```diff
@@ -0,0 +1,50 @@
...
+	monCm, err := context.Clientset.CoreV1().ConfigMaps(clusterNamespace).Get(ctx.TODO(), mons.MonConfigMap, v1.GetOptions{})
+	if err != nil {
+		log.Fatalf("failed to get mon configmap %s %v", mons.MonConfigMap, err)
+	}
+
+	monEndPoint := monCm.Data["data"]
+	cephArgs := []string{
+		"auth", "print-key", "client.admin",
+	}
+	adminKey := bash.RunCommandInOperatorPod(context, "ceph", cephArgs, operatorNamespace, clusterNamespace)
```
travisn, line 41: The client.admin key should already be found in a file in the operator pod. Can we just reference that file instead of querying ceph for it?

pkg/rook/purge_osd.go
```diff
@@ -0,0 +1,50 @@
...
+
+	cmd := "/bin/sh"
+	args := []string{
+		"-c",
+		fmt.Sprintf("export ROOK_MON_ENDPOINTS=%s ROOK_CEPH_USERNAME=client.admin ROOK_CEPH_SECRET=%s ROOK_CONFIG_DIR=/var/lib/rook && rook ceph osd remove --osd-ids=%s --force-osd-removal=%s", monEndPoint, adminKey, osdId, flag),
```
travisn, line 46: All of this command has the potential to be seen in plain text so we need to make sure the secrets are in a file path.

pkg/rook/status.go
```diff
@@ -0,0 +1,53 @@
...
+
+var scriptPrintSpecificCRStatus = `
+kubectl -n %s get %s -ojson | jq --monochrome-output '.items[].status'
+`
+
+var getCrdList = `
+kubectl -n  %s get crd | awk '{print $1}' | sed '1d'
+`
+
+func PrintCustomResourceStatus(_ *k8sutil.Context, clusterNamespace string, arg []string) {
```
travisn, line 35: Why isn't the context passed yet? The PR that defined the context was merged, right?
```suggestion
func PrintCustomResourceStatus(context *k8sutil.Context, clusterNamespace string, arg []string) {
```

travisn (changes requested):

pkg/rook/purge_osd.go
subhamkrai: it fails with ```rookcmd: failed to read ceph secret file from "/var/lib/rook-ceph-mon": open /var/lib/rook-ceph-mon/secret.keyring: no such file or directory``` and it come from https://github.com/rook/rook/blob/11e80b7047553c04eeeb5c11f1c6c76fda381d5e/cmd/rook/ceph/osd.go#L369-L386

it looks for file `"/var/lib/rook-ceph-mon/secret.keyring"` but I checked it is not present inside rook op pod

pkg/rook/status.go
subhamkrai: it is not used anywhere in the function so discarded it.

pkg/rook/purge_osd.go
subhamkrai: I tried reading from the file following the above approach that didn't work. Are you suggesting any other way?

I didn't see the secret printing in the logs.

pkg/bash/bash.go
```diff
@@ -0,0 +1,39 @@
+/*
```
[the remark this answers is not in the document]
subhamkrai: done

pkg/rook/purge_osd.go
subhamkrai:
```yaml
023-02-20 10:54:25.999916 W | cephcmd: loaded admin secret from env var ROOK_CEPH_SECRET instead of from file
2023-02-20 10:54:26.001142 I | rookcmd: starting Rook v1.11.0-alpha.0.20.g10abbd0d8 with arguments 'rook ceph osd remove --osd-ids=0 --force-osd-removal=true'
```

travisn: I'm thinking of these vars that the operator uses. This plugin could assume the files exist (depending on the namespace), then no need to look up the secret value.
```
--cluster=rook-ceph --conf=/var/lib/rook/rook-ceph/rook-ceph.config --name=client.admin --keyring=/var/lib/rook/rook-ceph/client.admin.keyring
```

travisn (changes requested):

pkg/rook/status.go
```diff
@@ -0,0 +1,53 @@
...
+	if len(arg) == 1 && arg[0] == "all" {
+		command := fmt.Sprintf(getCrdList, clusterNamespace)
+		allCrs := strings.Split(exec.ExecuteBashCommand(command), "\n")
```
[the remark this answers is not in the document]
subhamkrai: done

pkg/rook/status.go
```diff
@@ -0,0 +1,53 @@
...
+		fmt.Println(allCrs[0])
```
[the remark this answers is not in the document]
subhamkrai: no, first it prints the name of the CR and then the status

pkg/k8sutil/operator.go
```diff
@@ -37,6 +39,16 @@ func RestartDeployment(ctx *Context, namespace, deploymentName string) {
 	fmt.Printf("deployment.apps/%s restarted\n", deploymentName)
 }
 
+func GetOperatorPodFromPodList(ctx *Context, operatorNamespace string) *corev1.PodList {
+	opts := v1.ListOptions{LabelSelector: fmt.Sprintf("app=%s", "rook-ceph-operator")}
+	list, err := ctx.Clientset.CoreV1().Pods(operatorNamespace).List(context.TODO(), opts)
+	if err != nil || len(list.Items) == 0 {
+		log.Error("failed to get rook operator pod where the command could be executed")
+		log.Fatal(err)
+	}
```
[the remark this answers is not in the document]
subhamkrai: added a check

cmd/commands/rook.go
```diff
@@ -0,0 +1,70 @@
...
+	Args:  cobra.ExactArgs(1),
+}
+
+var versionCmd = &cobra.Command{
+	Use:   "version",
+	Short: "Prints rook version",
+	Args:  cobra.NoArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		args = append(args, "version")
```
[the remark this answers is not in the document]
subhamkrai: right, removed

pkg/rook/status.go
subhamkrai: removed

pkg/rook/purge_osd.go
subhamkrai: those args are used for ceph command, here we are running rook command that triggers the osd removal

pkg/k8sutil/operator.go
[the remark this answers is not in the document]
travisn: How about if this method just returns the running pod, instead of the pod list. Otherwise, the wrong pod might be chosen.

pkg/rook/purge_osd.go
travisn: While the secret can be in an env var, the intention with #11434 is that callers will move to the new approach of using a file for the secret, to avoid the command line being exposed with the secret. As seen in that PR, the osd purge job will mount the secret to the path `/var/lib/rook-ceph-mon`, which will be read from that path [here](https://github.com/rook/rook/blob/release-1.10/cmd/rook/ceph/osd.go#L201). I wonder if this plugin could first run a command to copy the file into that path so the purge command will then just find it there.

travisn (changes requested):

pkg/rook/purge_osd.go
travisn: Although a complication is that this would require at least Rook v1.10. So if the plugin is targeting an older version of rook, that wouldn't work. We should figure out how to target different versions of rook (separate from this PR).

pkg/k8sutil/operator.go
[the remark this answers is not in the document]
subhamkrai: done

pkg/k8sutil/operator.go
```diff
@@ -37,6 +39,16 @@ func RestartDeployment(ctx *Context, namespace, deploymentName string) {
 	fmt.Printf("deployment.apps/%s restarted\n", deploymentName)
 }
 
+func GetOperatorPodFromPodList(ctx *Context, operatorNamespace string) corev1.Pod {
```
travisn, line 42:
```suggestion
func GetOperatorPod(ctx *Context, operatorNamespace string) corev1.Pod {
```

pkg/k8sutil/operator.go
```diff
@@ -37,6 +39,16 @@ func RestartDeployment(ctx *Context, namespace, deploymentName string) {
...
+	opts := v1.ListOptions{LabelSelector: fmt.Sprintf("app=%s", "rook-ceph-operator")}
+	pods, err := ctx.Clientset.CoreV1().Pods(operatorNamespace).List(context.TODO(), opts)
+	if err != nil || len(pods.Items) == 0 || !pods.Items[0].DeletionTimestamp.IsZero() {
+		log.Error("failed to get rook operator pod where the command could be executed")
+		log.Fatal(err)
+	}
+	return pods.Items[0]
```
travisn, line 49: How about iterating through the list of pods and finding the pod with status Running and not deleted? Otherwise, we have the problem again to get the terminating pod in that corner case.

pkg/rook/purge_osd.go
travisn: As discussed, let's leave it this way for now and consider a future update based on version.

travisn (changes requested):
Just a couple final changes...

pkg/k8sutil/operator.go
subhamkrai: done

pkg/k8sutil/operator.go
```diff
@@ -37,6 +39,22 @@ func RestartDeployment(ctx *Context, namespace, deploymentName string) {
 	fmt.Printf("deployment.apps/%s restarted\n", deploymentName)
 }
 
+func GetOperatorPod(ctx *Context, operatorNamespace string) (corev1.Pod, error) {
```
travisn, line 42: If we're waiting, it's helpful to have that in the method name
```suggestion
func WaitForOperatorPod(ctx *Context, operatorNamespace string) (corev1.Pod, error) {
```

travisn (approved):
Approved with a small suggestion

subhamkrai:
Merging, since the nit is resolved.

PR #84, opened 2023-01-24
core: place context in central place similar to rook

subhamkrai:


BlaineEXE:
Consider this example of how context can be used with Cobra for a traditional CLI tool like this : https://github.com/BlaineEXE/go-cobra-with-interrupt

pkg/k8sutil/exec.go
```diff
@@ -36,28 +34,21 @@ var (
...
 }
 
 // ExecCmd exec command on specific pod and wait the command's output.
-func ExecCmd(restconfig *restclient.Config, command, podName, containerName, podNamespace, clusterNamespace string, args []string) {
-	// Create a Kubernetes core/v1 client.
-	coreclient, err := corev1client.NewForConfig(restconfig)
-	if err != nil {
-		log.Fatal(err)
-	}
+func ExecCmd(ctx *Context, command, podName, containerName, podNamespace, clusterNamespace string, args []string) {
```
travisn, line 50:
```suggestion
func ExecCmdInPod(ctx *Context, command, podName, containerName, podNamespace, clusterNamespace string, args []string) {
```

pkg/k8sutil/context.go
```diff
@@ -0,0 +1,34 @@
...
+
+package k8sutil
+
+import (
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Context struct {
```
travisn, line 25: This struct will have other types of contexts such as for the CLI. While there may be a better package name for this file, it doesn't seem worth creating a package just for that. Perhaps it will be clear in the future if there is a better package as more functionality is added.

pkg/k8sutil/exec.go
```diff
@@ -36,28 +34,21 @@ var (
 	CephClusterNamespace string // Cephcluster namespace
 )
 
-func RunCommandInOperatorPod(cmd string, args []string, operatorNamespace, clusterNamespace string) {
-
-	// Get Kubernetes Client
-	rest, _, client := GetKubeClient()
+func RunCommandInOperatorPod(ctx *Context, cmd string, args []string, operatorNamespace, clusterNamespace string) {
```
travisn, line 37: In Rook it's a more common pattern to use `context` as the variable name, and use `ctx` only when there is a conflict with a package name, or similar. So it would be nice to keep that same pattern and use `context` if we can.
```suggestion
func RunCommandInOperatorPod(context *Context, cmd string, args []string, operatorNamespace, clusterNamespace string) {
```

subhamkrai:
> Consider this example of how context can be used with Cobra for a traditional CLI tool like this : https://github.com/BlaineEXE/go-cobra-with-interrupt

thank @BlaineEXE for the sample example. will update in this pr only

pkg/k8sutil/context.go
subhamkrai: I saw in rook we have a different package so I followed that but I'm happy to revert too, let me know.

pkg/k8sutil/exec.go
subhamkrai: here I think we'll have to use  `ctx` only but I'll follow the rook's approach where I can

subhamkrai:
> > Consider this example of how context can be used with Cobra for a traditional CLI tool like this : https://github.com/BlaineEXE/go-cobra-with-interrupt
> 
> thank @BlaineEXE for the sample example. will update in this pr only

Now, I'm thinking I  should do in separate pr since, this pr is changing the structure and effects multiple files, it will be better we get this pr soon and then others prs.

BlaineEXE:
If this PR is just to add context, what changes here are needed that must be related to something else?

subhamkrai:
> If this PR is just to add context, what changes here are needed that must be related to something else?

I think all the changes for context are here and for cobra context. I'll open separate pr.

pkg/k8sutil/exec.go
subhamkrai: done

cmd/commands/root.go
```diff
@@ -46,3 +52,33 @@ func init() {
...
+		log.Fatal(err)
+	}
+
+	context.Clientset, err = k8s.NewForConfig(context.KubeConfig)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// // Create a Kubernetes core/v1 client.
+	// context.Client, err = v1.NewForConfig(context.KubeConfig)
```
travisn, line 78: delete the commented section?

pkg/mons/mon_endpoints.go
```diff
@@ -27,11 +27,8 @@ import (
 
 const monConfigMap = "rook-ceph-mon-endpoints"
 
-func GetMonEndpoint(clusterNamespace string) {
-	// Get Kubernetes Client
-	_, _, client := k8sutil.GetKubeClient()
-
-	monCm, err := client.ConfigMaps(clusterNamespace).Get(context.TODO(), monConfigMap, v1.GetOptions{})
+func GetMonEndpoint(ctx *k8sutil.Context, clusterNamespace string) {
```
travisn, line 30: The rook pattern is to import like this so the `context` var name can be used consistently:
```
ctx "context"
```

```suggestion
func GetMonEndpoint(context *k8sutil.Context, clusterNamespace string) {
```

travisn: Is there a unit test that could use a fake k8s context now? In a separate PR is fine though.

travisn (changes requested):
A couple small suggestions

travisn (approved):

PR #85, opened 2023-01-30
core: add start/stop debug command

subhamkrai:

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -34,6 +34,7 @@ jobs:
 
       - name: Test plugin with ceph commands
         run: |
+          set -e
           kubectl rook-ceph ceph status
           kubectl rook-ceph mons
           kubectl rook-ceph rbd ls replicapool
@@ -43,7 +44,11 @@ jobs:
           # let's wait for operator pod to be restart
           kubectl -n rook-ceph wait --for=delete pod/$POD --timeout=100s
           tests/github-action-helper.sh wait_for_operator_pod_to_be_ready_state_default
+          kubectl rook_ceph debug start rook-ceph-osd-0
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug rook-ceph
 
+          kubectl rook_ceph debug stop rook-ceph-osd-0
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0 rook-ceph
           kubectl rook-ceph operator set ROOK_LOG_LEVEL DEBUG
           kubectl rook-ceph rook version
           kubectl rook-ceph rook status
--- a/cmd/commands/debug.go
+++ b/cmd/commands/debug.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package command
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/debug"
+	"github.com/spf13/cobra"
+)
+
+// OperatorCmd represents the operator commands
+var DebugCmd = &cobra.Command{
+	Use:                "debug",
+	Short:              "Debug a deployment by scaling it down and creating a debug copy. This is supported for mons and OSDs only",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+}
+
+var startDebugCmd = &cobra.Command{
+	Use:   "start",
+	Short: "Start debugging a deployment with an optional alternative ceph container image",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		alternateImage := cmd.Flag("alternate-image").Value.String()
+		debug.StartDebug(context, CephClusterNamespace, args[0], alternateImage)
+	},
+}
+var stopDebugCmd = &cobra.Command{
+	Use:   "stop",
+	Short: "Stop debugging a deployment",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		debug.StopDebug(context, CephClusterNamespace, args[0])
+	},
+}
+
+func init() {
+	DebugCmd.AddCommand(startDebugCmd)
+	startDebugCmd.Flags().String("alternate-image", "", "To create deployment with alternate image")
+	DebugCmd.AddCommand(stopDebugCmd)
+}
--- a/pkg/debug/start_debug.go
+++ b/pkg/debug/start_debug.go
@@ -1,0 +1,169 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package debug
+
+import (
+	ctx "context"
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) {
+
+	err := startDebug(context, clusterNamespace, deploymentName, alternateImageValue)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) error {
+	deployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
+	if err != nil {
+		return fmt.Errorf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
+	}
+
+	if alternateImageValue != "" {
+		log.Printf("setting debug image to %s\n", alternateImageValue)
+		deployment.Spec.Template.Spec.Containers[0].Image = alternateImageValue
+	}
+
+	labels := deployment.Labels
+	labels["ceph.rook.io/do-not-reconcile"] = "true"
+
+	deployment.Spec.Template.Spec.Containers[0].LivenessProbe = nil
+	deployment.Spec.Template.Spec.Containers[0].StartupProbe = nil
+
+	fmt.Println("setting debug command to main container")
+
+	deployment.Spec.Template.Spec.Containers[0].Command = []string{"sleep", "infinity"}
+	deployment.Spec.Template.Spec.Containers[0].Args = []string{}
+
+	if err := updateDeployment(context, clusterNamespace, deployment); err != nil {
+		return fmt.Errorf("Failed to update deployment %s. %v\n", deployment.Name, err)
+	}
+
+	deploymentPodName, err := waitForPodToRun(context, clusterNamespace, deployment.Spec)
+	if err != nil {
+		fmt.Println(err)
+		return err
+	}
+
+	if err := setDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
+		return err
+	}
+
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deploymentPodName)
+
+	err = waitForPodDeletion(context, clusterNamespace, deploymentPodName)
+	if err != nil {
+		fmt.Println(err)
+		return err
+	}
+
+	debugDeploymentSpec := &appsv1.Deployment{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-debug", deploymentName),
+			Namespace: clusterNamespace,
+			Labels:    labels,
+		},
+		Spec: deployment.Spec,
+	}
+
+	debugDeployment, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Create(ctx.TODO(), debugDeploymentSpec, v1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("Error creating deployment %s. %v\n", debugDeploymentSpec, err)
+	}
+	fmt.Printf("ensure the debug deployment %s is scaled up\n", deploymentName)
+
+	if err := setDeploymentScale(context, clusterNamespace, debugDeployment.Name, 1); err != nil {
+		return err
+	}
+	return nil
+}
+
+func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentName string, scaleCount int) error {
+	scale := &autoscalingv1.Scale{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      deploymentName,
+			Namespace: clusterNamespace,
+		},
+		Spec: autoscalingv1.ScaleSpec{
+			Replicas: int32(scaleCount),
+		},
+	}
+	_, err := context.Clientset.AppsV1().Deployments(clusterNamespace).UpdateScale(ctx.TODO(), deploymentName, scale, v1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to update scale of deployment %s. %v\n", deploymentName, err)
+	}
+	return nil
+}
+
+func verifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
+	deployment, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Get(ctx.TODO(), deploymentName, v1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return deployment, nil
+}
+
+func updateDeployment(context *k8sutil.Context, clusterNamespace string, deploymentName *appsv1.Deployment) error {
+	_, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Update(ctx.TODO(), deploymentName, v1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func waitForPodToRun(context *k8sutil.Context, clusterNamespace string, deploymentSpec appsv1.DeploymentSpec) (string, error) {
+	labelSelector := fmt.Sprintf("ceph_daemon_type=%s,ceph_daemon_id=%s", deploymentSpec.Template.Labels["ceph_daemon_type"], deploymentSpec.Template.Labels["ceph_daemon_id"])
+	for i := 0; i < 60; i++ {
+		pod, _ := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), v1.ListOptions{LabelSelector: labelSelector})
+		if pod.Items[0].Status.Phase == corev1.PodRunning && pod.Items[0].DeletionTimestamp.IsZero() {
+			return pod.Items[0].Name, nil
+		}
+
+		fmt.Println("waiting for pod to be running")
+		time.Sleep(time.Second * 5)
+	}
+
+	return "", fmt.Errorf("No pod with labels matching %s:%s", deploymentSpec.Template.Labels, deploymentSpec.Template.Labels)
+
+}
+
+func waitForPodDeletion(context *k8sutil.Context, clusterNamespace, podName string) error {
+	for i := 0; i < 60; i++ {
+		_, err := context.Clientset.CoreV1().Pods(clusterNamespace).Get(ctx.TODO(), podName, v1.GetOptions{})
+		if kerrors.IsNotFound(err) {
+			return nil
+		}
+
+		fmt.Printf("waiting for pod %q to be deleted\n", podName)
+		time.Sleep(time.Second * 5)
+	}
+
+	return fmt.Errorf("failed to delete pod %s", podName)
+}
--- a/pkg/debug/stop_debug.go
+++ b/pkg/debug/stop_debug.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package debug
+
+import (
+	ctx "context"
+	"fmt"
+	"os"
+	"strings"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+)
+
+func StopDebug(context *k8sutil.Context, clusterNamespace, deploymentName string) {
+
+	err := stopDebug(context, clusterNamespace, deploymentName)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func stopDebug(context *k8sutil.Context, clusterNamespace, deploymentName string) error {
+
+	if !strings.HasSuffix(deploymentName, "-debug") {
+		deploymentName = deploymentName + "-debug"
+	}
+
+	debugDeployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
+	if err != nil {
+		return fmt.Errorf("Missing mon or osd debug deployment name %s. %v\n", deploymentName, err)
+	}
+
+	fmt.Printf("removing debug mode from deployment %s\n", debugDeployment.Name)
+	err = context.Clientset.AppsV1().Deployments(clusterNamespace).Delete(ctx.TODO(), debugDeployment.Name, v1.DeleteOptions{})
+	if err != nil && !kerrors.IsNotFound(err) {
+		return fmt.Errorf("Error deleting deployment %s: %v", debugDeployment.Name, err)
+	}
+
+	original_deployment_name := strings.ReplaceAll(deploymentName, "-debug", "")
+	if err := setDeploymentScale(context, clusterNamespace, original_deployment_name, 1); err != nil {
+		return err
+	}
+	return nil
+}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -29,5 +29,6 @@ func addcommands() {
 		command.RbdCmd,
 		command.OperatorCmd,
 		command.RookCmd,
+		command.DebugCmd,
 	)
 }


pkg/debug/start_debug.go
```diff
@@ -0,0 +1,160 @@
...
+
+	if err := updateDeployment(clientset, clusterNamespace, deployment); err != nil {
+		os.Exit(1)
+	}
+
+	deployment_Pod := exec.Command("/bin/bash",
+		"-e", // Exit immediately if a command exits with a non-zero status.
+		"-m", // Terminal job control, allows job to be terminated by SIGTERM
+		"-c", // Command to run
+		fmt.Sprintf(getPodFromDeployment, clusterNamespace, deploymentName),
```
travisn, line 80: Instead of the cli command, perhaps could do something similar to rook's `PodsRunningWithLabel()` method to look up the running pod from a label.

.github/workflows/go-test.yaml
```diff
@@ -43,5 +43,6 @@ jobs:
           # let's wait for operator pod to be restart
           kubectl -n rook-ceph wait --for=delete pod/$POD --timeout=100s
           tests/github-action-helper.sh wait_for_operator_pod_to_be_ready_state_default
+          kubectl rook_ceph debug start rook-ceph-osd-0
```
travisn, line 46: What tests can be added? Probably to wait for the debug pod to start, then also stop the debug pod and wait for it to stop?

travisn (changes requested):
Just took a quick look since it's in draft state

pkg/debug/start_debug.go
subhamkrai: done

.github/workflows/go-test.yaml
subhamkrai: added

.github/workflows/go-test.yaml
```diff
@@ -43,5 +44,8 @@ jobs:
...
+          tests/github-action-helper.sh wait_for_deployment_to_be_running rook-ceph-osd-0-debug rook-ceph
 
+          kubectl rook_ceph debug stop rook-ceph-osd-0
```
travisn, line 50: How about a step to wait for it to stop and the original pod to start up again?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+	"os/exec"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
+)
+
+var getPodFromDeployment = `
```
travisn, line 33: not used anymore?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+		deploymentSpec.Template.Spec.Containers[0].Image = flagValue
+	}
+
+	fmt.Println("setting debug command to main container")
+
+	deploymentSpec.Template.Spec.Containers[0].Command = []string{"sleep", "infinity"}
+	deploymentSpec.Template.Spec.Containers[0].Args = []string{}
+
+	if err := updateDeployment(clientset, clusterNamespace, deployment); err != nil {
+		os.Exit(1)
```
travisn, line 73: Whenever we exit, let's first print the error first. Otherwise, the user doesn't know what happened.

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deployment_Pod_Name)
+
+	waitForPodToDeleted := exec.Command("/bin/bash",
+		"-e", // Exit immediately if a command exits with a non-zero status.
+		"-m", // Terminal job control, allows job to be terminated by SIGTERM
+		"-c", // Command to run
+		fmt.Sprintf(scriptForPodDeletion, clusterNamespace, deployment_Pod_Name),
+	)
+	_, err = waitForPodToDeleted.Output()
```
travisn, line 93: Instead of this script, can we use a similar K8s helper to wait for the pod to be deleted?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := getAndUpdatetheScale(clientset, clusterNamespace, debugDeployment.Name, 1); err != nil {
+		os.Exit(1)
+	}
+
+}
+
+func getAndUpdatetheScale(clientset *k8s.Clientset, clusterNamespace, deploymentName string, scaleCount int) error {
+	scale, err := clientset.AppsV1().Deployments(clusterNamespace).GetScale(context.TODO(), deploymentName, v1.GetOptions{})
+	if err != nil {
+		fmt.Printf(" failed to get scale of deployment %s. %v\n", deploymentName, err)
```
travisn, line 124: Is the leading space intended on some of these messages?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+
+	scale.Spec.Replicas = int32(scaleCount)
+	_, err = clientset.AppsV1().Deployments(clusterNamespace).UpdateScale(context.TODO(), deploymentName, scale, v1.UpdateOptions{})
+	if err != nil {
+		fmt.Printf(" failed to update scale of deployment %s. %v\n", deploymentName, err)
```
travisn, line 131: In a future PR let's consider calling a helper whenever we print something. Then we can add a prefix with the timestamp, which is helpful in most cases to see what happened. Also, the helper could give us the different colors of output.

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+		return err
+	}
+	return nil
+}
+
+func verifyDeploymentExists(clientset *k8s.Clientset, clusterNamespace, deploymentName string) (*appsv1.Deployment, bool) {
+	deployment, err := clientset.AppsV1().Deployments(clusterNamespace).Get(context.TODO(), deploymentName, v1.GetOptions{})
+	if err != nil {
+		fmt.Printf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
```
travisn, line 140: If we just return this as an error, then the caller can print the error before exiting.

travisn (changes requested):

pkg/debug/start_debug.go
subhamkrai: removed

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: no, removed

pkg/debug/start_debug.go
subhamkrai: will follow up this in a new pr

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,167 @@
...
+	"time"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8s "k8s.io/client-go/kubernetes"
+)
+
+func StartDebug(clusterNamespace, deploymentName, flagValue string) {
```
travisn, line 33: This needs to be updated for the new context?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,167 @@
...
+	// Get Kubernetes Client
+	_, clientset, _ := k8sutil.GetKubeClient()
+
+	deployment, err := verifyDeploymentExists(clientset, clusterNamespace, deploymentName)
+	if err != nil {
+		fmt.Printf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
+		os.Exit(1)
+	}
+
+	deploymentSpec := deployment.Spec
```
travisn, line 43: nit: i'd suggest just still use `deployment.Spec`. This var doesn't simplify much
```suggestion
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,167 @@
...
+	deploymentSpec.Template.Spec.Containers[0].Args = []string{}
+
+	if err := updateDeployment(clientset, clusterNamespace, deployment); err != nil {
+		fmt.Printf("Failed to update deployment %s. %v\n", deployment.Name, err)
+		os.Exit(1)
+	}
+
+	deployment_Pod_Name, err := podRunningWithLabel(clientset, clusterNamespace, deploymentSpec)
+	if deployment_Pod_Name == "" || err != nil {
+		os.Exit(1)
```
travisn, line 72: can we print an error?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,167 @@
...
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deployment_Pod_Name)
+
+	err = waitForPodDeletion(clientset, clusterNamespace, deployment_Pod_Name)
+	if err != nil {
+		fmt.Println(err)
+	}
```
travisn, line 85: We don't need to exit after this error?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,167 @@
...
+	return "", fmt.Errorf("No pod with labels matching %s:%s", deploymentSpec.Template.Labels, deploymentSpec.Template.Labels)
+}
+
+func waitForPodDeletion(clientset *k8s.Clientset, clusterNamespace, podName string) error {
+	for i := 0; i < 60; i++ {
+		err := clientset.CoreV1().Pods(clusterNamespace).Delete(context.TODO(), podName, v1.DeleteOptions{})
+		if err != nil {
+			return nil
+		}
+		time.Sleep(time.Second * 5)
```
travisn, line 163: Whenever we are waiting, let's print a message so they know something is happening

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StopDebug(clusterNamespace, deploymentName string) {
```
travisn, line 29: context needed

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,54 @@
...
+	// Get Kubernetes Client
+	_, clientset, _ := k8sutil.GetKubeClient()
+
+	if !strings.Contains(deploymentName, "-debug") {
+		deploymentName = deploymentName + "-debug"
+	}
+
+	debug_deployment, err := verifyDeploymentExists(clientset, clusterNamespace, deploymentName)
+	if err != nil {
+		fmt.Printf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
```
travisn, line 39: If the error is that the debug deployment doesn't exist, let's continue and scale up the deployment. No need to fail when deleting something that doesn't exist.

travisn (changes requested):

pkg/debug/start_debug.go
subhamkrai: updated

pkg/debug/start_debug.go
subhamkrai: updated

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: added

pkg/debug/stop_debug.go
subhamkrai: updated

pkg/debug/stop_debug.go
subhamkrai: updated

cmd/commands/debug.go
```diff
@@ -0,0 +1,60 @@
...
+var startDebugCmd = &cobra.Command{
+	Use:   "start",
+	Short: "Start debugging a deployment with an optional alternative ceph container image",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		flagValue := ""
+		flag := cmd.Flags().Lookup("alternate-image").Changed
+		if flag {
+			flagValue = cmd.Flag("alternate-image").Value.String()
```
travisn, line 41: Do we need to check the `Changed` on the flag? Wouldn't it be blank by default anyway, just like the default value on line 38?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,163 @@
...
+	"log"
+	"os"
+	"time"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, flagValue string) {
```
travisn, line 31: The param is for the alternate image, right? Specific var names will help readability
```suggestion
func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImage string) {
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,163 @@
...
+	fmt.Printf("ensure the debug deployment %s is scaled up\n", deploymentName)
+
+	if err := getAndUpdatetheScale(context, clusterNamespace, debugDeployment.Name, 1); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+}
+
+func getAndUpdatetheScale(context *k8sutil.Context, clusterNamespace, deploymentName string, scaleCount int) error {
```
travisn, line 106: how about this name?
```suggestion
func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentName string, scaleCount int) error {
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,163 @@
...
+	_, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Update(ctx.TODO(), deploymentName, v1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func podRunningWithLabel(context *k8sutil.Context, clusterNamespace string, deploymentSpec appsv1.DeploymentSpec) (string, error) {
+
+	pods, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), v1.ListOptions{})
```
travisn, line 138: In the `ListOptions`, you can use the label selectors directly, so you will only get the pod needed instead of all pods in the namespace.

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,163 @@
...
+			pod.Labels["ceph_daemon_id"] == deploymentSpec.Template.Labels["ceph_daemon_id"] {
+			return pod.Name, nil
+		}
+	}
+	return "", fmt.Errorf("No pod with labels matching %s:%s", deploymentSpec.Template.Labels, deploymentSpec.Template.Labels)
+}
+
+func waitForPodDeletion(context *k8sutil.Context, clusterNamespace, podName string) error {
+	for i := 0; i < 60; i++ {
+		err := context.Clientset.CoreV1().Pods(clusterNamespace).Delete(ctx.TODO(), podName, v1.DeleteOptions{})
```
travisn, line 154: How about deleting the pod once, then get it in each iteration while we wait? No need to delete it in every iteration.

travisn (changes requested):

cmd/commands/debug.go
subhamkrai: L39 check the flag change value

pkg/debug/start_debug.go
subhamkrai: here we don't know what deployment name the user will be passing, for example, it can be
`rook-ceph-mon-{a,b,c...} and rook-ceph-osd-{0,1,2,..}` so that's why I'm comparing the labels between the deployment passed and the pod we are getting

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: changed

pkg/debug/start_debug.go
subhamkrai: updated

cmd/commands/debug.go
travisn: I don't understand why we need to check the `Changed` field? Why not just always get the laternate-image.Value.String()`?

cmd/commands/debug.go
```diff
@@ -0,0 +1,60 @@
...
+	Args:               cobra.ExactArgs(1),
+}
+
+var startDebugCmd = &cobra.Command{
+	Use:   "start",
+	Short: "Start debugging a deployment with an optional alternative ceph container image",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		context := GetContext()
+		flagValue := ""
```
travisn, line 38: Let's use variables to describe what they are instead of generic names
```suggestion
		alternateImage := ""
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deployment_Pod_Name)
+
+	err = waitForPodDeletion(context, clusterNamespace, deployment_Pod_Name)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	debug_deployment := &appsv1.Deployment{
```
travisn, line 83: Vars are usually named with camelCase, otherwise it feels like bash :)
```suggestion
	debugDeployment := &appsv1.Deployment{
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if err := setDeploymentScale(context, clusterNamespace, debugDeployment.Name, 1); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+}
+
+func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentName string, scaleCount int) error {
+	scale, err := context.Clientset.AppsV1().Deployments(clusterNamespace).GetScale(ctx.TODO(), deploymentName, v1.GetOptions{})
```
travisn, line 108: Instead of getting the scale from the API, can we just declare the scale struct?

pkg/debug/start_debug.go
travisn: Something like this should work:
```suggestion
	labelSelector := fmt.Sprintf("ceph_daemon_type=%s,ceph_daemon_id=%s", deploymentSpec.Template.Labels["ceph_daemon_type"], deploymentSpec.Template.Labels["ceph_daemon_id"])
	pods, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), v1.ListOptions{LabelSelector: labelSelector})
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func waitForPodDeletion(context *k8sutil.Context, clusterNamespace, podName string) error {
+	for i := 0; i < 60; i++ {
+		_, err := context.Clientset.CoreV1().Pods(clusterNamespace).Get(ctx.TODO(), podName, v1.GetOptions{})
+		if kerrors.IsNotFound(err) {
+			return nil
+		}
+
+		err = context.Clientset.CoreV1().Pods(clusterNamespace).Delete(ctx.TODO(), podName, v1.DeleteOptions{})
```
travisn, line 160: You could move this before the `for` loop, then your loop is just to get and sleep

travisn: Actually, do you even need to delete the pod? I think you don't even want to delete the pod since the deployment scale-down should automatically delete it. So this method is just waiting for the pod to go away.

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,51 @@
...
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StopDebug(context *k8sutil.Context, clusterNamespace, deploymentName string) {
+
+	if !strings.Contains(deploymentName, "-debug") {
+		deploymentName = deploymentName + "-debug"
+	}
+
+	debug_deployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
```
travisn, line 35:
```suggestion
	debugDeployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
```

travisn (changes requested):

pkg/debug/stop_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: right, update

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
subhamkrai: thanks, it worked

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,162 @@
...
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	if err := setDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deploymentPodName)
```
travisn, line 76: Do we have about the same output from the golang commands as in the bash implementation? It would be nice to be consistent, although it doesn't have to be exact. It's just a general question, not about this specific log line.

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,162 @@
...
+
+	deployment.Spec.Template.Spec.Containers[0].Command = []string{"sleep", "infinity"}
+	deployment.Spec.Template.Spec.Containers[0].Args = []string{}
+
+	if err := updateDeployment(context, clusterNamespace, deployment); err != nil {
+		fmt.Printf("Failed to update deployment %s. %v\n", deployment.Name, err)
+		os.Exit(1)
+	}
+
+	deploymentPodName, err := podRunningWithLabel(context, clusterNamespace, deployment.Spec)
```
travisn, line 65: Does this retry until it's running? Seems like it will take at least a few seconds to succeed.

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,162 @@
...
+		fmt.Printf("Error creating deployment %s. %v\n", debugDeploymentSpec, err)
+		os.Exit(1)
+	}
+	fmt.Printf("ensure the debug deployment %s is scaled up\n", deploymentName)
+
+	if err := setDeploymentScale(context, clusterNamespace, debugDeployment.Name, 1); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
```
travisn, line 104: How about logging at the end that it was successful? (perhaps the bash implementation does this?)

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,162 @@
...
+		},
+	}
+	_, err := context.Clientset.AppsV1().Deployments(clusterNamespace).UpdateScale(ctx.TODO(), deploymentName, scale, v1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to update scale of deployment %s. %v\n", deploymentName, err)
+	}
+	return nil
+}
+
+func verifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
```
travisn, line 124: I'd suggest no need for helper methods like this were it's such a simple wrapper around a simple method. How about just calling `Deployments.Get()` where it's needed instead of using the helper? Same suggestion for `updateDeployment()` below

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,51 @@
...
+
+func StopDebug(context *k8sutil.Context, clusterNamespace, deploymentName string) {
+
+	if !strings.Contains(deploymentName, "-debug") {
+		deploymentName = deploymentName + "-debug"
+	}
+
+	debugDeployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
+	if err != nil {
+		fmt.Printf("Missing mon or osd debug deployment name %s. %v\n", deploymentName, err)
```
travisn, line 37: need to exit?

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	fmt.Printf("removing debug mode from deployment %s\n", debugDeployment.Name)
+	err = context.Clientset.AppsV1().Deployments(clusterNamespace).Delete(ctx.TODO(), debugDeployment.Name, v1.DeleteOptions{})
+	if err != nil {
+		fmt.Printf("Error deleting deployment %s: %v", debugDeployment.Name, err)
```
travisn, line 43: - A common error during deletion is that it doesn't exist. But let's ignore that error. 
- Need to exit after the failure?

travisn (changes requested):

pkg/debug/start_debug.go
subhamkrai: yes, I tried to have similar logs.

pkg/debug/start_debug.go
subhamkrai: right, moved that to a loop

pkg/debug/start_debug.go
subhamkrai: this how it is in bash implementation

pkg/debug/start_debug.go
subhamkrai: I tried to make in a helper function as, these methods will be used commonly in other places, and in the future, if we implement mon recovery it will be required too.

I can still remove it if you suggest.

pkg/debug/stop_debug.go
subhamkrai: yes

pkg/debug/stop_debug.go
subhamkrai: done

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+		deployment.Spec.Template.Spec.Containers[0].Image = alternateImageValue
+	}
+
+	labels := deployment.Labels
+	labels["ceph.rook.io/do-not-reconcile"] = "true"
+
+	deployment.Spec.Template.Spec.Containers[0].LivenessProbe = nil
+	deployment.Spec.Template.Spec.Containers[0].StartupProbe = nil
+
+	if alternateImageValue != "" {
```
travisn, line 51: Is this the same as lines 41-43 above?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) {
```
travisn, line 34: It would be nice to exit in a single place, for example with this simple wrapper method. Then the wrapped method just returns the errors. Perhaps even better if the wrapper method were in the `cmd` package, and methods under `pkg` return the errors.
```suggestion
func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) {
{
  err := startDebug()
  if err != nil {
    // log the error
    // exit
  }
}
func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) error 
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+
+func updateDeployment(context *k8sutil.Context, clusterNamespace string, deploymentName *appsv1.Deployment) error {
+	_, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Update(ctx.TODO(), deploymentName, v1.UpdateOptions{})
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func podRunningWithLabel(context *k8sutil.Context, clusterNamespace string, deploymentSpec appsv1.DeploymentSpec) (string, error) {
```
travisn, line 141: If we're waiting, the method name should indicate
```suggestion
func waitForPodToRun(context *k8sutil.Context, clusterNamespace string, deploymentSpec appsv1.DeploymentSpec) (string, error) {
```

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func StopDebug(context *k8sutil.Context, clusterNamespace, deploymentName string) {
```
travisn, line 30: A similar wrapper would be good to exit in a single place.

travisn (changes requested):

pkg/debug/start_debug.go
subhamkrai: right, removed

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
travisn: Now we can return the err instead of printing it
```suggestion
		return err
```

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := setDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
+		return err
+	}
+
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deploymentPodName)
+
+	err = waitForPodDeletion(context, clusterNamespace, deploymentPodName)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
```
travisn, line 84:
```suggestion
		return err
```

travisn (changes requested):

pkg/debug/start_debug.go
subhamkrai: done

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+		return fmt.Errorf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
+	}
+
+	if alternateImageValue != "" {
+		log.Printf("setting debug image to %s\n", alternateImageValue)
+		deployment.Spec.Template.Spec.Containers[0].Image = alternateImageValue
+	}
+
+	labels := deployment.Labels
+	labels["ceph.rook.io/do-not-reconcile"] = "true"
```
microyahoo, line 55: This label has already defined in `pkg/apis/ceph.rook.io/v1/labels.go`

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+
+	deployment.Spec.Template.Spec.Containers[0].Command = []string{"sleep", "infinity"}
+	deployment.Spec.Template.Spec.Containers[0].Args = []string{}
+
+	if err := updateDeployment(context, clusterNamespace, deployment); err != nil {
+		return fmt.Errorf("Failed to update deployment %s. %v\n", deployment.Name, err)
+	}
+
+	deploymentPodName, err := waitForPodToRun(context, clusterNamespace, deployment.Spec)
+	if deploymentPodName == "" || err != nil {
```
microyahoo, line 70: Is there a situation where `deploymentPodName == ""` and `err == nil`? If not, is it more reasonable to change it to `if err != nil {`?

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,61 @@
...
+	err := stopDebug(context, clusterNamespace, deploymentName)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func stopDebug(context *k8sutil.Context, clusterNamespace, deploymentName string) error {
+
+	if !strings.Contains(deploymentName, "-debug") {
```
microyahoo, line 41: Is it more reasonable to change it to `strings.HasSuffix`?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func waitForPodDeletion(context *k8sutil.Context, clusterNamespace, podName string) error {
+	for i := 0; i < 60; i++ {
+		_, err := context.Clientset.CoreV1().Pods(clusterNamespace).Get(ctx.TODO(), podName, v1.GetOptions{})
+		if kerrors.IsNotFound(err) {
+			return nil
+		}
+
+		fmt.Println("waiting for pod to be deleted")
```
travisn, line 164:
```suggestion
		fmt.Println("waiting for pod %q to be deleted", podName)
```

pkg/debug/start_debug.go
subhamkrai: yes, but we don't want to pull all the dependencies rook pull.

pkg/debug/stop_debug.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	debugDeployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
+	if err != nil {
+		return fmt.Errorf("Missing mon or osd debug deployment name %s. %v\n", deploymentName, err)
+	}
+
+	fmt.Printf("removing debug mode from deployment %s\n", debugDeployment.Name)
+	err = context.Clientset.AppsV1().Deployments(clusterNamespace).Delete(ctx.TODO(), debugDeployment.Name, v1.DeleteOptions{})
+	if err != nil || kerrors.IsNotFound(err) {
```
zhucan, line 52: if err != nil && !kerrors.IsNotFound(err) ?

pkg/debug/start_debug.go
```diff
@@ -0,0 +1,169 @@
...
+package debug
+
+import (
+	ctx "context"
+	"fmt"
+	"log"
+	"os"
+	"time"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
```
zhucan, line 26: format import.

pkg/debug/stop_debug.go
subhamkrai: actually, it will be `kerrors.IsNotFound(err)` since we do not want to return error if it is not found since we are deleting only

zhucan: if we don not want to return error, only the err is not nil, return err; no need to check whether the err is "NotFound".

subhamkrai: lets say deletion is complete and we try Delete it will say notFound so to cover that situation I have used that

travisn: I think @zhucan was correct with this suggestion. We want to return an error in this condition. If it's a notFound error, this error will not be returned.
```suggestion
	if err != nil && !kerrors.IsNotFound(err) {
```

subhamkrai: done

travisn (approved):

PR #88, opened 2023-03-15
core: add restore mon quorum in go

subhamkrai:
core: add restore mon quorum in go

this commits add a go command for
restore the mon quorum from single
good mon. In the background we,
scale down the operator first, and run some
ceph commands like remove bad mans from
mon map, extracting/injecting monmap.
And, in the end, we scale up the operator which
bring back the mon quorum.

### DIFF
--- a/cmd/commands/mons.go
+++ b/cmd/commands/mons.go
@@ -29,10 +29,27 @@ var MonCmd = &cobra.Command{
 	Use:                "mons",
 	Short:              "Output mon endpoints",
 	DisableFlagParsing: true,
-	Run: func(cmd *cobra.Command, args []string) {
+	Args:               cobra.MaximumNArgs(1),
+	Run: func(_ *cobra.Command, args []string) {
 		if len(args) == 0 {
 			context := GetContext()
 			fmt.Println(mons.GetMonEndpoint(context, CephClusterNamespace))
 		}
 	},
+}
+
+// RestoreQuorum represents the mons command
+var RestoreQuorum = &cobra.Command{
+	Use:                "restore-quorum",
+	Short:              "When quorum is lost, restore quorum to the remaining healthy mon",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Run: func(_ *cobra.Command, args []string) {
+		context := GetContext()
+		mons.RestoreQuorum(context, OperatorNamespace, CephClusterNamespace, args[0])
+	},
+}
+
+func init() {
+	MonCmd.AddCommand(RestoreQuorum)
 }
--- a/kubectl-rook-ceph.sh
+++ b/kubectl-rook-ceph.sh
@@ -230,7 +230,7 @@ function path_cm_rook_ceph_operator_config() {
 # 'kubectl rook-ceph mons' commands
 ####################################################################################################
 
-function run_mons_command () {
+function run_mons_command() {
   if [ "$#" -ge 1 ] && [ "$1" = "restore-quorum" ]; then
     shift # remove the subcommand from the front of the arg list
     run_restore_quorum "$@"
@@ -253,16 +253,16 @@ function wait_for_deployment_to_be_running() {
 function run_restore_quorum() {
   parse_flags parse_image_flag "$@" # parse flags before the good mon name
   [[ -z "${REMAINING_ARGS[0]:-""}" ]] && fail_error "Missing healthy mon name"
-  good_mon="${REMAINING_ARGS[0]}"              # get the good mon being used to restore quorum
-  shift # remove the healthy mon from the front of the arg list
-  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}")           # remove mon name from remaining args
-  end_of_command_parsing "$@" # end of command tree
+  good_mon="${REMAINING_ARGS[0]}"           # get the good mon being used to restore quorum
+  shift                                     # remove the healthy mon from the front of the arg list
+  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}") # remove mon name from remaining args
+  end_of_command_parsing "$@"               # end of command tree
 
   # Parse the endpoints configmap for the mon endpoints
   bad_mons=()
   mon_endpoints=$(KUBECTL_NS_CLUSTER get cm rook-ceph-mon-endpoints -o jsonpath='{.data.data}')
   # split the endpoints into an array, separated by the comma
-  for single_mon in ${mon_endpoints//,/ } ; do
+  for single_mon in ${mon_endpoints//,/ }; do
     mon_name=$(echo "${single_mon/=/ }" | awk '{print $1}')
     mon_endpoint=$(echo "${single_mon/=/ }" | awk '{print $2}')
     echo "mon=$mon_name, endpoint=$mon_endpoint"
@@ -335,12 +335,11 @@ function run_restore_quorum() {
     --public-bind-addr=$ROOK_POD_IP \
     --extract-monmap=$monmap_path
 
-  info_msg "Printing monmap"; \
+  info_msg "Printing monmap"
   KUBECTL_NS_CLUSTER exec deploy/rook-ceph-mon-$good_mon-debug -c mon -- monmaptool --print $monmap_path
 
   # remove all the mons except the good one
-  for bad_mon in "${bad_mons[@]}"
-  do
+  for bad_mon in "${bad_mons[@]}"; do
     info_msg "Removing mon $bad_mon"
     KUBECTL_NS_CLUSTER exec deploy/rook-ceph-mon-$good_mon-debug -c mon -- monmaptool $monmap_path --rm $bad_mon
   done
@@ -381,8 +380,7 @@ function run_restore_quorum() {
   info_msg "Purging the bad mons: ${bad_mons[*]}"
   # ignore errors purging old mons if their resources don't exist
   set +e
-  for bad_mon in "${bad_mons[@]}"
-  do
+  for bad_mon in "${bad_mons[@]}"; do
     info_msg "purging old mon: $bad_mon"
     KUBECTL_NS_CLUSTER delete deploy rook-ceph-mon-$bad_mon
     KUBECTL_NS_CLUSTER delete svc rook-ceph-mon-$bad_mon
@@ -433,8 +431,7 @@ function wait_for_mon_status_response() {
   sleep_time=5
 
   exit_status=1
-  while [[ $exit_status != 0 ]]
-  do
+  while [[ $exit_status != 0 ]]; do
     # Don't fail the script if the ceph command fails
     set +e
     KUBECTL_NS_CLUSTER exec deploy/rook-ceph-tools -- ceph status --connect-timeout=3
@@ -642,8 +639,8 @@ function run_start_debug() {
   # 3) debug start deploymentName
   parse_flags parse_image_flag "$@" # parse flags before the deployment name
   [[ -z "${REMAINING_ARGS[0]:-""}" ]] && fail_error "Missing mon or osd deployment name"
-  deployment_name="${REMAINING_ARGS[0]}"              # get deployment name
-  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}")           # remove deploy name from remaining args
+  deployment_name="${REMAINING_ARGS[0]}"    # get deployment name
+  REMAINING_ARGS=("${REMAINING_ARGS[@]:1}") # remove deploy name from remaining args
   set +u
   parse_flags parse_image_flag "${REMAINING_ARGS[@]}" # parse flags after the deployment name
   set -u
@@ -694,8 +691,8 @@ function run_start_debug() {
     spec:
         $deployment_spec
 EOF
-    info_msg "ensure the debug deployment $deployment_name is scaled up"
-    KUBECTL_NS_CLUSTER scale deployments "$deployment_name-debug" --replicas=1
+  info_msg "ensure the debug deployment $deployment_name is scaled up"
+  KUBECTL_NS_CLUSTER scale deployments "$deployment_name-debug" --replicas=1
 }
 
 function run_stop_debug() {
--- a/pkg/debug/start_debug.go
+++ b/pkg/debug/start_debug.go
@@ -26,13 +26,11 @@ import (
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
-	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) {
-
 	err := startDebug(context, clusterNamespace, deploymentName, alternateImageValue)
 	if err != nil {
 		fmt.Println(err)
@@ -41,10 +39,13 @@ func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
 }
 
 func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) error {
-	deployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
+	originalDeployment, err := GetDeployment(context, clusterNamespace, deploymentName)
 	if err != nil {
 		return fmt.Errorf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
 	}
+
+	// We need to dereference the deployment as it is required for the debug deployment
+	deployment := *originalDeployment
 
 	if alternateImageValue != "" {
 		log.Printf("setting debug image to %s\n", alternateImageValue)
@@ -62,23 +63,21 @@ func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
 	deployment.Spec.Template.Spec.Containers[0].Command = []string{"sleep", "infinity"}
 	deployment.Spec.Template.Spec.Containers[0].Args = []string{}
 
-	if err := updateDeployment(context, clusterNamespace, deployment); err != nil {
-		return fmt.Errorf("Failed to update deployment %s. %v\n", deployment.Name, err)
-	}
-
-	deploymentPodName, err := waitForPodToRun(context, clusterNamespace, deployment.Spec)
+	labelSelector := fmt.Sprintf("ceph_daemon_type=%s,ceph_daemon_id=%s", deployment.Spec.Template.Labels["ceph_daemon_type"], deployment.Spec.Template.Labels["ceph_daemon_id"])
+	deploymentPodName, err := k8sutil.WaitForPodToRun(context, clusterNamespace, labelSelector)
 	if err != nil {
 		fmt.Println(err)
 		return err
 	}
 
-	if err := setDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
+	if err := SetDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
 		return err
 	}
+	fmt.Printf("deployment %s scaled down\n", deployment.Name)
 
-	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deploymentPodName)
+	fmt.Printf("waiting for the deployment pod %s to be deleted\n", deploymentPodName.Name)
 
-	err = waitForPodDeletion(context, clusterNamespace, deploymentPodName)
+	err = waitForPodDeletion(context, clusterNamespace, deploymentName)
 	if err != nil {
 		fmt.Println(err)
 		return err
@@ -99,13 +98,14 @@ func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
 	}
 	fmt.Printf("ensure the debug deployment %s is scaled up\n", deploymentName)
 
-	if err := setDeploymentScale(context, clusterNamespace, debugDeployment.Name, 1); err != nil {
+	if err := SetDeploymentScale(context, clusterNamespace, debugDeployment.Name, 1); err != nil {
 		return err
 	}
+
 	return nil
 }
 
-func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentName string, scaleCount int) error {
+func SetDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentName string, scaleCount int) error {
 	scale := &autoscalingv1.Scale{
 		ObjectMeta: v1.ObjectMeta{
 			Name:      deploymentName,
@@ -122,11 +122,14 @@ func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentNa
 	return nil
 }
 
-func verifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
+func GetDeployment(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
+	fmt.Printf("fetching the deployment %s to be running\n", deploymentName)
 	deployment, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Get(ctx.TODO(), deploymentName, v1.GetOptions{})
 	if err != nil {
+		fmt.Printf("deployment %s doesn't exist. %v", deploymentName, err)
 		return nil, err
 	}
+	fmt.Printf("deployment %s exists\n", deploymentName)
 	return deployment, nil
 }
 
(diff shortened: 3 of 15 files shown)


pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,252 @@
...
+		log.Fatalf("failed to get mon configmap %s %v", MonConfigMap, err)
+	}
+
+	monData := monCm.Data["data"]
+	monEndpoints := strings.Split(monData, ",")
+
+	var goodMonPublicIp, goodMonPort string
+	var badMons []string
+	for _, m := range monEndpoints {
+		monName, monEndpoint, _ := strings.Cut(m, "=")
```
travisn, line 54: If there is no `=`, seems like we need to handle it like this:
```suggestion
		monName, monEndpoint, ok := strings.Cut(m, "=")
		if !ok {
		  // handle the bad mon
		}
```

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,252 @@
...
+		if monName == goodMon {
+			goodMonPublicIp, goodMonPort, _ = strings.Cut(monEndpoint, ":")
```
travisn, line 56: Handle if there is no `:`
```suggestion
			goodMonPublicIp, goodMonPort, ok = strings.Cut(monEndpoint, ":")
```

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,252 @@
...
+		} else {
+			badMons = append(badMons, monName)
+		}
+	}
+
+	if goodMonPublicIp == "" {
+		return fmt.Errorf("error: good mon %s not found", goodMon)
+	}
+
+	cephFsidInBytes, err := base64.StdEncoding.DecodeString(monCm.Data["fsid"])
```
travisn, line 66: The fsid is encoded in the configmap? Are you sure the fsid is in this configmap? It might be in the `rook-ceph-mon` secret instead.

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,252 @@
...
+
+func RestoreQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace, goodMon string) {
+	err := restoreQuorum(context, operatorNamespace, clusterNamespace, goodMon)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func restoreQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace, goodMon string) error {
```
travisn, line 42: Instead of one long method that does what the same function did in the bash script, let's factor this out into helper methods where it makes sense. Then we can also unit test the pieces.

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,252 @@
...
+		log.Fatalf("failed to update mon configmap %s %v", MonConfigMap, err)
+	}
+
+	fmt.Printf("Stopping the debug pod for mon %s.\n", goodMon)
+	debug.StopDebug(context, clusterNamespace, fmt.Sprintf("rook-ceph-mon-%s", goodMon))
+
+	fmt.Println("Check that the restored mon is responding")
+	maxRetries := 1
+
+	for maxRetries == 20 {
```
travisn, line 190: Wherever we have a for loop, likely should be factored out to a helper. This might also make sense as a helper in the exec package to wait for a ceph command to succeed.

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,252 @@
...
+
+		output := exec.RunCommandInToolboxPod(context, "ceph", []string{"status", "--connect-timeout=3"}, operatorNamespace, clusterNamespace)
+		if output != "" && (strings.Contains(output, "HEALTH_WARN") || strings.Contains(output, "HEALTH_OK") || strings.Contains(output, "HEALTH_ERROR")) {
+			fmt.Println("finished waiting for ceph status")
+			maxRetries = 20
```
travisn, line 195: define a const for the max retries

travisn (changes requested):

pkg/exec/exec.go
```diff
@@ -60,12 +61,39 @@ func RunShellCommandInOperatorPod(ctx *k8sutil.Context, arg []string, operatorNa
...
+		os.Exit(1)
+	}
+
+	output := new(bytes.Buffer)
+
+	execCmdInPod(ctx, cmd, pod.Name, "rook-ceph-toolbox", pod.Namespace, clusterNamespace, args, output)
+	return output.String()
+}
+
+func RunCommandInMonPod(ctx *k8sutil.Context, label, cmd string, args []string, clusterNamespace string) string {
```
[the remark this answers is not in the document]
subhamkrai: done

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,255 @@
...
+		"--log-stderr-prefix=debug",
+		"--default-log-to-file=false",
+		"--default-mon-cluster-log-to-file=false",
+		"--mon-host=$(ROOK_CEPH_MON_HOST)",
+		"--mon-initial-members=$(ROOK_CEPH_MON_INITIAL_MEMBERS)",
+		fmt.Sprintf("--id=%s", goodMon),
+		"--foreground",
+		fmt.Sprintf("--public-addr=%s", goodMonPublicIp),
+		fmt.Sprintf("--setuser-match-path=/var/lib/ceph/mon/ceph-%s/store.db", goodMon),
+		"--public-bind-addr=$(ROOK_POD_IP)",
```
subhamkrai, line 141: I'm getting here ```Extracting the monmap
parse error setting 'public_bind_addr' to '$(ROOK_POD_IP)'```
@travisn any idea?

subhamkrai: https://github.com/rook/kubectl-rook-ceph/actions/runs/4465810745/jobs/7843239891?pr=88#step:6:94

travisn: I see in the test for the restore-quorum function in the script, this is the output:
```
parse error setting 'public_bind_addr' to ''
```

When resetting the monmap the bind address wouldn't matter, so try just setting it to blank.

pkg/mons/restore_quorum.go
subhamkrai: got it

pkg/mons/restore_quorum.go

subhamkrai:
found one issue working on the fix.

travisn (commented):
Testing locally, I saw that the restored mon pod has the `sleep infinity` command instead of having the original `ceph-mon` command restored. I wouldn't expect the command to be modified on the mon. It should have just been modified on the mon debug pod. If you fix this command when starting back up the mon, hopefully the quorum will start again.

subhamkrai:
> Testing locally, I saw that the restored mon pod has the `sleep infinity` command instead of having the original `ceph-mon` command restored. I wouldn't expect the command to be modified on the mon. It should have just been modified on the mon debug pod. If you fix this command when starting back up the mon, hopefully the quorum will start again.

thanks @travisn I,ll check and update.

pkg/debug/start_debug.go
```diff
@@ -39,11 +39,13 @@ func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
 }
 
 func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) error {
-	deployment, err := VerifyDeploymentExists(context, clusterNamespace, deploymentName)
+	OriginalDeployment, err := VerifyDeploymentExists(context, clusterNamespace, deploymentName)
 	if err != nil {
 		return fmt.Errorf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
 	}
 
+	deployment := *OriginalDeployment
```
subhamkrai, line 47: after these changes, I tested locally everything was working as expected.

cmd/commands/mons.go
```diff
@@ -29,10 +29,28 @@ var MonCmd = &cobra.Command{
...
+// RestoreQuorum represents the mons command
+var RestoreQuorum = &cobra.Command{
+	Use:                "restore-quorum",
+	Short:              "When quorum is lost, restore quorum to the remaining healthy mon",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+	Run: func(_ *cobra.Command, args []string) {
+		context := GetContext()
+		mons.RestoreQuorum(context, OperatorNamespace, CephClusterNamespace, args[0])
+
```
travisn, line 50: nit: remove blank line before closing brace

pkg/debug/start_debug.go
```diff
@@ -41,11 +39,13 @@ func StartDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
 }
 
 func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alternateImageValue string) error {
-	deployment, err := verifyDeploymentExists(context, clusterNamespace, deploymentName)
+	OriginalDeployment, err := VerifyDeploymentExists(context, clusterNamespace, deploymentName)
```
travisn, line 42: nit: local vars would be lower case
```suggestion
	originalDeployment, err := VerifyDeploymentExists(context, clusterNamespace, deploymentName)
```

pkg/debug/start_debug.go
travisn: How about adding a comment that it is needed to dereference the var

pkg/debug/start_debug.go
```diff
@@ -62,23 +62,21 @@ func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
...
 	if err != nil {
 		fmt.Println(err)
 		return err
 	}
 
-	if err := setDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
+	if err := SetDeploymentScale(context, clusterNamespace, deployment.Name, 0); err != nil {
 		return err
 	}
+	fmt.Printf("deployment.apps/%s scaled\n", deployment.Name)
```
travisn, line 75:
```suggestion
	fmt.Printf("deployment %s scaled down\n", deployment.Name)
```

pkg/debug/start_debug.go
```diff
@@ -122,11 +121,13 @@ func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentNa
 	return nil
 }
 
-func verifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
+func VerifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
+	fmt.Printf("waiting for deployment %s to be running\n", deploymentName)
```
travisn, line 125: It doesn't look like we're waiting for the deployment to be running? We're just retrieving it

pkg/debug/start_debug.go
```diff
@@ -122,11 +121,13 @@ func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentNa
...
 	deployment, err := context.Clientset.AppsV1().Deployments(clusterNamespace).Get(ctx.TODO(), deploymentName, v1.GetOptions{})
 	if err != nil {
 		return nil, err
```
travisn, line 128: Need to log a message that it doesn't exist?

pkg/debug/start_debug.go
```diff
@@ -122,11 +121,13 @@ func setDeploymentScale(context *k8sutil.Context, clusterNamespace, deploymentNa
 	return nil
 }
 
-func verifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
+func VerifyDeploymentExists(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
```
travisn, line 124: Are we just getting the deployment?
```suggestion
func GetDeployment(context *k8sutil.Context, clusterNamespace, deploymentName string) (*appsv1.Deployment, error) {
```

pkg/exec/exec.go
```diff
@@ -37,41 +38,63 @@ var (
...
 	}
 
-	output := new(bytes.Buffer)
+	var stdout, stderr bytes.Buffer
 
-	ExecCmdInPod(ctx, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, output)
-	return output.String()
+	err = execCmdInPod(ctx, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
+	if err != nil {
+		log.Error(err)
```
travisn, line 50: need to exit after logging the error?

pkg/exec/exec.go
```diff
@@ -37,41 +38,63 @@ var (
 
 func RunCommandInOperatorPod(ctx *k8sutil.Context, cmd string, args []string, operatorNamespace, clusterNamespace string) string {
 
-	pod, err := k8sutil.WaitForOperatorPod(ctx, operatorNamespace)
+	pod, err := k8sutil.WaitForPodToRun(ctx, operatorNamespace, "app=rook-ceph-operator")
 	if err != nil {
 		os.Exit(1)
```
travisn, line 43: need to log an error before exiting?

pkg/exec/exec.go
```diff
@@ -37,41 +38,63 @@ var (
...
-	args = append(args, arg...)
+	var stdout, stderr bytes.Buffer
 
-	output := new(bytes.Buffer)
-
-	ExecCmdInPod(ctx, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, output)
-	return output.String()
+	err = execCmdInPod(ctx, cmd, pod.Name, "rook-ceph-tools", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
+	if err != nil {
+		log.Error(err)
```
travisn, line 66: exit?

pkg/exec/exec.go
```diff
@@ -37,41 +38,63 @@ var (
...
+	list, err := ctx.Clientset.CoreV1().Pods(clusterNamespace).List(context.TODO(), opts)
+	if err != nil || len(list.Items) == 0 {
+		log.Error("failed to get rook mon pod where the command could be executed")
+		log.Fatal(err)
+	}
+	var stdout, stderr bytes.Buffer
+
+	err = execCmdInPod(ctx, cmd, list.Items[0].Name, container, list.Items[0].Namespace, clusterNamespace, args, &stdout, &stderr)
+	if err != nil {
+		log.Error(err)
```
travisn, line 83: exit?

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,298 @@
...
+		} else if monName == goodMon {
+			goodMonPublicIp, goodMonPort, ok = strings.Cut(monEndpoint, ":")
+			if !ok {
+				return fmt.Errorf("failed to get good mon endpoint and port")
+			}
+		} else {
+			badMons = append(badMons, monName)
+		}
+		fmt.Printf("mon=%s, endpoints=%s\n", monName, monEndpoint)
+
```
travisn, line 67: nit: remove blank line before closing brace

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,298 @@
...
+		fmt.Printf("%d: waiting for ceph status to confirm single mon quorum. \n", i+1)
+		fmt.Printf("current ceph status output %s\n", output)
+		fmt.Println("sleeping for 5 seconds")
+		time.Sleep(5 * time.Second)
+	}
+
+	return nil
+}
+
+func promtToContinueOrCancel(answer string) (string, error) {
```
travisn, line 284:
```suggestion
func promptToContinueOrCancel(answer string) (string, error) {
```

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,298 @@
...
+	var answer, output string
+	fmt.Printf("Are you sure you want to restore the quorum to mon %s? yes-really-restore\n", goodMon)
+	fmt.Scanf("%s", &answer)
+	output, err = promtToContinueOrCancel(answer)
+	if err != nil {
+		return fmt.Errorf(" restoring the mon quorum to mon %s cancelled", goodMon)
+	}
+	fmt.Println(output)
+
+	fmt.Println("Waiting for operator pods to stop")
```
travisn, line 110:
```suggestion
	fmt.Println("Waiting for operator pod to stop")
```

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,298 @@
...
+	err = debug.SetDeploymentScale(context, operatorNamespace, "rook-ceph-operator", 0)
+	if err != nil {
+		return fmt.Errorf("failed to stop deployment rook-ceph-operator. %v", err)
+	}
+	fmt.Println("deployment.apps/rook-ceph-operator scaled")
```
travisn, line 115:
```suggestion
	fmt.Println("rook-ceph-operator deployment scaled down")
```

pkg/mons/restore_quorum.go
travisn: How about factoring it out to more helper methods? It is still really long.

travisn (changes requested):

pkg/debug/start_debug.go
subhamkrai: I tried to keep logs similar to what we have currently. you want to update?

pkg/mons/restore_quorum.go
subhamkrai: this is same log that we get today, you want to update?

pkg/debug/start_debug.go
subhamkrai: added

pkg/debug/start_debug.go
subhamkrai: correct, update the name

pkg/debug/start_debug.go
subhamkrai: added

pkg/exec/exec.go
subhamkrai: actually, no because we want to retry running the command instead returning first time.

pkg/exec/exec.go
subhamkrai: actually, no because we want to retry running the command instead returning first time.

pkg/exec/exec.go
subhamkrai: actually, no because we want to retry running the command instead returning first time.

pkg/mons/restore_quorum.go
travisn: yes let's update this one too

subhamkrai: added

pkg/exec/exec.go
subhamkrai: done

pkg/exec/exec.go
subhamkrai: added

pkg/debug/start_debug.go
subhamkrai: right

pkg/mons/restore_quorum.go
subhamkrai: done

pkg/exec/exec.go
```diff
@@ -35,43 +36,77 @@ var (
 	CephClusterNamespace string // Cephcluster namespace
 )
 
-func RunCommandInOperatorPod(ctx *k8sutil.Context, cmd string, args []string, operatorNamespace, clusterNamespace string) string {
+func RunCommandInOperatorPod(ctx *k8sutil.Context, cmd string, args []string, operatorNamespace, clusterNamespace string, wantToExit bool) string {
```
travisn, line 39:
```suggestion
func RunCommandInOperatorPod(ctx *k8sutil.Context, cmd string, args []string, operatorNamespace, clusterNamespace string, exitOnError bool) string {
```

kubectl-rook-ceph.sh
```diff
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -eEuo pipefail
+set -exEuo pipefail
```
travisn, line 17: Why is the `-x` needed since this is a golang PR?

pkg/exec/exec.go
```diff
@@ -35,43 +36,77 @@ var (
...
+			os.Exit(1)
+		}
+	}
+	fmt.Println(stderr.String())
+	return stdout.String()
 }
 
-func RunShellCommandInOperatorPod(ctx *k8sutil.Context, arg []string, operatorNamespace, clusterNamespace string) string {
-	pod, err := k8sutil.WaitForOperatorPod(ctx, operatorNamespace)
+func RunCommandInToolboxPod(ctx *k8sutil.Context, cmd string, args []string, clusterNamespace string, wantToExit bool) string {
```
travisn, line 60:
```suggestion
func RunCommandInToolboxPod(ctx *k8sutil.Context, cmd string, args []string, clusterNamespace string, exitOnError bool) string {
```

pkg/exec/exec.go
```diff
@@ -35,43 +36,77 @@ var (
...
+			os.Exit(1)
+		}
+	}
+	fmt.Println(stderr.String())
+	return stdout.String()
 }
 
-// ExecCmdInPod exec command on specific pod and wait the command's output.
-func ExecCmdInPod(ctx *k8sutil.Context, command, podName, containerName, podNamespace, clusterNamespace string, args []string, stdout io.Writer) {
+func RunCommandInLabeledPod(ctx *k8sutil.Context, label, container, cmd string, args []string, clusterNamespace string, wantToExit bool) string {
```
travisn, line 80:
```suggestion
func RunCommandInLabeledPod(ctx *k8sutil.Context, label, container, cmd string, args []string, clusterNamespace string, exitOnError bool) string {
```

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,291 @@
...
+		return fmt.Errorf("failed to deployment rook-ceph-mon-%s-debug", goodMon)
+	}
+
+	labelSelector := fmt.Sprintf("ceph_daemon_type=%s,ceph_daemon_id=%s", debugDeploymentSpec.Spec.Template.Labels["ceph_daemon_type"], debugDeploymentSpec.Spec.Template.Labels["ceph_daemon_id"])
+	_, err = k8sutil.WaitForPodToRun(context, clusterNamespace, labelSelector)
+	if err != nil {
+		return fmt.Errorf("failed to start deployment %s", fmt.Sprintf("rook-ceph-mon-%s-debug", goodMon))
+	}
+
+	fmt.Println("Started debug pod, restoring the mon quorum in the debug pod")
```
travisn, line 126: How about factoring out lines 126-172 into a helper method for updating the monmap? That will help shorten this long method.

travisn (changes requested):
A few more small changes...

pkg/mons/restore_quorum.go
subhamkrai: refactored!

kubectl-rook-ceph.sh
subhamkrai: it was there for testing, removed now.

kubectl-rook-ceph.sh
```diff
@@ -230,7 +230,7 @@ function path_cm_rook_ceph_operator_config() {
 # 'kubectl rook-ceph mons' commands
 ####################################################################################################
 
-function run_mons_command () {
+function run_mons_command() {
```
travisn, line 233: The changes in this script are just all from the automatic editor, right?

pkg/mons/restore_quorum.go
```diff
@@ -0,0 +1,295 @@
...
+
+	err = debug.SetDeploymentScale(context, clusterNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		return fmt.Errorf("failed to start deployment rook-ceph-operator. %v", err)
+	}
+
+	return nil
+}
+
+func UpdateMonMap(context *k8sutil.Context, clusterNamespace, labelSelector, cephFsid, goodMon, goodMonPublicIp string, badMons []string) {
```
travisn, line 167: use lower case methods whenever the method is not needed outside the package
```suggestion
func updateMonMap(context *k8sutil.Context, clusterNamespace, labelSelector, cephFsid, goodMon, goodMonPublicIp string, badMons []string) {
```

travisn: If this function fails, it will just exit, right?

travisn (changes requested):
The commits look like they aren't strictly needed. How about squashing them? Then a separate commit for the log collector likely makes sense.

pkg/mons/restore_quorum.go
subhamkrai: yes, since we are passing `true` in exitOnError

kubectl-rook-ceph.sh
subhamkrai: yes.

travisn (changes requested):
LGTM. Can you just update the commit description? It should discuss the command to restore mon quorum, instead of just talking about other small changes.

travisn (approved):

PR #90, opened 2023-04-26
core: add health command in go

subhamkrai:
core: add health command in go

These changes are on top of the mon quorum pr, so check the last commit only. Thanks!

command
`kubectl rook-ceph health`

output
```
Info:  Checking if at least three mon pods are running on different nodes

Warning: At least three mon pods should running on different nodesrook-ceph-mon-a-655fbb547-7fbbn        Running         rook-ceph       minikube
rook-ceph-mon-b-6f74686789-bl5lj         Running         rook-ceph       minikube
rook-ceph-mon-c-7676c5485-bbg2v          Running         rook-ceph       minikube

Info:  Checking mon quorum and ceph health details


Info:  Checking if at least three osd pods are running on different nodes

Warning: At least three osd pods should running on different nodesrook-ceph-osd-0-599db9876f-qzv7f       Running         rook-ceph       minikube
rook-ceph-osd-1-77959f9db5-j2nw9         Running         rook-ceph       minikube

Info: Pods that are in 'Running' status
csi-cephfsplugin-provisioner-84cc595b78-m6xvt    Running         rook-ceph       minikube
csi-cephfsplugin-sgrx5   Running         rook-ceph       minikube
csi-rbdplugin-khg6h      Running         rook-ceph       minikube
csi-rbdplugin-provisioner-6f6b6b8cd6-s6jss       Running         rook-ceph       minikube
rook-ceph-mgr-a-5dd87bf686-xnkwp         Running         rook-ceph       minikube
rook-ceph-mon-a-655fbb547-7fbbn          Running         rook-ceph       minikube
rook-ceph-mon-b-6f74686789-bl5lj         Running         rook-ceph       minikube
rook-ceph-mon-c-7676c5485-bbg2v          Running         rook-ceph       minikube
rook-ceph-operator-cf4f7dfd4-4845g       Running         rook-ceph       minikube
rook-ceph-osd-0-599db9876f-qzv7f         Running         rook-ceph       minikube
rook-ceph-osd-1-77959f9db5-j2nw9         Running         rook-ceph       minikube
csi-cephfsplugin-provisioner-84cc595b78-m6xvt    Running         rook-ceph       minikube
csi-cephfsplugin-sgrx5   Running         rook-ceph       minikube
csi-rbdplugin-khg6h      Running         rook-ceph       minikube
csi-rbdplugin-provisioner-6f6b6b8cd6-s6jss       Running         rook-ceph       minikube
rook-ceph-mgr-a-5dd87bf686-xnkwp         Running         rook-ceph       minikube
rook-ceph-mon-a-655fbb547-7fbbn          Running         rook-ceph       minikube
rook-ceph-mon-b-6f74686789-bl5lj         Running         rook-ceph       minikube
rook-ceph-mon-c-7676c5485-bbg2v          Running         rook-ceph       minikube
rook-ceph-operator-cf4f7dfd4-4845g       Running         rook-ceph       minikube
rook-ceph-osd-0-599db9876f-qzv7f         Running         rook-ceph       minikube
rook-ceph-osd-1-77959f9db5-j2nw9         Running         rook-ceph       minikube

Warning: Pods that are 'Not' in 'Running' status
rook-ceph-osd-prepare-minikube-m6dks     Succeeded       rook-ceph       minikube
rook-ceph-osd-prepare-minikube-m6dks     Succeeded       rook-ceph       minikube
Info:  checking placement group status


INFO: 32 pgs: 32 active+clean; 463 KiB data, 40 MiB used, 39 GiB / 39 GiB avail


Info:  checking if at least one mgr pod is running
rook-ceph-mgr-a-5dd87bf686-xnkwp         Running         rook-ceph       minikube
```

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -35,6 +35,8 @@ jobs:
       - name: Test plugin with ceph commands
         run: |
           set -e
+          kubectl rook-ceph health
+
           kubectl rook-ceph ceph status
 
           # test the mon restore to restore to mon a, delete mons b and c, then add d and e
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -1,0 +1,205 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package health
+
+import (
+	ctx "context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"log"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+)
+
+type cephStatus struct {
+	PgMap  pgMap        `json:"pgmap"`
+	Health healthStatus `json:"health"`
+}
+
+type healthStatus struct {
+	Status string `json:"status"`
+}
+
+type pgMap struct {
+	PgsByState []PgStateEntry `json:"pgs_by_state"`
+}
+
+type PgStateEntry struct {
+	StateName string `json:"state_name"`
+	Count     int    `json:"count"`
+}
+
+func Health(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+
+	fmt.Println("Info: Checking if at least three mon pods are running on different nodes")
+	checkPodsOnNodes(context, clusterNamespace, "app=rook-ceph-mon")
+
+	fmt.Println()
+	fmt.Println("Info: Checking mon quorum and ceph health details")
+	checkMonQuorum(context, operatorNamespace, clusterNamespace)
+
+	fmt.Println()
+	fmt.Println("Info: Checking if at least three osd pods are running on different nodes")
+	checkPodsOnNodes(context, clusterNamespace, "app=rook-ceph-osd")
+
+	fmt.Println()
+	CheckAllPodsStatus(context, operatorNamespace, clusterNamespace)
+
+	fmt.Println("Info: Checking placement group status")
+	checkPgStatus(context, operatorNamespace, clusterNamespace)
+
+	fmt.Println()
+	fmt.Println("Info: Checking if at least one mgr pod is running")
+	checkMgrPodsStatusAndCounts(context, clusterNamespace)
+}
+
+func checkPodsOnNodes(context *k8sutil.Context, clusterNamespace, label string) {
+	var daemonType string
+	if strings.Contains(label, "osd") {
+		daemonType = "osd"
+	} else if strings.Contains(label, "mon") {
+		daemonType = "mon"
+	}
+
+	opts := metav1.ListOptions{LabelSelector: label}
+	podList, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), opts)
+	if err != nil {
+		fmt.Printf("\nfailed to list %s pods with label %s: %v\n", daemonType, opts.LabelSelector, err)
+		return
+	}
+
+	var nodeList = make(map[string]string)
+	for i := range podList.Items {
+		nodeName := podList.Items[i].Spec.NodeName
+		if _, okay := nodeList[nodeName]; !okay {
+			nodeList[nodeName] = podList.Items[i].Name
+		}
+	}
+
+	if len(nodeList) < 3 {
+		fmt.Printf("\nWarning: At least three %s pods should running on different nodes\n", daemonType)
+	}
+
+	for i := range podList.Items {
+		fmt.Println(podList.Items[i].Name, "\t", podList.Items[i].Status.Phase, "\t", podList.Items[i].Namespace, "\t", podList.Items[i].Spec.NodeName)
+	}
+}
+
+func checkMonQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	cephHealthStatus, _ := unMarshalCephStatus(context, operatorNamespace, clusterNamespace)
+	if cephHealthStatus == "HEALTH_OK" {
+		fmt.Println("Info:", cephHealthStatus)
+	} else if cephHealthStatus == "HEALTH_WARN" {
+		fmt.Println("Warning:", cephHealthStatus)
+	} else if cephHealthStatus == "HEALTH_ERR" {
+		fmt.Println("Error: ", cephHealthStatus)
+	}
+}
+
+func CheckAllPodsStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	var podNotRunning, podRunning []v1.Pod
+	podRunning, podNotRunning = getPodRunningStatus(context, operatorNamespace)
+	if operatorNamespace != clusterNamespace {
+		clusterRunningPod, clusterNotRunningPod := getPodRunningStatus(context, clusterNamespace)
+		podRunning = append(podRunning, clusterRunningPod...)
+		podNotRunning = append(podNotRunning, clusterNotRunningPod...)
+	}
+
+	if podRunning != nil {
+		fmt.Println("Info: Pods that are in 'Running' status")
+		for i := range podRunning {
+			fmt.Println(podRunning[i].Name, "\t", podRunning[i].Status.Phase, "\t", podRunning[i].Namespace, "\t", podRunning[i].Spec.NodeName)
+		}
+	}
+
+	if podNotRunning != nil {
+		fmt.Println("\nWarning: Pods that are 'Not' in 'Running' status")
+		for i := range podNotRunning {
+			fmt.Println(podNotRunning[i].Name, "\t", podNotRunning[i].Status.Phase, "\t", podNotRunning[i].Namespace, "\t", podNotRunning[i].Spec.NodeName)
+		}
+	}
+}
+
+func getPodRunningStatus(context *k8sutil.Context, namespace string) ([]v1.Pod, []v1.Pod) {
+	var podNotRunning, podRunning []v1.Pod
+	podList, err := context.Clientset.CoreV1().Pods(namespace).List(ctx.TODO(), metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("\nfailed to list pods in namespace %s: %v\n", namespace, err)
+		return []v1.Pod{}, []v1.Pod{}
+	}
+
+	for i := range podList.Items {
+		if podList.Items[i].Status.Phase != v1.PodRunning {
+			podNotRunning = append(podNotRunning, podList.Items[i])
+		} else {
+			podRunning = append(podRunning, podList.Items[i])
+		}
+	}
+	return podRunning, podNotRunning
+}
+
+func checkPgStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	_, pgStateEntryList := unMarshalCephStatus(context, operatorNamespace, clusterNamespace)
+	for _, pgStatus := range pgStateEntryList {
+		if pgStatus.StateName == "active+clean" {
+			fmt.Println("Info:\n", "\tPgState: ", pgStatus.StateName+",", "PgCount: ", pgStatus.Count)
+		} else if strings.Contains(pgStatus.StateName, "down") || strings.Contains(pgStatus.StateName, "incomplete") || strings.Contains(pgStatus.StateName, "snaptrim_error") {
+			fmt.Println("Warning:\n", "\tPgState: ", pgStatus.StateName+",", "PgCount: ", pgStatus.Count)
+		} else {
+			fmt.Println("Error:\n", "\tPgState: ", pgStatus.StateName+",", "PgCount: ", pgStatus.Count)
+		}
+	}
+}
+
+func checkMgrPodsStatusAndCounts(context *k8sutil.Context, clusterNamespace string) {
+	opts := metav1.ListOptions{LabelSelector: "app=rook-ceph-mgr"}
+	podList, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), opts)
+	if err != nil {
+		fmt.Printf("\nfailed to list mgr pods with label %s: %v\n", opts.LabelSelector, err)
+		return
+	}
+
+	if len(podList.Items) < 1 {
+		fmt.Println("At least one mgr pod should be running")
+	}
+	for i := range podList.Items {
+		fmt.Println(podList.Items[i].Name, "\t", podList.Items[i].Status.Phase, "\t", podList.Items[i].Namespace, "\t", podList.Items[i].Spec.NodeName)
+	}
+}
+
+func unMarshalCephStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) (string, []PgStateEntry) {
+	cephStatusOut := exec.RunCommandInOperatorPod(context, "ceph", []string{"-s", "--format", "json"}, operatorNamespace, clusterNamespace, false)
+
+	ecodedText := base64.StdEncoding.EncodeToString([]byte(cephStatusOut))
+	decodeCephStatus, err := base64.StdEncoding.DecodeString(ecodedText)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var cephStatus *cephStatus
+
+	err = json.Unmarshal(decodeCephStatus, &cephStatus)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return cephStatus.Health.Status, cephStatus.PgMap.PgsByState
+}
--- a/cmd/commands/health.go
+++ b/cmd/commands/health.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2023 The Rook Authors. All rights reserved.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
@@ -13,23 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package command
 
-import command "github.com/rook/kubectl-rook-ceph/cmd/commands"
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/health"
+	"github.com/spf13/cobra"
+)
 
-func main() {
-	addcommands()
-	command.RootCmd.Execute()
-}
-
-func addcommands() {
-	command.RootCmd.AddCommand(
-		command.CephCmd,
-		command.MonCmd,
-		command.RbdCmd,
-		command.OperatorCmd,
-		command.RookCmd,
-		command.DebugCmd,
-		command.Health,
-	)
+var Health = &cobra.Command{
+	Use:                "health",
+	Short:              "check health of the cluster and common configuration issues",
+	DisableFlagParsing: true,
+	Args:               cobra.NoArgs,
+	Run: func(_ *cobra.Command, _ []string) {
+		context := GetContext()
+		health.Health(context, OperatorNamespace, CephClusterNamespace)
+	},
 }


subhamkrai (commented):
The logs are not mirrored images of what we have in bash, since keeping the same logs as bash is difficult to maintain in golang. Also, will have a pr in the future(very soon to modify the logging system)

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+
+func checkMgrPodsStatusAndCounts(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
+	opts := metav1.ListOptions{LabelSelector: "app=rook-ceph-mgr"}
+	podList, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), opts)
+	if err != nil {
+		fmt.Printf("\nfailed to list mgr pods with label %s: %v\n", opts.LabelSelector, err)
+	}
+
+	if len(podList.Items) < 1 {
+		fmt.Println("At least one mgr pod should running")
```
travisn, line 156:
```suggestion
		fmt.Println("At least one mgr pod should be running")
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		fmt.Printf("\nfailed to list pods in namespace %s: %v\n", operatorNamespace, err)
+	}
+
+	podListInClNs, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("\nfailed to list pods in namespace %s: %v\n", clusterNamespace, err)
+	}
+
+	var PodNotRunning, PodRunning []v1.Pod
```
travisn, line 107: Local var names should be lower case
```suggestion
	var podNotRunning, podRunning []v1.Pod
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func CheckAllPodsStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
+	podListInOpNs, err := context.Clientset.CoreV1().Pods(operatorNamespace).List(ctx.TODO(), metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("\nfailed to list pods in namespace %s: %v\n", operatorNamespace, err)
+	}
+
+	podListInClNs, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), metav1.ListOptions{})
```
travisn, line 102:
```suggestion
	clusterNamespacePods, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), metav1.ListOptions{})
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+		fmt.Printf("\n%s", cephHealthDetails)
+	} else if cephHealthDetails == "HEALTH_WARN" {
+		fmt.Println("Warning: ", cephHealthDetails)
+	} else if cephHealthDetails == "HEALTH_ERR" {
+		fmt.Println("Error: ", cephHealthDetails)
+	}
+}
+
+func CheckAllPodsStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
+	podListInOpNs, err := context.Clientset.CoreV1().Pods(operatorNamespace).List(ctx.TODO(), metav1.ListOptions{})
```
travisn, line 97: for better readability, spell out var names. If they're long, then find simpler names.
```suggestion
	operatorNamespacePods, err := context.Clientset.CoreV1().Pods(operatorNamespace).List(ctx.TODO(), metav1.ListOptions{})
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	for i := range podList.Items {
+		fmt.Println(podList.Items[i].Name, "\t", podList.Items[i].Status.Phase, "\t", podList.Items[i].Namespace, "\t", podList.Items[i].Spec.NodeName)
+	}
+}
+
+func checkMonQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
+	cephHealthDetails := exec.RunCommandInOperatorPod(context, "ceph", []string{"health", "detail"}, operatorNamespace, clusterNamespace, false)
+	if cephHealthDetails == "HEALTH_OK" {
```
travisn, line 87: `cephHealthDetails` contains a lot of status, right? Seems like this will never be equal.

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+
+	fmt.Println("Info:  checking placement group status")
+	checkPgStatus(context, operatorNamespace, clusterNamespace, exitOnError)
+
+	fmt.Println()
+	fmt.Println("Info:  checking if at least one mgr pod is running")
+	checkMgrPodsStatusAndCounts(context, operatorNamespace, clusterNamespace, exitOnError)
+}
+
+func checkPodsOnNodes(context *k8sutil.Context, operatorNamespace, clusterNamespace, label string, exitOnError bool) {
```
travisn, line 54: not used
```suggestion
func checkPodsOnNodes(context *k8sutil.Context, operatorNamespace, clusterNamespace, label string) {
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+)
+
+func Health(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
```
travisn, line 31: No need for the parameter, just define the local var to keep the method signature simpler. Or maybe this local var isn't even needed if each of these child methods can just pass `false`
```suggestion
func Health(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
   exitOnError := true
```

travisn (changes requested):
mainly some nits

pkg/health/health.go
subhamkrai: one of my favourite mistakes 😅

pkg/health/health.go
subhamkrai: right, currently only using `strings.Contains()` but I'm planning on working locally, to use json marshaling whenever we are comparing ceph output status but will not be part of this pr since it will require changes in other commands too.

pkg/health/health.go
subhamkrai: got it

.github/workflows/go-test.yaml
```diff
@@ -35,6 +35,8 @@ jobs:
       - name: Test plugin with ceph commands
         run: |
           set -e
+          kubectl rook-ceph health
+
           kubectl rook-ceph ceph status
```
travisn, line 40:
```suggestion
      - name: ceph status
          kubectl rook-ceph ceph status
```

.github/workflows/go-test.yaml
```diff
@@ -35,6 +35,8 @@ jobs:
       - name: Test plugin with ceph commands
```
travisn, line 35: Can we get these tests all under different `name`s in the test? Or is that going to be a separate PR?
```suggestion
      - name: rook-ceph health
```

.github/workflows/go-test.yaml
travisn: and so on

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+)
+
+func Health(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	exitOnError := true
+
+	fmt.Println("Info:  Checking if at least three mon pods are running on different nodes")
```
travisn, line 34: Why the extra space?
```suggestion
	fmt.Println("Info: Checking if at least three mon pods are running on different nodes")
```

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	checkMonQuorum(context, operatorNamespace, clusterNamespace, exitOnError)
+
+	fmt.Println()
+	fmt.Println("Info:  Checking if at least three osd pods are running on different nodes")
+	checkPodsOnNodes(context, operatorNamespace, clusterNamespace, "app=rook-ceph-osd")
+
+	fmt.Println()
+	CheckAllPodsStatus(context, operatorNamespace, clusterNamespace, exitOnError)
+
+	fmt.Println("Info:  checking placement group status")
```
travisn, line 48:
```suggestion
	fmt.Println("Info: Checking placement group status")
```

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	checkPgStatus(context, operatorNamespace, clusterNamespace, exitOnError)
+
+	fmt.Println()
+	fmt.Println("Info:  checking if at least one mgr pod is running")
```
travisn, line 52:
```suggestion
	fmt.Println("Info: Checking if at least one mgr pod is running")
```

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	if strings.Contains(cephHealthDetails, "HEALTH_OK") {
+		fmt.Printf("\n%s", cephHealthDetails)
+	} else if strings.Contains(cephHealthDetails, "HEALTH_WARN") {
+		fmt.Println("Warning: ", cephHealthDetails)
+	} else if strings.Contains(cephHealthDetails, "HEALTH_ERR") {
+		fmt.Println("Error: ", cephHealthDetails)
+	}
+}
+
+func CheckAllPodsStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
```
travisn, line 98: not used
```suggestion
func CheckAllPodsStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
```

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	operatorNamespacePods, err := context.Clientset.CoreV1().Pods(operatorNamespace).List(ctx.TODO(), metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("\nfailed to list pods in namespace %s: %v\n", operatorNamespace, err)
+	}
+
+	clusterNamespacePods, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), metav1.ListOptions{})
```
travisn, line 104: If the operator and cluster namespace are the same, will these pods be duplicate of the operatorNamespacePods?

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		fmt.Printf("\nfailed to list pods in namespace %s: %v\n", clusterNamespace, err)
+	}
+
+	var podNotRunning, podRunning []v1.Pod
+
+	for i := range operatorNamespacePods.Items {
```
travisn, line 111: How about a helper method for this `for` loop? so we don't have to duplicate it on line 119.

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+		fmt.Println(podRunning[i].Name, "\t", podRunning[i].Status.Phase, "\t", podRunning[i].Namespace, "\t", podRunning[i].Spec.NodeName)
+	}
+
+	fmt.Println("\nWarning: Pods that are 'Not' in 'Running' status")
+	for i := range podNotRunning {
+		fmt.Println(podNotRunning[i].Name, "\t", podNotRunning[i].Status.Phase, "\t", podNotRunning[i].Namespace, "\t", podNotRunning[i].Spec.NodeName)
+	}
+}
+
+func checkPgStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
```
travisn, line 138:
```suggestion
func checkPgStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
```

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	if strings.Contains(pg_stat_code, "active+clean;") {
+		fmt.Printf("\nINFO: %s\n", pg_stat_code)
+	} else if strings.Contains(pg_stat_code, "down") || strings.Contains(pg_stat_code, "incomplete") || strings.Contains(pg_stat_code, "snaptrim_error") {
+		fmt.Printf("\nError: %s\n", pg_stat_code)
+	} else {
+		fmt.Printf("\nWarning: %s\n", pg_stat_code)
+	}
+}
+
+func checkMgrPodsStatusAndCounts(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
```
travisn, line 150:
```suggestion
func checkMgrPodsStatusAndCounts(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
```

travisn (changes requested):

pkg/health/health.go
```diff
@@ -0,0 +1,163 @@
...
+	fmt.Println("\nWarning: Pods that are 'Not' in 'Running' status")
+	for i := range podNotRunning {
+		fmt.Println(podNotRunning[i].Name, "\t", podNotRunning[i].Status.Phase, "\t", podNotRunning[i].Namespace, "\t", podNotRunning[i].Spec.NodeName)
+	}
+}
+
+func checkPgStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
+	pg_stat_code := exec.RunCommandInOperatorPod(context, "ceph", []string{"pg", "stat"}, operatorNamespace, clusterNamespace, false)
+
+	if strings.Contains(pg_stat_code, "active+clean;") {
```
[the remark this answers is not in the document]
subhamkrai: right, we need to find a better way. 
We can try using the json parsing with ceph commands but n the other hand, it will put them on either `Info, Warn or Error` mode. We'll have to log for each pg status type.

Or We can just print the ceph command.
thoughts?

For, now I have move the condition checking for `error`  first so that is any of the error state is there is the pg status we say it is `Error`

pkg/health/health.go
subhamkrai: done

pkg/health/health.go
subhamkrai: good catch, now it should be good

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(nodeList) < 3 {
+		fmt.Printf("\nWarning: At least three %s pods should running on different nodes", daemonType)
+	}
+
+	for i := range podList.Items {
+		fmt.Println(podList.Items[i].Name, "\t", podList.Items[i].Status.Phase, "\t", podList.Items[i].Namespace, "\t", podList.Items[i].Spec.NodeName)
+	}
+}
+
+func checkMonQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace string, exitOnError bool) {
```
travisn, line 87: not used
```suggestion
func checkMonQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	var nodeList = make(map[string]string)
+	for i := range podList.Items {
+		nodeName := podList.Items[i].Spec.NodeName
+		if _, okay := nodeList[nodeName]; !okay {
+			nodeList[nodeName] = podList.Items[i].Name
+		}
+	}
+
+	if len(nodeList) < 3 {
+		fmt.Printf("\nWarning: At least three %s pods should running on different nodes", daemonType)
```
travisn, line 79: Looks like the println on line 83 will end up on the same line as this warning
```suggestion
		fmt.Printf("\nWarning: At least three %s pods should running on different nodes\n", daemonType)
```

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	return podRunning, podNotRunning
+}
+
+func checkPgStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	pg_stat_code := exec.RunCommandInOperatorPod(context, "ceph", []string{"pg", "stat"}, operatorNamespace, clusterNamespace, false)
+
+	if strings.Contains(pg_stat_code, "down") || strings.Contains(pg_stat_code, "incomplete") || strings.Contains(pg_stat_code, "snaptrim_error") {
+		fmt.Printf("\nError: %s\n", pg_stat_code)
+	} else if strings.Contains(pg_stat_code, "active+clean;") {
```
travisn, line 141: There are a lot of pg statuses, I wonder if we are missing something. Can we check if the only status is exactly `active+clean` for this case?

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+		fmt.Printf("\nInfo: %s\n", pg_stat_code)
+	} else {
+		fmt.Printf("\nWarning: %s\n", pg_stat_code)
```
travisn, line 144: I wonder if we can just print info/warning/error depending on ceph's HEALTH_OK, HEALTH_WARN, or HEALTH_ERR. Then we don't need to try and guess what the state is, we can just let ceph tell us. For the case where the health issue is because of something other than PGs, we could just print info if all the PGs are exactly `active+clean`, otherwise just match ceph's health status for this output.

pkg/health/health.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func checkMgrPodsStatusAndCounts(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	opts := metav1.ListOptions{LabelSelector: "app=rook-ceph-mgr"}
+	podList, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), opts)
+	if err != nil {
+		fmt.Printf("\nfailed to list mgr pods with label %s: %v\n", opts.LabelSelector, err)
```
travisn, line 152: need to return here? the `podList` won't be valid below
```suggestion
		fmt.Printf("\nfailed to list mgr pods with label %s: %v\n", opts.LabelSelector, err)
		return
```

travisn (changes requested):

pkg/health/health.go
subhamkrai: should be better now with Json parsing.

pkg/health/health.go
subhamkrai: should be better now with Json parsing.

pkg/health/health.go
travisn: I see, `unMarshalCephCommands()` only returns the status. Then we need to rename that method and also `cephHealthDetails` so it makes sense that it's only the status.

pkg/health/health.go
```diff
@@ -0,0 +1,199 @@
...
+		fmt.Printf("\nWarning: At least three %s pods should running on different nodes\n", daemonType)
+	}
+
+	for i := range podList.Items {
+		fmt.Println(podList.Items[i].Name, "\t", podList.Items[i].Status.Phase, "\t", podList.Items[i].Namespace, "\t", podList.Items[i].Spec.NodeName)
+	}
+}
+
+func checkMonQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	cephHealthDetails, _ := unMarshalCephCommands(context, operatorNamespace, clusterNamespace)
```
travisn, line 108:
```suggestion
	cephHealthStatus, _ := unMarshalCephStatus(context, operatorNamespace, clusterNamespace)
```

travisn (changes requested):

pkg/health/health.go
subhamkrai: done

pkg/health/health.go
```diff
@@ -0,0 +1,199 @@
...
+	if strings.Contains(label, "osd") {
+		daemonType = "osd"
+	} else if strings.Contains(label, "mon") {
+		daemonType = "mon"
+	}
+
+	opts := metav1.ListOptions{LabelSelector: label}
+	podList, err := context.Clientset.CoreV1().Pods(clusterNamespace).List(ctx.TODO(), opts)
+	if err != nil {
+		fmt.Printf("\nfailed to list %s pods with label %s: %v\n", daemonType, opts.LabelSelector, err)
```
travisn, line 87: return on error? The `podList` isn't valid.

travisn (approved):
Just one more small question, feel free to merge after fixing.

pkg/health/health.go
subhamkrai: right, done

pkg/health/health.go
```diff
@@ -0,0 +1,205 @@
...
+			podNotRunning = append(podNotRunning, podList.Items[i])
+		} else {
+			podRunning = append(podRunning, podList.Items[i])
+		}
+	}
+	return podRunning, podNotRunning
+}
+
+func checkPgStatus(context *k8sutil.Context, operatorNamespace, clusterNamespace string) {
+	_, PgStateEntryList := unMarshalCephStatus(context, operatorNamespace, clusterNamespace)
```
travisn, line 162:
```suggestion
	_, pgStateEntryList := unMarshalCephStatus(context, operatorNamespace, clusterNamespace)
```

subhamkrai: done

subhamkrai:
Go ahead with this pr, since go-test ci is passing `with-pr-changes-in-custom-namespace` is failing but only valid for a bash script.

PR #91, opened 2023-04-27
core: add dr health command in go

subhamkrai:
core: add dr health command in go

testing result
```
./bin/kubectl-rook-ceph dr health
INFO: fetching the cephblockpools with mirroring enabled
found ocs-storagecluster-cephblockpool cephblockpool with mirroring enabled
running ceph status from peer cluster

  cluster:
    id:     3d724ca6-a0d9-4299-8ab0-f136c145357a
    health: HEALTH_OK
 
  services:
    mon:        3 daemons, quorum d,e,f (age 26h)
    mgr:        a(active, since 35h)
    mds:        1/1 daemons up, 1 hot standby
    osd:        3 osds: 3 up (since 26h), 3 in (since 35h)
    rbd-mirror: 1 daemon active (1 hosts)
    rgw:        1 daemon active (1 hosts, 1 zones)
 
  data:
    volumes: 1/1 healthy
    pools:   12 pools, 169 pgs
    objects: 579 objects, 510 MiB
    usage:   2.3 GiB used, 1.5 TiB / 1.5 TiB avail
    pgs:     169 active+clean
 
  io:
    client:   1023 B/s rd, 12 KiB/s wr, 1 op/s rd, 1 op/s wr
 

running mirroring daemon health

health: OK
daemon health: OK
image health: OK
images: 0 total
```

```
./bin/kubectl-rook-ceph ceph -s

  cluster:
    id:     2d411e76-2e99-4d80-a6eb-76a6864b122f
    health: HEALTH_OK
 
  services:
    mon:        3 daemons, quorum d,e,f (age 24h)
    mgr:        a(active, since 35h)
    mds:        1/1 daemons up, 1 hot standby
    osd:        3 osds: 3 up (since 24h), 3 in (since 35h)
    rbd-mirror: 1 daemon active (1 hosts)
    rgw:        1 daemon active (1 hosts, 1 zones)
 
  data:
    volumes: 1/1 healthy
    pools:   12 pools, 169 pgs
    objects: 616 objects, 594 MiB
    usage:   3.4 GiB used, 1.5 TiB / 1.5 TiB avail
    pgs:     169 active+clean
 
  io:
    client:   5.6 KiB/s rd, 43 KiB/s wr, 7 op/s rd, 6 op/s wr
```

we can see that command is able to get ceph status of mirroring cluster.


subhamkrai (commented):
FYI, I don't have CI for dr health but I have tested the changes and shared the output. 
I'll add the CI for dr health soon

pkg/dr/health.go
```diff
@@ -0,0 +1,81 @@
...
+	rookv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type secretData struct {
+	Key     string `json:"key"`
+	MonHost string `json:"mon_host"`
+}
+
+func Health(context *k8sutil.Context, operatorNamespace, cephClusterNamespace string, args []string, ExitOnError bool) {
```
travisn, line 23: not used
```suggestion
func Health(context *k8sutil.Context, operatorNamespace, cephClusterNamespace string, args []string) {
```

pkg/dr/health.go
```diff
@@ -0,0 +1,81 @@
...
+	blockPoolList, err := context.RookClientset.CephV1().CephBlockPools(cephClusterNamespace).List(ctx.TODO(), v1.ListOptions{})
+	if err != nil {
+		log.Error(err)
+	}
+
+	var mirrorBlockPool rookv1.CephBlockPool
+	for _, blockPool := range blockPoolList.Items {
+		if blockPool.Spec.Mirroring.Enabled {
+			mirrorBlockPool = blockPool
+			fmt.Printf("found %s cephblockpool with mirroring enabled\n", mirrorBlockPool.Name)
```
travisn, line 34: Quotes around the resource name is a nice pattern, even if it is slightly different from the script plugin
```suggestion
			fmt.Printf("found %q cephblockpool with mirroring enabled\n", mirrorBlockPool.Name)
```

pkg/dr/health.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	secretNameList := mirrorBlockPool.Spec.Mirroring.Peers.SecretNames
+	secretName := secretNameList[0]
+	secretName = strings.TrimSuffix(secretName, "]")
+	secretName = strings.TrimPrefix(secretName, "[")
+
+	secret, err := context.Clientset.CoreV1().Secrets(cephClusterNamespace).Get(ctx.TODO(), secretName, v1.GetOptions{})
+	if err != nil {
+		log.Fatal(err)
```
travisn, line 49: Instead of making these fatal, how about only failing this section of the output? For example, factor out lines 42-76 to a separate method where the secret must be parsed. Then even if the secret fails to load, the status on lines 78-80 can still continue.

travisn (changes requested):
Looks like pulling in the rook api worked nicely!

pkg/dr/health.go
subhamkrai: got it

pkg/dr/health.go
subhamkrai: we want the secrets details, otherwise ceph status will loaded from local cluster and not from pee cluster.

subhamkrai: but refactored based on secret extraction

pkg/dr/health.go
```diff
@@ -0,0 +1,94 @@
...
+	fmt.Println(cephStatus)
+
+	fmt.Println("running mirroring daemon health")
+
+	fmt.Println(exec.RunCommandInOperatorPod(context, "rbd", []string{"-p", mirrorBlockPool.Name, "mirror", "pool", "status"}, cephClusterNamespace, operatorNamespace, true))
+}
+
+func extractSecretName(context *k8sutil.Context, operatorNamespace, cephClusterNamespace, secretDetail string) (*secretData, error) {
+
+	secretName := strings.TrimSuffix(secretDetail, "]")
```
sp98, line 73: `secretDetail` is a string and the name of the secret. Can you please elaborate why the `TrimSuffix` method needed here?

pkg/dr/health.go
```diff
@@ -0,0 +1,94 @@
...
+
+	secretData, err := extractSecretName(context, operatorNamespace, cephClusterNamespace, secretDetail)
+	if err != nil {
+		fmt.Printf("failed to extract secret %s", secretDetail)
+		log.Fatal(err)
+	}
+
+	fmt.Println("running ceph status from peer cluster")
+
+	cephArgs := []string{"-s", "--mon-host", secretData.MonHost, "--id", "rbd-mirror-peer", "--key", secretData.Key}
```
sp98, line 52: client-id (`--id`) should be retrieved from the decoded token as well.

pkg/dr/health.go
```diff
@@ -0,0 +1,94 @@
...
+	fmt.Println(exec.RunCommandInOperatorPod(context, "rbd", []string{"-p", mirrorBlockPool.Name, "mirror", "pool", "status"}, cephClusterNamespace, operatorNamespace, true))
+}
+
+func extractSecretName(context *k8sutil.Context, operatorNamespace, cephClusterNamespace, secretDetail string) (*secretData, error) {
+
+	secretName := strings.TrimSuffix(secretDetail, "]")
+	secretName = strings.TrimPrefix(secretName, "[")
+	// TODO: While testing Rook canary mirroring test, found that there is extra suffic `-config`
+	// in blockPool cr under mirroring which needs to be removed. Below is working but need to do RCA.
+	// secretName = strings.TrimSuffix(secretName, `-config`)
```
sp98, line 77: The CI test patches the secret name to the cephBlockPool CR
`kubectl -n rook-ceph patch cephblockpool replicapool --type merge -p '{"spec":{"mirroring":{"peers": {"secretNames": ["pool-peer-token-replicapool-config"]}}}}'`

This name is being send as an argument to this method (`secretDetail`). I don't think we need the `TrimPrefix` method here.

subhamkrai: got it, removed.

pkg/dr/health.go
subhamkrai: you are right, it is not required. removed.

pkg/dr/health.go
subhamkrai: thanks, should be good now.

pkg/dr/health.go
```diff
@@ -0,0 +1,87 @@
...
+			mirrorBlockPool = blockPool
+			fmt.Printf("found %q cephblockpool with mirroring enabled\n", mirrorBlockPool.Name)
+		}
+	}
+
+	if mirrorBlockPool.Name == "" {
+		fmt.Printf("DR is not confiqured, cephblockpool with mirroring enabled not found.\n")
+	}
+
+	secretDetail := mirrorBlockPool.Spec.Mirroring.Peers.SecretNames[0]
```
travisn, line 42: What if there are multiple pools that are enabled for mirroring? The loop above only will assign the last one found to this `mirrorBlockPool`. Seems like we should instead loop over all of the mirrored pools.

pkg/dr/health.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if len(args) == 0 {
+		cephArgs = append(cephArgs, "--debug-ms", "0")
+	} else {
+		cephArgs = append(cephArgs, args...)
+	}
+
+	cephStatus := exec.RunCommandInOperatorPod(context, "ceph", cephArgs, operatorNamespace, cephClusterNamespace, false)
+	if cephStatus == "" {
+		fmt.Println("failed to get ceph status from peer cluster, please check for network issues between the clusters")
```
travisn, line 62: Need to return if there was an error? If we can continue with other commands, line 64 just needs to be in an `else` block.

pkg/dr/health.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	fmt.Println(cephStatus)
+
+	fmt.Println("running mirroring daemon health")
+
+	fmt.Println(exec.RunCommandInOperatorPod(context, "rbd", []string{"-p", mirrorBlockPool.Name, "mirror", "pool", "status"}, cephClusterNamespace, operatorNamespace, true))
+}
+
+func extractSecretName(context *k8sutil.Context, operatorNamespace, cephClusterNamespace, secretName string) (*secretData, error) {
```
travisn, line 71:
```suggestion
func extractSecretData(context *k8sutil.Context, operatorNamespace, cephClusterNamespace, secretName string) (*secretData, error) {
```

pkg/dr/health.go
```diff
@@ -0,0 +1,87 @@
...
+	secret, err := context.Clientset.CoreV1().Secrets(cephClusterNamespace).Get(ctx.TODO(), secretName, v1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	decodeSecretData, err := base64.StdEncoding.DecodeString(string(secret.Data["token"]))
+	if err != nil {
+		return nil, err
+	}
+
+	var secretData *secretData
```
travisn, line 81: nit: I'd suggest a different var name from the type name, it's a bit confusing if the same.
```suggestion
	var data *secretData
```

travisn (changes requested):

pkg/dr/health.go
subhamkrai: right

pkg/dr/health.go
subhamkrai: I'm looping over secret name and when the name is not empty break out of the loop.

go.mod
```diff
@@ -3,56 +3,133 @@ module github.com/rook/kubectl-rook-ceph
...
 
 require (
+	github.com/rook/rook v1.11.4
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.6.1
-	github.com/stretchr/testify v1.8.0
-	k8s.io/api v0.26.0
-	k8s.io/apimachinery v0.26.0
-	k8s.io/client-go v0.26.0
+	github.com/stretchr/testify v1.8.1
```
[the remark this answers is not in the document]
subhamkrai: updated the rook version but didn't help

cmd/commands/dr_health.go
```diff
@@ -0,0 +1,28 @@
...
+var DrCmd = &cobra.Command{
+	Use:                "dr",
+	Short:              "Calls subcommand health",
+	DisableFlagParsing: true,
+	Args:               cobra.ExactArgs(1),
+}
+
+var healthCmd = &cobra.Command{
+	Use:                "health",
+	Short:              "Print the ceph status of a peer cluster in a mirroring-enabled cluster. Optional ceph args such as: '--debug-ms 1'.",
```
travisn, line 17: It seems odd to include the optional arg in the command description, they should be able to discovery the args in another place
```suggestion
	Short:              "Print the ceph status of a peer cluster in a mirroring-enabled cluster,
```

travisn (approved):
approved with a small suggestion

cmd/commands/dr_health.go
subhamkrai: done

pkg/dr/health.go
```diff
@@ -0,0 +1,94 @@
...
+	fmt.Println("INFO: fetching the cephblockpools with mirroring enabled")
+	blockPoolList, err := context.RookClientset.CephV1().CephBlockPools(cephClusterNamespace).List(ctx.TODO(), v1.ListOptions{})
+	if err != nil {
+		log.Error(err)
+	}
+
+	var mirrorBlockPool rookv1.CephBlockPool
+	for _, blockPool := range blockPoolList.Items {
+		if blockPool.Spec.Mirroring.Enabled && blockPool.Spec.Mirroring.Peers != nil {
+			mirrorBlockPool = blockPool
```
sp98, line 33: can't there be more than one cephblockpools with mirroring enabled?

pkg/dr/health.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+type secretData struct {
+	Key      string `json:"key"`
+	MonHost  string `json:"mon_host"`
+	ClientId string `json:"client_id"`
+}
+
+func Health(context *k8sutil.Context, operatorNamespace, cephClusterNamespace string, args []string) {
+	fmt.Println("INFO: fetching the cephblockpools with mirroring enabled")
```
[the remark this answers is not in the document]
subhamkrai: yes, PR is in review

pkg/dr/health.go
subhamkrai: right, and I think in that case we will need to ask the user to pass the pool name but since currently we don't have that support/requirement we are good. wdyt?

sp98 (approved):

PR #92, opened 2023-04-28
ci: move commands in specific section

subhamkrai:
ci: move commands in a specific section

currently, all the commands are in the same
test name, it's better to split the commands
in the individual section.

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -32,13 +32,19 @@ jobs:
           sudo cp bin/kubectl-rook-ceph /usr/local/bin/kubectl-rook_ceph
           make test
 
-      - name: Test plugin with ceph commands
+      - name: Cluster Health
         run: |
           set -e
           kubectl rook-ceph health
 
+      - name: Ceph status
+        run: |
+          set -ex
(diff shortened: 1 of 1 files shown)


travisn:
LGTM after we get #90 merged

.github/workflows/go-test.yaml
```diff
@@ -48,25 +54,41 @@ jobs:
           kubectl -n rook-ceph wait deployment rook-ceph-mon-d --for condition=Available=True --timeout=90s
           kubectl -n rook-ceph wait deployment rook-ceph-mon-e --for condition=Available=True --timeout=90s
 
+      - name: Rbd command
+        run: |
+          set -ex
+          kubectl rook-ceph rbd ls replicapool
 
+      - name: Basic command
```
travisn, line 62: Even this section, could you break it out to one command per section?

travisn (changes requested):

.github/workflows/go-test.yaml
subhamkrai: done

travisn (approved):

PR #93, opened 2023-05-02
core: update logging, remove unused var

subhamkrai:
core: update logging, remove unused var

This commits brings bash color logging in GoLang.
Also, it has little cleanup with unused variable/function
and improve loggng.

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -16,9 +16,8 @@ limitations under the License.
 package command
 
 import (
-	"log"
-
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	rookclient "github.com/rook/rook/pkg/client/clientset/versioned"
 	"github.com/spf13/cobra"
 
@@ -67,17 +66,22 @@ func GetContext() *k8sutil.Context {
 
 	context.KubeConfig, err = kubeconfig.ClientConfig()
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 
 	context.RookClientset, err = rookclient.NewForConfig(context.KubeConfig)
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
+	}
+
+	context.RookClientset, err = rookclient.NewForConfig(context.KubeConfig)
+	if err != nil {
+		logging.Fatal(err)
 	}
 
 	context.Clientset, err = k8s.NewForConfig(context.KubeConfig)
 	if err != nil {
-		log.Fatal(err)
+		logging.Fatal(err)
 	}
 
 	return context
--- a/pkg/debug/start_debug.go
+++ b/pkg/debug/start_debug.go
@@ -19,11 +19,10 @@ package debug
 import (
 	ctx "context"
 	"fmt"
-	"log"
-	"os"
 	"time"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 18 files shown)


cmd/commands/root.go
```diff
@@ -66,12 +65,12 @@ func GetContext() *k8sutil.Context {
 
 	context.KubeConfig, err = kubeconfig.ClientConfig()
 	if err != nil {
-		log.Fatal(err)
+		logging.ErrorLogging(err)
```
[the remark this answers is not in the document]
travisn: How about a new helper, which is `logging.Fatal()`?

pkg/logging/log.go
```diff
@@ -0,0 +1,38 @@
...
+
+package logging
+
+import (
+	"fmt"
+
+	"github.com/fatih/color"
+)
+
+func InfoLogging(output string) {
```
travisn, line 25: How about naming these methods `Info()`, `Warning()`, and `Error()`? The "logging" is already in the package name, so we don't need it in the method name.
```suggestion
func Info(output string) {
```

pkg/logging/log.go
```diff
@@ -0,0 +1,38 @@
...
+	color.Blue("Info:")
+	fmt.Printf("%s\n", output)
+}
+
+func WarningLogging(output string) {
```
travisn, line 30: Instead of the calling needing to call format on the args, they could be passed like this and then this method can format it.
```suggestion
func Warning(output string, args ...string) {
```

travisn (changes requested):

pkg/logging/log.go
travisn: It should be `interface` instead of `string`. In the rook code base I see similar helpers that make it nicer for logging.
```go
Infof(format string, args ...interface{})
Debugf(format string, args ...interface{})
Warningf(format string, args ...interface{})
```

subhamkrai: done

pkg/debug/start_debug.go
```diff
@@ -48,7 +46,7 @@ func startDebug(context *k8sutil.Context, clusterNamespace, deploymentName, alte
 	deployment := *originalDeployment
 
 	if alternateImageValue != "" {
-		log.Printf("setting debug image to %s\n", alternateImageValue)
+		logging.Info(fmt.Sprintf("setting debug image to %s\n", alternateImageValue))
```
travisn, line 49: No more need for `Sprintf()` when logging
```suggestion
		logging.Info("setting debug image to %s\n", alternateImageValue)
```

pkg/logging/log.go
```diff
@@ -0,0 +1,44 @@
...
+package logging
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/fatih/color"
+)
+
+func Info(output string) {
```
travisn, line 26:
```suggestion
func Info(output string, args ...interface{}) {
```

pkg/logging/log.go
```diff
@@ -0,0 +1,44 @@
...
+	color.Blue("Info:")
+	fmt.Printf("%s\n", output)
```
travisn, line 28: With this pattern I believe this will allow you to get rid of all the `Sprintf()` calls when the logging methods are called. 
```suggestion
	fmt.Printf(output, args...)
```

travisn: All the Info, Warning, Error, and Fatal methods here should follow the same pattern.

travisn (changes requested):

pkg/logging/log.go
subhamkrai: done

pkg/logging/log.go
```diff
@@ -0,0 +1,47 @@
...
+package logging
+
+import (
+	"fmt"
+
+	"github.com/fatih/color"
+)
+
+func Info(output string, args ...interface{}) {
+	color.Blue("Info:")
```
travisn, line 26: In the CI output, all the "Info:" show up on their own line. Is there a way to combine them on the same line as the rest of the info? The output looks awkward currently with no details on the same line as "Info:"

travisn (changes requested):

pkg/logging/log.go
subhamkrai: agree, should be good now

travisn (approved):

PR #94, opened 2023-05-03
core: don't use context.ToDo everywhere

subhamkrai:
core: don't use context.ToDo everywhere

using context.ToDo always is not a good practice.
So, now having a central place of context which
all methods will use.

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -84,5 +84,7 @@ func GetContext() *k8sutil.Context {
 		logging.Fatal(err)
 	}
 
+	context.Context = RootCmd.Context()
+
 	return context
 }
--- a/pkg/debug/start_debug.go
+++ b/pkg/debug/start_debug.go
@@ -17,7 +17,6 @@ limitations under the License.
 package debug
 
 import (
-	ctx "context"
 	"fmt"
 	"time"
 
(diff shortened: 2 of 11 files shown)


travisn:
Do you see a difference in the behavior of killing a running command with Control-C with this context change? I'm assuming it will cancel the request sooner.

subhamkrai:
> Do you see a difference in the behavior of killing a running command with Control-C with this context change? I'm assuming it will cancel the request sooner.

well for me, it is fast with or without these changes I am hardly able to find a diff but when I used `time ` command I see diff

without these changes
```yaml
real	0m0.589s
user	0m0.028s
sys	0m0.014s
```

with these changes
```yaml
real	0m0.387s
user	0m0.052s
sys	0m0.022s
```

subhamkrai:
will rebase once logging pr is merged

cmd/commands/root.go
```diff
@@ -84,5 +86,7 @@ func GetContext() *k8sutil.Context {
 		logging.Fatal(err)
 	}
 
+	context.Context = ctx.TODO()
```
travisn, line 89: I thought the intention was to not use `ctx.TODO()`.

subhamkrai: now, using cobra root command context. Similar to multus validation in rook

travisn (approved):

Issue #97, opened 2023-05-19
Validation not working for restore-quorum implementation in golang


travisn:
When running the restore-quorum command, the user must type "yes-really-restore". However, if the user types anything, the command will continue instead of expecting that exact string.

Repro:
- ` bin/kubectl-rook-ceph mons restore-quorum a`
- Type something other than `yes-really-restore`
- See the restore continue instead of aborted

Also, the closing quote on the prompt is missing:
```
Warning: Are you sure you want to restore the quorum to mon a? If so, enter 'yes-really-restore
```

travisn:
This message is printed, but I do not have that env var set:
```
Info: skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true
```

PR #104, opened 2023-05-22
mons: improve mon restore quorum

subhamkrai:
validating user input `yes-really-restore` and also fixing a couple of typos.

fixes: #97

### DIFF
--- a/pkg/mons/restore_quorum.go
+++ b/pkg/mons/restore_quorum.go
@@ -84,7 +84,7 @@ func restoreQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace
 	logging.Info("The cluster fsid is %s\n", cephFsid)
 
 	var answer, output string
-	logging.Warning("Are you sure you want to restore the quorum to mon %s? If so, enter 'yes-really-restore\n", goodMon)
+	logging.Warning("Are you sure you want to restore the quorum to mon %s? If so, enter 'yes-really-restore'\n", goodMon)
 	fmt.Scanf("%s", &answer)
 	output, err = promptToContinueOrCancel(answer)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/mons/restore_quorum.go
```diff
@@ -84,7 +84,7 @@ func restoreQuorum(context *k8sutil.Context, operatorNamespace, clusterNamespace
 	logging.Info("The cluster fsid is %s\n", cephFsid)
 
 	var answer, output string
-	logging.Warning("Are you sure you want to restore the quorum to mon %s? If so, enter 'yes-really-restore\n", goodMon)
+	logging.Warning("Are you sure you want to restore the quorum to mon %s? If so, enter 'yes-really-restore`\n", goodMon)
```
travisn, line 87:
```suggestion
	logging.Warning("Are you sure you want to restore the quorum to mon %s? If so, enter 'yes-really-restore'\n", goodMon)
```

pkg/mons/restore_quorum.go
```diff
@@ -279,13 +279,13 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 func promptToContinueOrCancel(answer string) (string, error) {
 	var ROOK_PLUGIN_SKIP_PROMPTS string
 	_, ok := os.LookupEnv(ROOK_PLUGIN_SKIP_PROMPTS)
-	if ok {
+	if !ok {
 		if answer == "yes-really-restore" {
 			return "proceeding", nil
 		} else if answer == "" {
 			return "continuing", nil
```
travisn, line 286: If the answer is blank we don't want to continue. They have to type the string to continue.

travisn (changes requested):

pkg/mons/restore_quorum.go

travisn (commented):
Now the CI is failing with this change

pkg/mons/restore_quorum.go
```diff
@@ -277,15 +277,12 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 }
 
 func promptToContinueOrCancel(answer string) (string, error) {
-	var ROOK_PLUGIN_SKIP_PROMPTS string
-	_, ok := os.LookupEnv(ROOK_PLUGIN_SKIP_PROMPTS)
-	if ok {
+	_, ok := os.LookupEnv("ROOK_PLUGIN_SKIP_PROMPTS")
```
travisn, line 280: Instead of just seeing if the env var exists, let's check if it also equals `true`.

travisn (approved):

pkg/mons/restore_quorum.go
```diff
@@ -277,17 +277,15 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
-	var ROOK_PLUGIN_SKIP_PROMPTS string
-	_, ok := os.LookupEnv(ROOK_PLUGIN_SKIP_PROMPTS)
-	if ok {
-		if answer == "yes-really-restore" {
-			return "proceeding", nil
-		} else if answer == "" {
-			return "continuing", nil
-		} else {
-			return "", fmt.Errorf("canncelled")
+	if skip, ok := os.LookupEnv("ROOK_PLUGIN_SKIP_PROMPTS"); ok {
```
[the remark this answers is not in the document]
subhamkrai: it's better

Issue #102, opened 2023-05-19
Blank info lines should not print "Info:"


travisn:
There are many blank lines with “Info:”. There should only be “Info” if there is some text on the same line. So if the helper gets a blank info request, we should just print a blank line without "Info"

```
% bin/kubectl-rook-ceph rook status all
...
Info: cephrbdmirrors.ceph.rook.io
Info: 
Info: objectbucketclaims.objectbucket.io
Info: 
Info: objectbuckets.objectbucket.io
Info: 
Info: 
Info: 
```

PR #107, opened 2023-05-22
status: remove extra empty line logging

subhamkrai:
status: remove extra empty line logging

changing Info() method to print `Info` and
logs with new line only when output is not empty
otherwise it will print `Info:` with empty line.

fixes: #102

### DIFF
--- a/pkg/logging/log.go
+++ b/pkg/logging/log.go
@@ -24,8 +24,11 @@ import (
 
 func Info(output string, args ...interface{}) {
 	blue := color.New(color.FgBlue).SprintFunc()
-	fmt.Print(blue("Info: "))
-	fmt.Printf(output, args...)
+	if output != "" {
+		fmt.Print(blue("Info: "))
+		fmt.Printf(output, args...)
+	}
+
 	fmt.Println()
 }
 
--- a/pkg/rook/status.go
+++ b/pkg/rook/status.go
@@ -36,11 +36,11 @@ func PrintCustomResourceStatus(clusterNamespace string, arg []string) {
 	if len(arg) == 1 && arg[0] == "all" {
 		command := fmt.Sprintf(getCrdList, clusterNamespace)
 		allCRs := strings.Split(exec.ExecuteBashCommand(command), "\n")
-		logging.Info(allCRs[0])
+		allCRs = allCRs[:len(allCRs)-1] // remove last empty line which is not a CR
 		for _, cr := range allCRs {
 			logging.Info(cr)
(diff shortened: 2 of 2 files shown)


pkg/rook/status.go
```diff
@@ -36,11 +36,12 @@ func PrintCustomResourceStatus(clusterNamespace string, arg []string) {
 	if len(arg) == 1 && arg[0] == "all" {
 		command := fmt.Sprintf(getCrdList, clusterNamespace)
 		allCRs := strings.Split(exec.ExecuteBashCommand(command), "\n")
-		logging.Info(allCRs[0])
+		allCRs = allCRs[:len(allCRs)-1]
 		for _, cr := range allCRs {
 			logging.Info(cr)
 			command := fmt.Sprintf(scriptPrintSpecificCRStatus, clusterNamespace, cr)
-			logging.Info(exec.ExecuteBashCommand(command))
```
travisn, line 43: The `logging.Info()` is printing "Info:" with no other info for other commands besides the example given in the issue. Instead of changing the implementation here, try just change inside the `logging.Info()` method to only print a blank line if there is no message.

travisn (changes requested):

pkg/rook/status.go
```diff
@@ -36,11 +36,12 @@ func PrintCustomResourceStatus(clusterNamespace string, arg []string) {
 	if len(arg) == 1 && arg[0] == "all" {
```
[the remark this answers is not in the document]
subhamkrai: got it

pkg/logging/log.go
```diff
@@ -24,8 +24,11 @@ import (
 
 func Info(output string, args ...interface{}) {
 	blue := color.New(color.FgBlue).SprintFunc()
-	fmt.Print(blue("Info: "))
-	fmt.Printf(output, args...)
+	if output != "" {
+		fmt.Print(blue("Info: "))
+		fmt.Printf(output, args...)
+	}
```
travisn, line 30: I think a blank line will still be helpful if it's blank
```suggestion
	} else {
		fmt.Println()
	}
```

travisn (changes requested):

pkg/logging/log.go
subhamkrai: there is already new line https://github.com/rook/kubectl-rook-ceph/blob/df1ab8b44787e1286eddeb4b46938edd43b58fb3/pkg/logging/log.go#L26-L32

travisn: oh right, I overlooked it, thanks

travisn (approved):

PR #112, opened 2023-05-24
Use recommended context.Context passing

BlaineEXE:
Documentation in context.Context states never to embed a context.Context in a struct. It should always be passed to functions and methods, ideally as the first parameter. Kubernetes' Go API follows this syntax, and since it is much easier to fix the issue in kubectl-rook-ceph compared to in rook/rook, we should do so.

The old k8sutil.Context has been renamed to k8sutil.Clientsets to reflect its new purpose.

### DIFF
--- a/cmd/commands/ceph.go
+++ b/cmd/commands/ceph.go
@@ -31,7 +31,7 @@ var CephCmd = &cobra.Command{
 	DisableFlagParsing: true,
 	Args:               cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		context := GetContext()
-		fmt.Println(exec.RunCommandInOperatorPod(context, cmd.Use, args, OperatorNamespace, CephClusterNamespace, true))
+		clientsets := GetClientsets()
+		fmt.Println(exec.RunCommandInOperatorPod(cmd.Context(), clientsets, cmd.Use, args, OperatorNamespace, CephClusterNamespace, true))
 	},
 }
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -53,10 +53,10 @@ func init() {
 	RootCmd.PersistentFlags().StringVar(&CephClusterNamespace, "namespace", "rook-ceph", "Kubernetes namespace where ceph cluster is created")
 }
 
-func GetContext() *k8sutil.Context {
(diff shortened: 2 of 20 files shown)


cmd/commands/root.go
```diff
@@ -53,38 +53,31 @@ func init() {
...
-	context.RookClientset, err = rookclient.NewForConfig(context.KubeConfig)
+	clientsets.Kube, err = k8s.NewForConfig(clientsets.KubeConfig)
 	if err != nil {
 		logging.Fatal(err)
 	}
 
-	context.Clientset, err = k8s.NewForConfig(context.KubeConfig)
-	if err != nil {
-		logging.Fatal(err)
-	}
```
BlaineEXE, line 85: This also fixes a duplicate `context.RookClientset, err = ...`

cmd/commands/ceph.go
```diff
@@ -31,7 +31,7 @@ var CephCmd = &cobra.Command{
 	DisableFlagParsing: true,
 	Args:               cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		context := GetContext()
-		fmt.Println(exec.RunCommandInOperatorPod(context, cmd.Use, args, OperatorNamespace, CephClusterNamespace, true))
+		clientsets := GetClientsets()
+		fmt.Println(exec.RunCommandInOperatorPod(cmd.Context(), clientsets, cmd.Use, args, OperatorNamespace, CephClusterNamespace, true))
```
BlaineEXE, line 35: One of the core reasons to make this change, versus letting it stand, is to make sure all commands explicitly use `cmd.Context()` for their context.

travisn: > Documentation in context.Context states never to embed a context.Context in a struct. It should always be passed to functions and methods, ideally as the first parameter.

We don't need to pass the context as a pointer? Seems like it would be similar behavior to pass it by value as to pass it around with a struct, but I didn't go dig into the whole issue.

subhamkrai (approved):
Overall looks good to me.

cmd/commands/ceph.go
BlaineEXE: context.Context is an interface, and I think it fails to compile if a `*` is in front of the param.

Issue #101, opened 2023-05-19
Operator namespace and cluster namespace parameters are not recognized in golang


travisn:
The `--operator-namespace` and `--namespace` flags are not recognized.

```
% bin/kubectl-rook-ceph -n foo health
Error: unknown command "-n" for "rook-ceph health"
```

```
% bin/kubectl-rook-ceph --operator-namespace foo health
Error: unknown command "--operator-namespace" for "rook-ceph health"
```

subhamkrai:
@BlaineEXE these flags are not working as expected https://github.com/rook/kubectl-rook-ceph/blob/04e4ffa5a031b00c507a5616e6c5a3b8318746b4/cmd/commands/root.go#L52

I tried using `StringVarP` also but no luck

travisn:
I wasn't able to repro this issue today. Building with the latest master of the krew plugin, I was able to run commands against other namespaces for both `--namespace` and `--operator-namespace`. Not sure what changed now. 

An improvement is that if the namespace doesn't exist, we should print an error message. Currently, it prints nothing if the namespace doesn't exist:
```
% bin/kubectl-rook-ceph --namespace=foo rook status
Info: 
```

PR #113, opened 2023-05-24
Fix parsing for namespace flags

BlaineEXE:
Namespace flags should be parsed by the parent command and not by child
    commands. This poses a particular problem for the 'ceph' and 'rbd'
    commands, as they normally have flag parsing disabled. The parent
    command should use the cobra.Command option 'TraverseChildren' in order
    to parse parent args before parsing child commands.
    
This is built on top of #112

### DIFF
--- a/cmd/commands/ceph.go
+++ b/cmd/commands/ceph.go
@@ -20,7 +20,7 @@ import (
 	"fmt"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
-
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


subhamkrai (changes requested):
one suggestion: 
later in this file could you modify the flags to 
```
RootCmd.PersistentFlags().StringVarP(&OperatorNamespace, "operator-namespace", "o","rook-ceph", "Kubernetes namespace where rook operator is running")
RootCmd.PersistentFlags().StringVarP(&CephClusterNamespace, "namespace", "n","rook-ceph", "Kubernetes namespace where ceph cluster is created")
```

and one question:
with these changes should we remove `DisableFlagParsing: true,` for other commands or we should have it like it is currently as I don't see any harm in keeping it. Thoughts?

BlaineEXE:
I would suggest not making a shorthand for --operator-namespace. `-o` is often used for `--output`, and assuming to keep that trend would probably be best. Most users won't have a different operator namespace, and so I think it's okay to not give it a shorthand.

travisn:
> I would suggest not making a shorthand for --operator-namespace. `-o` is often used for `--output`, and assuming to keep that trend would probably be best. Most users won't have a different operator namespace, and so I think it's okay to not give it a shorthand.

We may want to keep `-o` for compatibility with the existing script options.

subhamkrai:
@BlaineEXE could you look at above comment

subhamkrai (approved):
Based on offline discussion, approving.

subhamkrai:
fixes #101

Issue #98, opened 2023-05-19
Purging an OSD hangs with the golang implementation


travisn:
Purging the OSD hangs after the first line prints that says it's going to purge the OSD. 
This is in the case that the OSD pod is scaled down and is valid to remove. 
```
tnielsen@Traviss-MBP ~/src/go/src/github.com/rook/kubectl-rook-ceph % bin/kubectl-rook-ceph rook purge-osd 0
Info: Running purge osd command
^C
```

If the OSD is running, the purge is correctly aborted:
```
% bin/kubectl-rook-ceph rook purge-osd 1        
Info: Running purge osd command
2023-05-19 21:37:06.431260 W | cephcmd: loaded admin secret from env var ROOK_CEPH_SECRET instead of from file
2023-05-19 21:37:06.431382 I | rookcmd: starting Rook v1.11.0-alpha.0.331.gdb2402365-dirty with arguments 'rook ceph osd remove --osd-ids=1 --force-osd-removal=false'
2023-05-19 21:37:06.431712 I | rookcmd: flag values: --force-osd-removal=false, --help=false, --log-level=INFO, --osd-ids=1, --preserve-pvc=false
2023-05-19 21:37:06.431892 I | op-mon: parsing mon endpoints: e=10.107.26.9:6789,f=10.102.104.129:6789,g=10.106.3.131:6789
2023-05-19 21:37:06.436002 I | cephclient: writing config file /var/lib/rook/rook-ceph/rook-ceph.config
2023-05-19 21:37:06.436168 I | cephclient: generated admin config in /var/lib/rook/rook-ceph
2023-05-19 21:37:06.610550 I | cephosd: validating status of osd.1
2023-05-19 21:37:06.610574 I | cephosd: osd.1 is healthy. It cannot be removed unless it is 'down'
```

Using the script is able to purge the OSD without error:
```
tnielsen@Traviss-MBP ~/src/go/src/github.com/rook/kubectl-rook-ceph % ./kubectl-rook-ceph.sh rook purge-osd 0 --force
2023-05-19 21:34:54.055399 W | cephcmd: loaded admin secret from env var ROOK_CEPH_SECRET instead of from file
2023-05-19 21:34:54.055453 I | rookcmd: starting Rook v1.11.0-alpha.0.331.gdb2402365-dirty with arguments 'rook ceph osd remove --osd-ids=0 --force-osd-removal=true'
[...]

PR #115, opened 2023-05-31
core: fix osd hang during purge-osd

subhamkrai:
with earlier approach purging an OSD was hanging.
since, with bash implementation, it was passing so, this commit change the way osd removal command runs. Earlier, we were creating connection between the operator and then running the command now we are using os/exec package.

fixes: #98

### DIFF
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -36,60 +36,61 @@ var (
 	CephClusterNamespace string // Cephcluster namespace
 )
 
-func RunCommandInOperatorPod(ctx context.Context, clientsets *k8sutil.Clientsets, cmd string, args []string, operatorNamespace, clusterNamespace string, exitOnError bool) string {
+func RunCommandInOperatorPod(ctx context.Context, clientsets *k8sutil.Clientsets, cmd string, args []string, operatorNamespace, clusterNamespace string, returnOutput, exitOnError bool) string {
+	var pod v1.Pod
+	var err error
 
-	pod, err := k8sutil.WaitForPodToRun(ctx, clientsets.Kube, operatorNamespace, "app=rook-ceph-operator")
+	pod, err = k8sutil.WaitForPodToRun(ctx, clientsets.Kube, operatorNamespace, "app=rook-ceph-operator")
 	if err != nil {
 		logging.Fatal(fmt.Errorf("failed to wait for operator pod to run: %v", err))
 	}
 
 	var stdout, stderr bytes.Buffer
 
-	err = execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
-	if err != nil {
-		logging.Error(err)
-		if exitOnError {
-			os.Exit(1)
-		}
+	execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr, returnOutput, exitOnError)
+	if !returnOutput {
+		return ""
 	}
+	fmt.Println(stdout.String())
 
 	fmt.Print(stderr.String())
 	return stdout.String()
 }
 
-func RunCommandInToolboxPod(ctx context.Context, clientsets *k8sutil.Clientsets, cmd string, args []string, clusterNamespace string, exitOnError bool) string {
-	pod, err := k8sutil.WaitForPodToRun(ctx, clientsets.Kube, clusterNamespace, "app=rook-ceph-tools")
+func RunCommandInToolboxPod(ctx context.Context, clientsets *k8sutil.Clientsets, cmd string, args []string, clusterNamespace string, returnOutput, exitOnError bool) string {
+	var pod v1.Pod
+	var err error
+
+	pod, err = k8sutil.WaitForPodToRun(ctx, clientsets.Kube, clusterNamespace, "app=rook-ceph-tools")
 	if err != nil {
 		logging.Fatal(err)
 	}
 
 	var stdout, stderr bytes.Buffer
 
-	err = execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-tools", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
-	if err != nil {
-		logging.Error(err)
-		if exitOnError {
-			os.Exit(1)
-		}
+	execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-tools", pod.Namespace, clusterNamespace, args, &stdout, &stderr, returnOutput, exitOnError)
+	if !returnOutput {
+		return ""
 	}
+
 	fmt.Println(stderr.String())
 	return stdout.String()
 }
 
-func RunCommandInLabeledPod(ctx context.Context, clientsets *k8sutil.Clientsets, label, container, cmd string, args []string, clusterNamespace string, exitOnError bool) string {
+func RunCommandInLabeledPod(ctx context.Context, clientsets *k8sutil.Clientsets, label, container, cmd string, args []string, clusterNamespace string, returnOutput, exitOnError bool) string {
+	var list *v1.PodList
+	var err error
+
 	opts := metav1.ListOptions{LabelSelector: label}
-	list, err := clientsets.Kube.CoreV1().Pods(clusterNamespace).List(ctx, opts)
+	list, err = clientsets.Kube.CoreV1().Pods(clusterNamespace).List(ctx, opts)
 	if err != nil || len(list.Items) == 0 {
 		logging.Fatal(fmt.Errorf("failed to get rook mon pod where the command could be executed. %v", err))
 	}
 	var stdout, stderr bytes.Buffer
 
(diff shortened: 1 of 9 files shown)


pkg/rook/purge_osd.go
```diff
@@ -37,14 +39,27 @@ func PurgeOsd(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNames
...
+		"exec",
+		"deployments/rook-ceph-operator",
+		"--",
+		"/bin/bash",
 		"-c",
 		fmt.Sprintf("export ROOK_MON_ENDPOINTS=%s ROOK_CEPH_USERNAME=client.admin ROOK_CEPH_SECRET=%s ROOK_CONFIG_DIR=/var/lib/rook && rook ceph osd remove --osd-ids=%s --force-osd-removal=%s", monEndPoint, adminKey, osdId, flag),
 	}
 
 	logging.Info("Running purge osd command")
-	return exec.RunCommandInOperatorPod(ctx, clientsets, cmd, args, operatorNamespace, clusterNamespace, true)
```
travisn, line 49: So this is working when you run the command directly here, but not with the `RunCommandInOperatorPod()` helper? What's different? I assume this was an experiment, and now you can find a change in `RunCommandInOperatorPod()` that will work?

subhamkrai: Hmm, I'm still unable to find the fix. I tried a few things but didn't help.  The problem is it working as expected until we are calling osd deployment to 0 which is not giving me any hints what the could be issue.

@BlaineEXE could you help us here?

also, I checked the rook code, and I see there also we're using `os/exec` pkg instead of creating `stream`. The stream method is only used in integration tests when we are running commands in the toolbox.

@travisn will this be a good idea to move to `os/exec` if it takes more time?

travisn: > @travisn will this be a good idea to move to os/exec if it takes more time?

What do you mean it will take more time? It appears to me that `StreamWithContext()` is not working as expected since it doesn't stream the output from stdout. Was there a need for `StreamWithContext()`? Since the `os/exec` package is working for us, seems like we can implement it instead of the streaming method.

Are there any other commands that take a while to run in the operator, similar to the osd purge? Perhaps we didn't notice that the streaming wasn't working with other commands since they are all quick in their call to the operator.

subhamkrai: @travisn Okay, finally I could find the root issue, and the issue you opened in rook https://github.com/rook/rook/issues/12320 gave me a clue.

Let's discuss this in a call, meanwhile, I'll try to fix the issue soon

pkg/exec/exec.go
```diff
@@ -37,15 +37,22 @@ var (
...
-	pod, err := k8sutil.WaitForPodToRun(ctx, clientsets.Kube, operatorNamespace, "app=rook-ceph-operator")
+	pod, err = k8sutil.WaitForPodToRun(ctx, clientsets.Kube, operatorNamespace, "app=rook-ceph-operator")
 	if err != nil {
 		logging.Fatal(fmt.Errorf("failed to wait for operator pod to run: %v", err))
 	}
 
 	var stdout, stderr bytes.Buffer
 
-	err = execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
+	if exitOnError {
```
travisn, line 50: This condition seems backwards if I'm reading it correctly
```suggestion
	if !exitOnError {
```

travisn: Or maybe i'm just confused by what it means that the method is named "without return", and then line 52 returns.

pkg/exec/exec.go
```diff
@@ -96,8 +119,56 @@ func RunCommandInLabeledPod(ctx context.Context, clientsets *k8sutil.Clientsets,
 	return stdout.String()
 }
 
-// execCmdInPod exec command on specific pod and wait the command's output.
-func execCmdInPod(ctx context.Context, clientsets *k8sutil.Clientsets, command, podName, containerName, podNamespace, clusterNamespace string, args []string, stdout, stderr io.Writer) error {
+// execCmdInPodWithoutReturn exec command on specific pod and print the command's output directly on shel.
+func execCmdInPodWithoutReturn(ctx context.Context, clientsets *k8sutil.Clientsets, command, podName, containerName, podNamespace, clusterNamespace string, args []string) {
```
travisn, line 123: This method seems mostly duplicated of the original method. What if this method took the `existOnError` parameter? Then it could just change the behavior needed instead of duplicating the method. That might help with my confusion on the other comment as well.

travisn (changes requested):

pkg/exec/exec.go
subhamkrai: Merge both in one method and also added one small comment on what `exitOnError` condition doing.

pkg/exec/exec.go
```diff
@@ -37,67 +37,70 @@ var (
...
 	var stdout, stderr bytes.Buffer
 
-	err = execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
-	if err != nil {
-		logging.Error(err)
-		if exitOnError {
-			os.Exit(1)
-		}
+	execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr, exitOnError)
+	if exitOnError {
```
travisn, line 51: If exitOnError, why would we return here? If there was an error it already exited, right? And if there wasn't an error, shouldn't we still continue with line 55?

travisn: Using exitOnError for the stderr and stdout is confusing me. How about a new flag? `exitOnError` should really just be used for deciding whether to exit immediately when there is an error. Today these two flags might be the same value, but there may be a command in the future that will need different values for these flags.

subhamkrai: Actually, the name `exitOnError` was confusing. We don't need two variables, I have renamed that to `returnOutput` and the bool value passed, it should be clear now. Let's discuss it today if it is still not clear. Thanks for the reviews.

pkg/exec/exec.go
```diff
@@ -36,68 +36,78 @@ var (
...
 	var stdout, stderr bytes.Buffer
 
-	err = execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
-	if err != nil {
-		logging.Error(err)
-		if exitOnError {
-			os.Exit(1)
-		}
+	execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr, returnOutput)
+	// when `returnOuput is false, We are printing the command in the os.Output/os.Stderr and not storing in stdout/sderr declared above.
```
travisn, line 51:
```suggestion
	// when `returnOutput` is false, write the output to the os.Output/os.Stderr rather than in stdout/sderr declared above.
```

pkg/exec/exec.go
```diff
@@ -132,11 +142,29 @@ func execCmdInPod(ctx context.Context, clientsets *k8sutil.Clientsets, command,
...
+	if !returnOutput {
+		// Connect this process' std{in,out,err} to the remote shell process.
+		err = exec.StreamWithContext(ctx, remotecommand.StreamOptions{
+			Stdin:  os.Stdin,
+			Stdout: os.Stdout,
+			Stderr: os.Stderr,
+			Tty:    false,
+		})
+		if err != nil {
+			logging.Fatal(err)
```
travisn, line 155: Why is this one fatal, but line 167 is not? Seems like we either need the `exitOnError` or I need to understand better how it is tied to the `returnOutput`

pkg/exec/exec.go
```diff
@@ -36,68 +36,78 @@ var (
...
 
-	err = execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr)
-	if err != nil {
-		logging.Error(err)
-		if exitOnError {
-			os.Exit(1)
-		}
+	execCmdInPod(ctx, clientsets, cmd, pod.Name, "rook-ceph-operator", pod.Namespace, clusterNamespace, args, &stdout, &stderr, returnOutput)
+	// when `returnOuput is false, We are printing the command in the os.Output/os.Stderr and not storing in stdout/sderr declared above.
+	// So returning empty string and if `returnOutput` is true we'll return the command output.
```
travisn, line 52: I think no need for this comment since it's self explanatory with `returnOutput`
```suggestion
```

travisn (changes requested):

pkg/exec/exec.go
subhamkrai: oh that will overlook, we need to use `logging.Fatal` both places since if we are not able to create the stream to capture or print the output then we want to exit right there.

pkg/exec/exec.go
subhamkrai: removed as suggested below

subhamkrai:
> LGTM, but the CI failed, not sure if it's related.

it's related, let me check

travisn (approved):

Issue #110, opened 2023-05-24
Print a warning if an unsupported version of rook is detected


travisn:
If the user is running an unsupported version of rook, the user should be aware that it is unsupported.

Since the rook version can be detected from inside the operator pod, I would propose that every krew command that connects to the operator pod should print a warning if the rook version is unsupported. An unsupported version can be detected if "alpha" or "beta" is found in the `rook version` string.

PR #116, opened 2023-06-05
core: validate namespace before proceeding

subhamkrai:
core: validate namespace before proceeding
 
we should validate if namespaces exist or not
before running any commands.

core: throw a warning when an unsupported version of rook is used

this, commit throws warning when unsupported version of rook
is used for example if rook version contain alpha or beta,
it will throw warning.

fixes: https://github.com/rook/kubectl-rook-ceph/issues/110
    
### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -18,7 +18,9 @@ package command
 import (
 	"context"
 	"fmt"
+	"strings"
 
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	rookclient "github.com/rook/rook/pkg/client/clientset/versioned"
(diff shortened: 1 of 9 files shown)


cmd/commands/root.go
```diff
@@ -84,5 +90,23 @@ func GetClientsets() *k8sutil.Clientsets {
...
 }
+
+func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string) {
+	_, err := k8sclientset.Kube.CoreV1().Namespaces().Get(ctx, operatorNamespace, v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Operator namespace '%s' does not exist. %v", operatorNamespace, err))
+	}
+	_, err = k8sclientset.Kube.CoreV1().Namespaces().Get(ctx, cephClusterNamespace, v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("CephCluster namespace '%s' does not exist. %v", cephClusterNamespace, err))
```
travisn, line 105: Some of the commands may not require this to be fatal. For example, the rook version and updating an operator config setting doesn't require the ceph cluster to be running. And if the cluster namespace doesn't exist, the command will anyway fail later on something else. How about if we just log this error for the cluster namespace, but not exit? 
```suggestion
		logging.Error(fmt.Errorf("CephCluster namespace '%s' does not exist. %v", cephClusterNamespace, err))
```

travisn (changes requested):

cmd/commands/root.go
subhamkrai: agreed.

travisn (approved):

Issue #118, opened 2023-06-13
Version information is printed every time unexpectedly


travisn:
Every command seems to print the namespace and version info. This makes it difficult to find the output that I care about. For example, here is the output for the `operator set` command:

```
% bin/kubectl-rook-ceph operator set foo bar
Info: CephCluster namespace: "rook-ceph"
Info: Rook operator namespace: "rook-ceph"
rook: v1.11.0-alpha.0.410.g97bf52d76
go: go1.19.7

Warning: rook version 'rook: v1.11.0-alpha.0.410.g97bf52d76
go: go1.19.7
' is running a pre-release version of Rook.
Info: configmap/rook-ceph-operator-config patched
```

I would expect output to only include the warning and then the related output:
```
% bin/kubectl-rook-ceph operator set foo bar
Warning: rook version 'rook: v1.11.0-alpha.0.410.g97bf52d76' is running a pre-release version of Rook.

Info: configmap/rook-ceph-operator-config patched
```

travisn:
The version is fixed, but the namespace is still unexpected to print every time. How about making these debug level?
```
Info: CephCluster namespace: "rook-ceph"
Info: Rook operator namespace: "rook-ceph"
```

PR #121, opened 2023-06-14
core: make rook version more readable

subhamkrai:
core: make rook version more readable

during pre-validation check, rook version was
printed in multiple lines with golang version
which was making it difficult to read. Now, it
prints in one single like with the golang version.

core: print version info only once
there was extra `fmt.Print()` in method `RunCommandInOperatorPod`
due to which command output was printed two times.

fixes: #118

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -18,6 +18,7 @@ package command
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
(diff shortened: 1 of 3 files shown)


subhamkrai:
earlier 
```
Warning: rook version 'rook: v1.11.0-alpha.0.410.g97bf52d76
go: go1.19.7
' is running a pre-release version of Rook.
```

now
```
Info: CephCluster namespace: "rook-ceph"
Info: Rook operator namespace: "rook-ceph"
Warning: rook version 'rook: v1.11.0-alpha.0.420.gd9a17691c' is running a pre-release version of Rook.
```

cmd/commands/root.go
```diff
@@ -106,7 +107,11 @@ func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, o
 	}
 
 	rookVersion := exec.RunCommandInOperatorPod(ctx, k8sclientset, "rook", []string{"version"}, operatorNamespace, cephClusterNamespace, true, false)
+	re := regexp.MustCompile("(?m)[\r\n]+^.*go: go.*$") // remove the go version from the output
+	rookVersion = re.ReplaceAllString(rookVersion, "")
```
travisn, line 111: How about factoring out this string replacement to a small helper method, then add a unit test for it? Whenever there is a regular expression, a unit test is always helpful.

subhamkrai: added. 

I want to add unit tests for all the methods but seems like that will take more time to release and require my bandwidth. XD

travisn: Unit tests for now can avoid the high level commands since they are covered by the integration tests, but for small things like this, unit tests will be perfect.

subhamkrai: make sense

PR #124, opened 2023-06-15
core: comment unnecessary namespace for now

subhamkrai:
core: comment unnecessary namespace for now

fixes: #118

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -38,8 +38,8 @@ var RootCmd = &cobra.Command{
 	Args:             cobra.MinimumNArgs(1),
 	TraverseChildren: true,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		logging.Info("CephCluster namespace: %q", CephClusterNamespace)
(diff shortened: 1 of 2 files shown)


subhamkrai:
this is not the most elegant solution,  I would like to do something similar to rook, but for that, I'll need to change the whole `fmt.Print` to another logging method which I'll avoid until the golang release.

cmd/commands/root.go
```diff
@@ -38,8 +39,10 @@ var RootCmd = &cobra.Command{
 	Args:             cobra.MinimumNArgs(1),
 	TraverseChildren: true,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		logging.Info("CephCluster namespace: %q", CephClusterNamespace)
```
travisn, line 41: Before we create a new debug option, how about if we just comment out these two lines for now? Perhaps we don't even need a debug mode, or may use a different approach.

travisn (changes requested):

PR #127, opened 2023-06-16
clean: removing all the content related to bash

subhamkrai:
clean: removing all the content related to bash

removing bash script for tool and also removing the CI's
which are not useful anymore since we are moving to go.

nit: this commits cleans the go-related codes.

this commits clean the code like,
1. removing the unused var
2. formating the imports


travisn (approved):
Is it a happy or sad day when the bash script goes away? :)

Issue #129, opened 2023-06-21
--namespace does not work


lunderhage:
In v0.5.0, the --namespace argument is ignored and the plugin can not be used if the rook-ceph installation is not in the rook-ceph namespace.

lunderhage:
I'm using Redhat's ODF so the rook operator and ceph cluster are in the openshift-storage namespace.

subhamkrai:
> I'm using Redhat's ODF so the rook operator and ceph cluster are in the openshift-storage namespace.

@lunderhage okay, try this workaround and see if this works. until we have a fix
```kubectl rook-ceph --operator-namespace {namespace} ...```

subhamkrai:
@lunderhage with 0.5.1 this issue should be fixed. Try and let us know.

PR #130, opened 2023-06-22
namespace: namespace flags are not working

subhamkrai:
namespace: namespace flags are not working

cluster namespace and operator namespace needs
better handling. It was throwing an error and was exited
when namespace args are passed.
Now, the default cluster ns is rook-ceph, and for
operator it is empty but we check later in the code
that if CephClusterNamespace != "" && OperatorNamespace == ""
then assign the cluster ns to operator ns.

ci: add CI to test the plugin in different namespace

fixes: #129

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  go-test:
+  default-namespace:
     runs-on: ubuntu-20.04
     steps:
       - name: checkout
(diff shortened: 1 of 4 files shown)


travisn (approved):
Nice to see the tests in the non-default namespace, and the issues they found already!

travisn:
Maybe early next week we can plan on releasing v0.5.1, in case any other fixes are needed as well

subhamkrai:
> Maybe early next week we can plan on releasing v0.5.1, in case any other fixes are needed as well


yes, since the issue with then namespace is really annoying.

Issue #134, opened 2023-06-27
[v0.5.0] oidc error


HaveFun83:
After plugin upgrade from 0.4.0 to 0.5.0 i got the following error when i want to use the kubectl rook-ceph plugin
```
Error: no Auth Provider found for name "oidc"
```
Maybe someone can help here?

subhamkrai:
@HaveFun83 what command you were trying to run?

HaveFun83:
I tried: `kubectl rook-ceph ceph status`
Cluster: On-Prem kubeadm v1.26.5

I manually rolled back to plugin v0.4.0 and everything is working again.

HaveFun83:
We use keycloak as odic provider to authenticate our user (kubectl) against the kubernetes api.
https://kubernetes.io/docs/reference/access-authn-authz/authentication/#option-1-oidc-authenticator

HaveFun83:
@subhamkrai thank you so much. Works like a charm :rocket: :1st_place_medal:

subhamkrai:
> @subhamkrai thank you so much. Works like a charm 🚀 🥇

Good to know it is working 🙂

PR #135, opened 2023-06-27
add-auth: add auth plugin package

subhamkrai:
some clusters may have some kind of kubectl authenticator example 'OIDC', `gcb`, etc. So adding the auth pkg will help.

Fixes: #134


subhamkrai:
reading [this](https://github.com/kubernetes/client-go/issues/345#issuecomment-352139468) seems like importing auth pkg will fix it. I marked pr in the draft since I don't have much idea how we can test or if is testing needed.

travisn (approved):

Issue #136, opened 2023-06-30
--context flag no longer supported


taxilian:
The --context flag has been working in previous versions, but after the recent update (I think you changed to a golang version?) it no longer seems to be supported, which makes it a lot harder to use in my configuration

subhamkrai:
@taxilian seems like we missed adding this flag, created the pr to add it back, apologies.

taxilian:
I will of course hold a grudge for all of my days over the egregious oversight -- but it will probably be overshadowed by my appreciation of the tool, so I expect it'll even out in the end ;-) Let me know if I can help somehow.

subhamkrai:
since pr is already there, so just wait for the release let me know if that works :XD

subhamkrai:
@taxilian Test with the latest v0.5.2 version hopefully this should be fixed.

PR #137, opened 2023-07-03
flag: add context flag to switch between clusters

subhamkrai:
we missed adding `--context` flag when migrating
to golang. Adding the flag back.

fixes: #136

### DIFF
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -37,6 +37,7 @@ var (
 	KubeConfig           string
 	OperatorNamespace    string
 	CephClusterNamespace string
+	KubeContext          string
 )
 
 // rookCmd represents the rook command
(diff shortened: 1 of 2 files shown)


cmd/commands/root.go
```diff
@@ -37,6 +37,7 @@ var (
 	KubeConfig           string
 	OperatorNamespace    string
 	CephClusterNamespace string
+	Context              string
```
travisn, line 40: How about a more specific name?
```suggestion
	KubeContext          string
```

travisn (approved):
Just a small suggestion, feel free to merge and get a release out when this is updated.

cmd/commands/root.go
subhamkrai: done

PR #141, opened 2023-07-13
Redirect logs to stderr.

lunderhage:
Core: redirect log messages to stderr

The PR redirects the log messages to stderr so that the 
json output from some commands can be parsed with eg jq.

### DIFF
--- a/pkg/logging/log.go
+++ b/pkg/logging/log.go
@@ -26,31 +26,31 @@ import (
 func Info(output string, args ...interface{}) {
 	blue := color.New(color.FgBlue).SprintFunc()
 	if output != "" {
-		fmt.Print(blue("Info: "))
-		fmt.Printf(output, args...)
+		fmt.Fprintf(os.Stderr, blue("Info: "))
+		fmt.Fprintf(os.Stderr, output, args...)
 	}
 
-	fmt.Println()
+	fmt.Fprintf(os.Stderr, "\n")
 }
 
 func Warning(output string, args ...interface{}) {
 	yellow := color.New(color.FgYellow).SprintFunc()
-	fmt.Print(yellow("Warning: "))
-	fmt.Printf(output, args...)
-	fmt.Println()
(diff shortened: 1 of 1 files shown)


subhamkrai (changes requested):
please sign-off  your commits

https://github.com/rook/rook/blob/master/Documentation/Contributing/development-flow.md#commit-structure

subhamkrai:
@lunderhage can you update your pr commit message with something like

```
Core: redirect log messages to stderr

The PR redirects the log messages to stderr so that the 
json output from some commands can be parsed with eg jq.
```

Thanks

subhamkrai (commented):
PR looks good to me and also working locally but somehow CI is not happy.
```
srai@192 ~ (pr/141) $ ./bin/kubectl-rook-ceph rbd ls replicapool
Warning: rook version 'rook: v1.12.0-alpha.0.10.g647a4e353' is running a pre-release version of Rook.

csi-vol-45398f0b-276c-4e93-ab33-df24dfb9dff8
```

@travisn do you see any immediate issue with the CI or we can go ahead with and debug CI in the follow up pr?

travisn:
> @travisn do you see any immediate issue with the CI or we can go ahead with and debug CI in the follow up pr?

I don't immediately see the cause either. Does any other PR see this issue? Until we know it's not caused by this PR, let's troubleshoot before merging, otherwise the CI could be broken by this.

subhamkrai:
> > @travisn do you see any immediate issue with the CI or we can go ahead with and debug CI in the follow up pr?
> 
> I don't immediately see the cause either. Does any other PR see this issue? Until we know it's not caused by this PR, let's troubleshoot before merging, otherwise the CI could be broken by this.

#142 should fix the ci issue

subhamkrai:
@lunderhage please rebase your pr to get green CI. Thanks for the PR!

subhamkrai (approved):

PR #143, opened 2023-07-18
core: update golang version to 1.20

subhamkrai:
core: update golang version to 1.20

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: setup golang
         uses: actions/setup-go@v2
         with:
-          go-version: 1.19
+          go-version: "1.20"
 
       - name: setup cluster
         uses: ./.github/workflows/cluster-setup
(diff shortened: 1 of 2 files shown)


subhamkrai:
> Looks like there is an issue with the CI. Maybe a timeout is needed since they have been running for four hours.

yeah, we can add the meantime #142 will fix this CI issue once that is merged it will be green again and we can add timeout in follow-up PR.

.github/workflows/go-test.yaml
```diff
@@ -131,6 +131,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - name: setup golang
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.20"
```
BlaineEXE, line 137: Did you want to keep this step with the one above on purpose?

BlaineEXE (approved):
LGTM, just a question about spacing before merge.

.github/workflows/go-test.yaml
subhamkrai: no, space was missed added now

subhamkrai:
> LGTM, just a question about spacing before merge.

yes, added the space

PR #156, opened 2023-07-20
build: update client-go and other go mods

subhamkrai:
manually, updating a few go modules which are
causing error when dependabot is trying to update.

Closes:  #154


subhamkrai:
#160 will fix the build issue

Issue #68, opened 2022-10-21
Add Krew command for recovering the cephcluster CR after accidental deletion


travisn:
As seen in the Rook disaster recovery guide, [restoring the cephcluster CR after deletion](https://rook.io/docs/rook/latest/Troubleshooting/disaster-recovery/#restoring-crds-after-deletion) is a very tedious and accident-prone process. Similar to the command just added for `mons restore-quorum` in #61, we could add a command such as:

```
kubectl rook-ceph undelete cephcluster [name]
```

Where the `name` could be `rook-ceph` by default, which is also the default for the namespace, and is the default in our example deployments.

We could also have undelete commands for other CRs, which is why we might consider undelete as a top level command.

PR #170, opened 2023-08-30
restore: restore crd after accidental deletion

subhamkrai:
with these changes we'll be able to restore any
crd which stuck deletion due some dependencies
after after accidental deletion.
example:
kubectl rook-ceph -n <ns> restore <cr> <cr_name>

fixes: #68

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -15,6 +15,8 @@ concurrency:
 jobs:
   default-namespace:
     runs-on: ubuntu-20.04
+    env:
+      ROOK_PLUGIN_SKIP_PROMPTS: true
     steps:
       - name: checkout
         uses: actions/checkout@v4
@@ -53,8 +55,6 @@ jobs:
           kubectl rook-ceph --context=$(kubectl config current-context) ceph status
 
       - name: Mon restore
-        env:
-          ROOK_PLUGIN_SKIP_PROMPTS: true
         run: |
           set -ex
           # test the mon restore to restore to mon a, delete mons b and c, then add d and e
@@ -112,6 +112,22 @@ jobs:
           kubectl -n rook-ceph scale deployment rook-ceph-osd-0 --replicas 0
           kubectl rook-ceph rook purge-osd 0 --force
 
+      - name: Restore CRD without CRName
+        run: |
+          # First let's delete the cephCluster
+          kubectl -n rook-ceph delete cephcluster my-cluster --timeout 3s --wait=false
+
+          kubectl rook-ceph -n rook-ceph restore-deleted cephcluster
+          tests/github-action-helper.sh wait_for_crd_to_be_ready_default
+
+      - name: Restore CRD with CRName
+        run: |
+          # First let's delete the cephCluster
+          kubectl -n rook-ceph delete cephcluster my-cluster --timeout 3s --wait=false
+
+          kubectl rook-ceph -n rook-ceph restore-deleted cephcluster my-cluster
+          tests/github-action-helper.sh wait_for_crd_to_be_ready_default
+
       - name: collect common logs
         if: always()
         uses: ./.github/workflows/collect-logs
@@ -126,6 +142,8 @@ jobs:
 
   custom-namespace:
     runs-on: ubuntu-20.04
+    env:
+      ROOK_PLUGIN_SKIP_PROMPTS: true
     steps:
       - name: checkout
         uses: actions/checkout@v4
@@ -166,8 +184,6 @@ jobs:
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster --context=$(kubectl config current-context) ceph status
 
       - name: Mon restore
-        env:
-          ROOK_PLUGIN_SKIP_PROMPTS: true
         run: |
           set -ex
           # test the mon restore to restore to mon a, delete mons b and c, then add d and e
@@ -224,6 +240,22 @@ jobs:
           set -ex
           kubectl -n test-cluster scale deployment rook-ceph-osd-0 --replicas 0
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster rook purge-osd 0 --force
+
+      - name: Restore CRD without CRName
+        run: |
+          # First let's delete the cephCluster
+          kubectl -n test-cluster delete cephcluster my-cluster --timeout 3s --wait=false
+
+          kubectl rook-ceph --operator-namespace test-operator -n test-cluster restore-deleted cephcluster
+          tests/github-action-helper.sh wait_for_crd_to_be_ready_custom
+
+      - name: Restore CRD with CRName
+        run: |
+          # First let's delete the cephCluster
+          kubectl -n test-cluster delete cephcluster my-cluster --timeout 3s --wait=false
+
+          kubectl rook-ceph --operator-namespace test-operator -n test-cluster restore-deleted cephcluster my-cluster
+          tests/github-action-helper.sh wait_for_crd_to_be_ready_custom
 
       - name: collect common logs
         if: always()
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ These are args currently supported:
 - `dr` :
   - `health [ceph status args]`: Print the `ceph status` of a peer cluster in a mirroring-enabled environment thereby validating connectivity between ceph clusters. Ceph status args can be optionally passed, such as to change the log level: `--debug-ms 1`.
 
+- `restore-deleted <CRD> [CRName]`: Restore the ceph resources which are stuck in deleting state due to underlying resources being present in the cluster
+
 - `help` : Output help text
 
 ## Documentation
@@ -101,6 +103,7 @@ Visit docs below for complete details about each command and their flags uses.
 1. [Debug OSDs and Mons](docs/debug.md)
 1. [Restore mon quorum](docs/mons.md#restore-quorum)
 1. [Disaster Recovery](docs/dr-health.md)
+1. [Restore deleted CRs](docs/crd.md)
 
 ## Examples
 
--- a/cmd/commands/restore.go
+++ b/cmd/commands/restore.go
@@ -1,0 +1,34 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package command
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/restore"
+	"github.com/spf13/cobra"
+)
+
+// RestoreCmd represents the restore commands
+var RestoreCmd = &cobra.Command{
+	Use:   "restore-deleted",
+	Short: "Restores a CR that was accidentally deleted and is still in terminating state. Ex: restore cephcluster <my-cluster>",
+	Args:  cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		clientsets := GetClientsets(cmd.Context())
+		VerifyOperatorPodIsRunning(cmd.Context(), clientsets, OperatorNamespace, CephClusterNamespace)
+		restore.RestoreCrd(cmd.Context(), clientsets, OperatorNamespace, CephClusterNamespace, args)
+	},
+}
--- a/docs/crd.md
+++ b/docs/crd.md
@@ -1,0 +1,112 @@
+# Restoring Deleted CRs
+
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
+
+> [!NOTE]
+> If there are multiple deleted resources in the cluster and no specific resource is mentioned, the first resource will be restored. To restore all deleted resources, re-run the command multiple times.
+
+## restore-deleted Command
+
+The `restore-deleted` command has one required and one optional parameter:
+
+- `<CRD>`: The CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
+- `[CRName]`: The name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
+
+```bash
+kubectl rook-ceph restore-deleted <CRD> [CRName]
+```
+
+## CephCluster Restore Example
+
+```bash
+kubectl rook-ceph restore-deleted cephcluster
+
+Info: Detecting which resources to restore for crd "cephcluster"
+Info: Restoring CR my-cluster
+Warning: The resource my-cluster was found deleted. Do you want to restore it? yes | no
+
+Info: skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true
+Info: Scaling down the operator to 0
+Info: Backing up kubernetes and crd resources
+Info: Backed up crd cephcluster/my-cluster in file cephcluster-my-cluster.yaml
+Info: Deleting validating webhook rook-ceph-webhook if present
+Info: Fetching the UID for cephcluster/my-cluster
+Info: Successfully fetched uid 8366f79a-ae1f-4679-a62b-8abc6e1528fa from cephcluster/my-cluster
+Info: Removing ownerreferences from resources with matching uid 8366f79a-ae1f-4679-a62b-8abc6e1528fa
+Info: Removing owner references for secret cluster-peer-token-my-cluster
+Info: Removed ownerReference for Secret: cluster-peer-token-my-cluster
+
+Info: Removing owner references for secret rook-ceph-admin-keyring
+Info: Removed ownerReference for Secret: rook-ceph-admin-keyring
+
+Info: Removing owner references for secret rook-ceph-config
+Info: Removed ownerReference for Secret: rook-ceph-config
+
+Info: Removing owner references for secret rook-ceph-crash-collector-keyring
+Info: Removed ownerReference for Secret: rook-ceph-crash-collector-keyring
+
+Info: Removing owner references for secret rook-ceph-mgr-a-keyring
+Info: Removed ownerReference for Secret: rook-ceph-mgr-a-keyring
+
+Info: Removing owner references for secret rook-ceph-mons-keyring
+Info: Removed ownerReference for Secret: rook-ceph-mons-keyring
+
+Info: Removing owner references for secret rook-csi-cephfs-node
+Info: Removed ownerReference for Secret: rook-csi-cephfs-node
+
+Info: Removing owner references for secret rook-csi-cephfs-provisioner
+Info: Removed ownerReference for Secret: rook-csi-cephfs-provisioner
+
+Info: Removing owner references for secret rook-csi-rbd-node
+Info: Removed ownerReference for Secret: rook-csi-rbd-node
+
+Info: Removing owner references for secret rook-csi-rbd-provisioner
+Info: Removed ownerReference for Secret: rook-csi-rbd-provisioner
+
+Info: Removing owner references for configmaps rook-ceph-mon-endpoints
+Info: Removed ownerReference for configmap: rook-ceph-mon-endpoints
+
+Info: Removing owner references for service rook-ceph-exporter
+Info: Removed ownerReference for service: rook-ceph-exporter
+
+Info: Removing owner references for service rook-ceph-mgr
+Info: Removed ownerReference for service: rook-ceph-mgr
+
+Info: Removing owner references for service rook-ceph-mgr-dashboard
+Info: Removed ownerReference for service: rook-ceph-mgr-dashboard
+
+Info: Removing owner references for service rook-ceph-mon-a
+Info: Removed ownerReference for service: rook-ceph-mon-a
+
+Info: Removing owner references for service rook-ceph-mon-d
+Info: Removed ownerReference for service: rook-ceph-mon-d
+
+Info: Removing owner references for service rook-ceph-mon-e
+Info: Removed ownerReference for service: rook-ceph-mon-e
+
+Info: Removing owner references for deployemt rook-ceph-mgr-a
+Info: Removed ownerReference for deployment: rook-ceph-mgr-a
+
+Info: Removing owner references for deployemt rook-ceph-mon-a
+Info: Removed ownerReference for deployment: rook-ceph-mon-a
+
+Info: Removing owner references for deployemt rook-ceph-mon-d
+Info: Removed ownerReference for deployment: rook-ceph-mon-d
+
+Info: Removing owner references for deployemt rook-ceph-mon-e
+Info: Removed ownerReference for deployment: rook-ceph-mon-e
+
+Info: Removing owner references for deployemt rook-ceph-osd-0
+Info: Removed ownerReference for deployment: rook-ceph-osd-0
+
+Info: Removing finalizers from cephcluster/my-cluster
+Info: cephcluster.ceph.rook.io/my-cluster patched
+
+Info: Re-creating the CR cephcluster from file cephcluster-my-cluster.yaml created above
+Info: cephcluster.ceph.rook.io/my-cluster created
+
+Info: Scaling up the operator to 1
+Info: CR is successfully restored. Please watch the operator logs and check the crd
+```
--- a/pkg/mons/restore_quorum.go
+++ b/pkg/mons/restore_quorum.go
@@ -71,7 +71,7 @@ func restoreQuorum(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 	logging.Info("printing fsid secret %s\n", cephFsid)
 	logging.Info("Check for the running toolbox")
 
-	_, err = debug.GetDeployment(ctx, clientsets.Kube, clusterNamespace, "rook-ceph-tools")
+	_, err = k8sutil.GetDeployment(ctx, clientsets.Kube, clusterNamespace, "rook-ceph-tools")
 	if err != nil {
 		return fmt.Errorf("failed to deployment rook-ceph-tools. %v", err)
 	}
@@ -85,17 +85,17 @@ func restoreQuorum(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 	logging.Info("The mons to discard are: %s\n", badMons)
 	logging.Info("The cluster fsid is %s\n", cephFsid)
 
-	var answer, output string
+	var answer string
 	logging.Warning("Are you sure you want to restore the quorum to mon %s? If so, enter 'yes-really-restore'\n", goodMon)
 	fmt.Scanf("%s", &answer)
-	output, err = promptToContinueOrCancel(answer)
+	err = PromptToContinueOrCancel("restore-quorum", "yes-really-restore", answer)
 	if err != nil {
 		return fmt.Errorf("restoring the mon quorum to mon %s cancelled", goodMon)
 	}
-	logging.Info(output)
+	logging.Info("proceeding with resorting quorum")
 
 	logging.Info("Waiting for operator pod to stop")
-	err = debug.SetDeploymentScale(ctx, clientsets.Kube, operatorNamespace, "rook-ceph-operator", 0)
+	err = k8sutil.SetDeploymentScale(ctx, clientsets.Kube, operatorNamespace, "rook-ceph-operator", 0)
 	if err != nil {
 		return fmt.Errorf("failed to stop deployment rook-ceph-operator. %v", err)
 	}
@@ -103,7 +103,7 @@ func restoreQuorum(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 
 	logging.Info("Waiting for bad mon pod to stop")
 	for _, badMon := range badMons {
-		err = debug.SetDeploymentScale(ctx, clientsets.Kube, clusterNamespace, fmt.Sprintf("rook-ceph-mon-%s", badMon), 0)
+		err = k8sutil.SetDeploymentScale(ctx, clientsets.Kube, clusterNamespace, fmt.Sprintf("rook-ceph-mon-%s", badMon), 0)
 		if err != nil {
 			return fmt.Errorf("deployment %s still exist. %v", fmt.Sprintf("rook-ceph-mon-%s", badMon), err)
 		}
@@ -112,7 +112,7 @@ func restoreQuorum(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 
 	debug.StartDebug(ctx, clientsets.Kube, clusterNamespace, fmt.Sprintf("rook-ceph-mon-%s", goodMon), "")
 
-	debugDeploymentSpec, err := debug.GetDeployment(ctx, clientsets.Kube, clusterNamespace, fmt.Sprintf("rook-ceph-mon-%s-debug", goodMon))
+	debugDeploymentSpec, err := k8sutil.GetDeployment(ctx, clientsets.Kube, clusterNamespace, fmt.Sprintf("rook-ceph-mon-%s-debug", goodMon))
 	if err != nil {
 		return fmt.Errorf("failed to deployment rook-ceph-mon-%s-debug", goodMon)
 	}
@@ -150,13 +150,14 @@ func restoreQuorum(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 	logging.Info("Mon quorum was successfully restored to mon %s\n", goodMon)
 	logging.Info("Only a single mon is currently running")
 	logging.Info("Press Enter to start the operator and expand to full mon quorum again")
-	output, err = promptToContinueOrCancel(answer)
+
+	err = PromptToContinueOrCancel("restore-quorum", "yes-really-restore", answer)
 	if err != nil {
 		return fmt.Errorf("skipping operator start to expand full mon quorum.")
 	}
-	logging.Info(output)
+	logging.Info("proceeding with resorting quorum")
 
-	err = debug.SetDeploymentScale(ctx, clientsets.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	err = k8sutil.SetDeploymentScale(ctx, clientsets.Kube, operatorNamespace, "rook-ceph-operator", 1)
 	if err != nil {
 		return fmt.Errorf("failed to start deployment rook-ceph-operator. %v", err)
 	}
@@ -278,14 +279,15 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 	return badMons, goodMonPublicIp, goodMonPort, nil
 }
 
-func promptToContinueOrCancel(answer string) (string, error) {
+func PromptToContinueOrCancel(inputSource, expectedAnswer, answer string) error {
 	if skip, ok := os.LookupEnv("ROOK_PLUGIN_SKIP_PROMPTS"); ok && skip == "true" {
-		return "skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true", nil
+		logging.Info("skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true")
+		return nil
 	}
 
-	if answer == "yes-really-restore" {
-		return "proceeding", nil
+	if answer == expectedAnswer {
+		return nil
 	}
 
-	return "", fmt.Errorf("cancelled")
+	return fmt.Errorf("cancelled")
 }
--- a/pkg/restore/crd.go
+++ b/pkg/restore/crd.go
@@ -1,0 +1,229 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package restore
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/rook/kubectl-rook-ceph/pkg/mons"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+	crd := args[0]
+	logging.Info("Detecting which resources to restore for crd %q", crd)
+	getCrName := `kubectl -n %s get %s -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.deletionGracePeriodSeconds}{"\n"}{end}' | awk '$2=="0" {print $1}' | head -n 1`
+	command := fmt.Sprintf(getCrName, clusterNamespace, crd)
+	crName := strings.TrimSpace(exec.ExecuteBashCommand(command))
+
+	if crName == "" {
+		logging.Info("Nothing to do here, no %q resources in deleted state", crd)
+		return
+	}
+
+	if len(args) == 2 {
+		crName = args[1]
+	}
+
+	logging.Info("Restoring CR %s", crName)
+	var answer string
+	logging.Warning("The resource %s was found deleted. Do you want to restore it? yes | no\n", crName)
+	fmt.Scanf("%s", &answer)
+	err := mons.PromptToContinueOrCancel("restore-deleted", "yes", answer)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Restoring the resource %s cancelled", crName))
+	}
+	logging.Info("Proceeding with restoring deleting CR")
+
+	logging.Info("Scaling down the operator")
+	err = k8sutil.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 0)
+	if err != nil {
+		deploy, er := k8sutil.GetDeployment(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator")
+		if er != nil && !apierrors.IsNotFound(er) {
+			logging.Fatal(er)
+		}
+		if deploy.Status.ReadyReplicas != 0 {
+			logging.Fatal(fmt.Errorf("Failed to scale down the operator deployment. %v", err))
+		}
+	}
+
+	logging.Info("Backing up kubernetes and crd resources")
+	crFileName := crd + "-" + crName + ".yaml"
+	getCrYamlContent := `kubectl -n %s get %s %s -oyaml > %s`
+	command = fmt.Sprintf(getCrYamlContent, clusterNamespace, crd, crName, crFileName)
+	exec.ExecuteBashCommand(command)
+	logging.Info("Backed up crd %s/%s in file %s", crd, crName, crFileName)
+
+	webhookConfigName := "rook-ceph-webhook"
+	logging.Info("Deleting validating webhook %s if present", webhookConfigName)
+	err = k8sclientset.Kube.AdmissionregistrationV1().ValidatingWebhookConfigurations().Delete(ctx, webhookConfigName, v1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		logging.Fatal(fmt.Errorf("failed to delete validating webhook %s. %v", webhookConfigName, err))
+	}
+
+	logging.Info("Fetching the UID for %s/%s", crd, crName)
+	getCrUID := `kubectl -n %s get %s %s -o 'jsonpath={.metadata.uid}'`
+	command = fmt.Sprintf(getCrUID, clusterNamespace, crd, crName)
+	uid := exec.ExecuteBashCommand(command)
+	logging.Info("Successfully fetched uid %s from %s/%s", uid, crd, crName)
+
+	removeOwnerRefOfUID(ctx, k8sclientset, operatorNamespace, clusterNamespace, uid)
+
+	logging.Info("Removing finalizers from %s/%s", crd, crName)
+	removeFinalizers := `kubectl -n %s patch %s/%s --type json --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'`
+	command = fmt.Sprintf(removeFinalizers, clusterNamespace, crd, crName)
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Re-creating the CR %s from file %s created above", crd, crFileName)
+	recreateCR := `kubectl create -f %s`
+	command = fmt.Sprintf(recreateCR, crFileName)
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Scaling up the operator")
+	err = k8sutil.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Operator pod still being scaled up"))
+	}
+
+	if err = os.Remove(crFileName); err != nil {
+		logging.Warning("Unable to remove. Please remove the file %s manually.%v", crFileName, err)
+	}
+
+	logging.Info("CR is successfully restored. Please watch the operator logs and check the crd")
+}
+
+func removeOwnerRefOfUID(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace, targetUID string) {
+	logging.Info("Removing ownerreferences from resources with matching uid %s", targetUID)
+
+	secrets, err := k8sclientset.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Failed to list secrets"))
+	}
+
+	for _, secret := range secrets.Items {
+		for _, owner := range secret.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("Removing owner references for secret %s", secret.Name)
+				// Remove ownerReferences.
+				secret.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.CoreV1().Secrets(clusterNamespace).Update(ctx, &secret, v1.UpdateOptions{})
+				if err != nil {
+					logging.Fatal(errors.Wrapf(err, "Failed to update ownerReferences for secret %s", secret.Name))
+				}
+				logging.Info("Removed ownerReference for Secret: %s\n", secret.Name)
+			}
+		}
+	}
+
+	cms, err := k8sclientset.Kube.CoreV1().ConfigMaps(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Failed to list configMaps"))
+	}
+
+	for _, cm := range cms.Items {
+		for _, owner := range cm.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("Removing owner references for configmaps %s", cm.Name)
+				// Remove ownerReferences.
+				cm.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.CoreV1().ConfigMaps(clusterNamespace).Update(ctx, &cm, v1.UpdateOptions{})
+				if err != nil {
+					logging.Fatal(errors.Wrapf(err, "Failed to update ownerReferences for configmaps %s", cm.Name))
+				}
+				logging.Info("Removed ownerReference for configmap: %s\n", cm.Name)
+			}
+		}
+	}
+
+	services, err := k8sclientset.Kube.CoreV1().Services(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Failed to list services"))
+	}
+
+	for _, service := range services.Items {
+		for _, owner := range service.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("Removing owner references for service %s", service.Name)
+				// Remove ownerReferences.
+				service.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.CoreV1().Services(clusterNamespace).Update(ctx, &service, v1.UpdateOptions{})
+				if err != nil {
+					logging.Fatal(errors.Wrapf(err, "Failed to update ownerReferences for service %s", service.Name))
+				}
+				logging.Info("Removed ownerReference for service: %s\n", service.Name)
+			}
+		}
+	}
+
+	deploys, err := k8sclientset.Kube.AppsV1().Deployments(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Failed to list deployments"))
+	}
+
+	for _, deploy := range deploys.Items {
+		for _, owner := range deploy.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("Removing owner references for deployment %s", deploy.Name)
+				// Remove ownerReferences.
+				deploy.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.AppsV1().Deployments(clusterNamespace).Update(ctx, &deploy, v1.UpdateOptions{})
+				if err != nil {
+					logging.Fatal(errors.Wrapf(err, "Failed to update ownerReferences for deployemt %s", deploy.Name))
+				}
+				logging.Info("Removed ownerReference for deployment: %s\n", deploy.Name)
+			}
+		}
+	}
+
+	pvcs, err := k8sclientset.Kube.CoreV1().PersistentVolumeClaims(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Failed to list pvc"))
+	}
+
+	for _, pvc := range pvcs.Items {
+		for _, owner := range pvc.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("Removing owner references for pvc %s", pvc.Name)
+				// Remove ownerReferences.
+				pvc.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.CoreV1().PersistentVolumeClaims(clusterNamespace).Update(ctx, &pvc, v1.UpdateOptions{})
+				if err != nil {
+					logging.Fatal(errors.Wrapf(err, "Failed to update ownerReferences for pvc %s", pvc.Name))
+				}
+				logging.Info("Removed ownerReference for pvc: %s\n", pvc.Name)
+			}
+		}
+	}
+}
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -38,5 +38,6 @@ func addcommands() {
 		command.DebugCmd,
 		command.Health,
 		command.DrCmd,
+		command.RestoreCmd,
 	)
 }
--- a/pkg/debug/start_debug.go
+++ b/pkg/debug/start_debug.go
@@ -25,7 +25,6 @@ import (
 	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 
 	appsv1 "k8s.io/api/apps/v1"
-	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
@@ -39,7 +38,7 @@ func StartDebug(ctx context.Context, k8sclientset kubernetes.Interface, clusterN
 }
 
 func startDebug(ctx context.Context, k8sclientset kubernetes.Interface, clusterNamespace, deploymentName, alternateImageValue string) error {
-	originalDeployment, err := GetDeployment(ctx, k8sclientset, clusterNamespace, deploymentName)
+	originalDeployment, err := k8sutil.GetDeployment(ctx, k8sclientset, clusterNamespace, deploymentName)
 	if err != nil {
 		return fmt.Errorf("Missing mon or osd deployment name %s. %v\n", deploymentName, err)
 	}
@@ -69,7 +68,7 @@ func startDebug(ctx context.Context, k8sclientset kubernetes.Interface, clusterN
 		return err
 	}
 
-	if err := SetDeploymentScale(ctx, k8sclientset, clusterNamespace, deployment.Name, 0); err != nil {
+	if err := k8sutil.SetDeploymentScale(ctx, k8sclientset, clusterNamespace, deployment.Name, 0); err != nil {
 		return err
 	}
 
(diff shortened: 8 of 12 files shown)


cmd/commands/restore.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/restore"
+	"github.com/spf13/cobra"
+)
+
+// RestoreCmd represents the restore commands
+var RestoreCmd = &cobra.Command{
+	Use:   "restore",
+	Short: "Reads the crd and crd name to restore. Ex: restore cephcluster my-cluster",
```
travisn, line 27: How about if the CR name is optional? If they only provide the CRD type name, then we can automatically restore all instances that have been marked for deletion.

subhamkrai: Before making changes, I think thinking is this a valid condition to support since for deletion of a resource we need to pass the name of the resource.

Thoughts?

subhamkrai:
Also, I need help getting the right error message rather than just the error code from the below code

https://github.com/rook/kubectl-rook-ceph/blob/e9dffafad86f8f7ed74337bf0c0f840c59e7344f/pkg/exec/bash.go#L25-L38

for example: when running a kubectl command to create a resource that already exits, the above code throughs `exit code 1` and then exits but what I really want is to print the Kubernetes error that `Already exits ......` and then exit.
 
I tried a few things but failed. any suggestion? 
@travisn @BlaineEXE 

Note, I can use `os.stdout, os.stderr`  but want to return the output from the method and not just log in to the terminal.

cmd/commands/restore.go
subhamkrai: > To clarify my suggestion, there are two cases:
> 
> 1. If they pass the (optional) resource name, then check if that resource is marked for deletion, and then restore it.
> 2. If they don't pass the resource name, query for all resources of that type (the name would also be queried in this step), then restore the instances that are marked for deletion.

I'm thinking of option 2, how to get the specific rook resource like cephcluster, cephfilesystem... and use rook API to fetch them with the right call.

> 
> For CephCluster, there is most commonly only a single resource anyway, but for pools or other resource types perhaps there could be multiple accidentally marked for deletion.

travisn: I'm thinking it may be better not to use the Rook API since then the CRD settings would be tied to the version of the API that this plugin is referencing. For this scenario, we need to make a copy of the CR, then delete it, then create it again. So we need to make sure all settings are preserved, independent of the Rook version. So if we can use more generic queries for the CRs it will be better.

travisn (commented):
> Also, I need help getting the right error message rather than just the error code from the below code
> 
> https://github.com/rook/kubectl-rook-ceph/blob/e9dffafad86f8f7ed74337bf0c0f840c59e7344f/pkg/exec/bash.go#L25-L38
> 
> for example: when running a kubectl command to create a resource that already exits, the above code throughs `exit code 1` and then exits but what I really want is to print the Kubernetes error that `Already exits ......` and then exit.
> 
> I tried a few things but failed. any suggestion? @travisn @BlaineEXE
> 
> Note, I can use `os.stdout, os.stderr` but want to return the output from the method and not just log in to the terminal.

Do you see the "Already exists" error code in the stderr? Do we just need to look for that string since no error code is returned?

subhamkrai:
Testing result

### Test Passing argument cr and crName
```yaml
 ./bin/kubectl-rook-ceph restore cephblockpool replicapool
Warning: rook version 'rook: v1.12.0-alpha.0.227.g3e248bf9a' is running a pre-release version of Rook.

Info: Restoring crd cephblockpool name replicapool
Info: Scaling down the operator to 0
Info: Backing up kubernetes and crd resources
Info: Backed up crd cephblockpool named replicapool in file cephblockpool.yaml
Info: Backed up secret in file secret.yaml
Info: Backed up configmap in file configmap.yaml
Info: deleting validating webhook rook-ceph-webhook if present
Info: Fetching cephblockpool replicapool uid
Info: Successfully fetched uid 0f8b59df-816d-4cc8-b888-ec128b14c308 from cephblockpool replicapool
Info: removing ownerreferences from resources with matching uid 0f8b59df-816d-4cc8-b888-ec128b14c308
Info: removing finalizers from cephblockpool replicapool
Info: cephblockpool.ceph.rook.io/replicapool patched

Info: re-creating the cr cephblockpool from file replicapool.yaml created above
Info: cephblockpool.ceph.rook.io/replicapool created

Info: Scaling up the operator to 1
Info: CR should be successfully restored. Please watch the operator logs and check the crd
~/go/src/github.com/kubectl-rook-ceph
~/go/src/github.com/kubectl-rook-ceph
srai@fedora ~ (restore-crd) $ kc get cephblockpool
NAME          PHASE
replicapool   Progressing
~/go/src/github.com/kubectl-rook-ceph
srai@fedora ~ (restore-crd) $ kc get cephblockpool
NAME          PHASE
replicapool   Ready
```

### Test passing only cephcluster

```yaml
./bin/kubectl-rook-ceph  restore cephcluster
Warning: rook version 'rook: v1.12.0-alpha.0.227.g3e248bf9a' is running a pre-release version of Rook.

Info: Restoring crd cephcluster name my-cluster
Info: Scaling down the operator to 0
Info: Backing up kubernetes and crd resources
Info: Backed up crd cephcluster named my-cluster in file cephcluster.yaml
Info: Backed up secret in file secret.yaml
Info: Backed up configmap in file configmap.yaml
Info: deleting validating webhook rook-ceph-webhook if present
Info: Fetching cephcluster my-cluster uid
Info: Successfully fetched uid a20a0b18-52d5-4844-a542-782e2d224f98 from cephcluster my-cluster
Info: removing ownerreferences from resources with matching uid a20a0b18-52d5-4844-a542-782e2d224f98
Info: removing owner references for service rook-ceph-exporter
Info: Removed ownerReference for service: rook-ceph-exporter

Info: removing finalizers from cephcluster my-cluster
Info: cephcluster.ceph.rook.io/my-cluster patched

Info: re-creating the cr cephcluster from file my-cluster.yaml created above
Info: cephcluster.ceph.rook.io/my-cluster created

Info: Scaling up the operator to 1
Info: CR should be successfully restored. Please watch the operator logs and check the crd
~/go/src/github.com/kubectl-rook-ceph
srai@fedora ~ (restore-crd) $ kc get cephcluster
NAME         DATADIRHOSTPATH   MONCOUNT   AGE     PHASE   MESSAGE                        HEALTH      EXTERNAL   FSID
my-cluster   /var/lib/rook     1          7m44s   Ready   Cluster created successfully   HEALTH_OK              5dc8aca4-cb12-47ee-a83e-a9f773b39a0a
```

subhamkrai:
> > Also, I need help getting the right error message rather than just the error code from the below code
> > https://github.com/rook/kubectl-rook-ceph/blob/e9dffafad86f8f7ed74337bf0c0f840c59e7344f/pkg/exec/bash.go#L25-L38
> > 
> > for example: when running a kubectl command to create a resource that already exits, the above code throughs `exit code 1` and then exits but what I really want is to print the Kubernetes error that `Already exits ......` and then exit.
> > I tried a few things but failed. any suggestion? @travisn @BlaineEXE
> > Note, I can use `os.stdout, os.stderr` but want to return the output from the method and not just log in to the terminal.
> 
> Do you see the "Already exists" error code in the stderr? Do we just need to look for that string since no error code is returned?

yes, In the error stderr I saw the right error string but after checking the error from cmd. output it has error code. 
Anyway, everything is working as expected now. I had to add `cmd.Stderr = os.Stderr` and had to remove few args when running command directly using `os/exec`

subhamkrai:
@travisn I was thinking of adding CI test for this as another pr or do you think I should include it in this pr only?

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
+/*
```
travisn, line 1: Since it's already in the restore pkg, you could name it crd.go.

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/debug"
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var getCrName = `
+	kubectl -n %s get %s -o json | jq -r '.items[] | select(.metadata.deletionGracePeriodSeconds == 0) | .metadata.name'
```
travisn, line 33: Does the plugin know if it's running in the context of `oc` or `kubectl`? We would need to use the same tool.

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		logging.Fatal(err)
+	}
+
+	logging.Info("Backing up kubernetes and crd resources")
+	command = fmt.Sprintf(getCrYamlContent, clusterNamespace, cr, crName, cr)
+	exec.ExecuteBashCommand(command)
+	logging.Info("Backed up crd %s named %s in file %s.yaml", cr, crName, cr)
+
+	command = fmt.Sprintf(getK8sYamlContent, clusterNamespace, "secret", "secret.yaml")
```
travisn, line 78: This is backing up all secrets?

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+	exec.ExecuteBashCommand(command)
+	logging.Info("Backed up secret in file secret.yaml")
+
+	command = fmt.Sprintf(getK8sYamlContent, clusterNamespace, "configmap", "configmap.yaml")
+	exec.ExecuteBashCommand(command)
+	logging.Info("Backed up configmap in file configmap.yaml")
```
travisn, line 84: If we have full confidence on the automation to restore the deleted CRs, I don't think we need to backup the configmaps or secrets. The important configmaps and secrets anyway should have their own finalizers to prevent them from being deleted. The important thing will be to remove the ownerReferences from all resources so they won't be deleted when the parent resource is deleted.

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+	command = fmt.Sprintf(recreateCr, cr+".yaml")
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Scaling up the operator to 1")
+	err = debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
+	}
+
+	logging.Info("CR should be successfully restored. Please watch the operator logs and check the crd")
```
travisn, line 114: Let's have full confidence with testing that it is restored.
```suggestion
	logging.Info("CR is successfully restored. Please watch the operator logs and check the crd")
```

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+				if err != nil {
+					scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
+				}
+				logging.Info("Removed ownerReference for pvc: %s\n", pvc.Name)
+			}
+		}
+	}
+}
+
+func scaleBackOperatorOnError(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace string, errGot error) {
```
travisn, line 231: Do we have helpers in the plugin already for scaling the operator instead of creating this new func?

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+	logging.Info("Scaling up the operator to 1")
+	err = debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
+	}
+
+	logging.Info("CR should be successfully restored. Please watch the operator logs and check the crd")
+}
+
+func getResourceWithMatchingUid(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace, targetUID string) {
```
travisn, line 117: This method is about removing owner references, right? How about a different method name?

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+	logging.Info("removing ownerreferences from resources with matching uid %s", targetUID)
+
+	secrets, err := k8sclientset.Kube.CoreV1().Secrets(clusterNamespace).List(context.TODO(), v1.ListOptions{})
+	if err != nil {
+		scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
```
travisn, line 122: This method should return the `err`, and the caller can decide how to handle the error.

travisn: If there is a failure, perhaps no need to scale back up the operator.

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+
+	if len(args) == 2 {
+		crName = args[1]
+	}
+
+	logging.Info("Restoring crd %s name %s", cr, crName)
+
+	logging.Info("Scaling down the operator to 0")
+	err := debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 0)
+	if err != nil {
```
travisn, line 69: If the operator is already scaled down, will this fail?

travisn (changes requested):
> @travisn I was thinking of adding CI test for this as another pr or do you think I should include it in this pr only?

Let's have the CI test with this PR. It's critical that we test this feature fully before anyone tries to use it.

pkg/restore/restore_crd.go
subhamkrai: currently, the tool is doesn't have the capability of finding if it `kubectl` or `oc`. But I think it will be better if in the future if the tool is able to find the differences.

pkg/restore/restore_crd.go
subhamkrai: yes

pkg/restore/restore_crd.go
subhamkrai: I have confidence in the too to restore based on my testing yet and also it's better to remove the secret/configmap backup as you mentioned these already have the finalizers

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+		scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
+	}
+
+	logging.Info("CR should be successfully restored. Please watch the operator logs and check the crd")
+}
+
+func getResourceWithMatchingUid(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace, targetUID string) {
+	logging.Info("removing ownerreferences from resources with matching uid %s", targetUID)
+
+	secrets, err := k8sclientset.Kube.CoreV1().Secrets(clusterNamespace).List(context.TODO(), v1.ListOptions{})
```
Madhu-1, line 120: use the context already passed as function agrument

pkg/restore/restore_crd.go
Madhu-1: if required, we can do this in defer function in starting of this function if there is an error?

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+		for _, owner := range secret.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("removing owner references for secret %s", secret.Name)
+				// Remove ownerReferences.
+				secret.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.CoreV1().Secrets(clusterNamespace).Update(context.TODO(), &secret, v1.UpdateOptions{})
+				if err != nil {
+					scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
```
Madhu-1, line 135: log the error in case of failure in all the places in this function?

pkg/restore/restore_crd.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+	}
+}
+
+func scaleBackOperatorOnError(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace string, errGot error) {
+	err := debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		logging.Fatal(err)
+	}
+	logging.Fatal(errGot)
```
Madhu-1, line 236: IMO i think it doesn't sounds good to log this error here, Just for logging purposes we should not pass it as an argument. you will miss the actual error if scaling up deployment fails

pkg/restore/restore_crd.go
subhamkrai: We do have a helper method to scale the deployments but in command, we first scale the rook operator to 0 and scale back the operator in the end once all the operations are done but if an error occurs between scale 0 to scale 1 the rook operator will be down. So, what I'm doing here is, if any error occurs after scaling the rook operator pod 0 and before scaling back to 1, we'll call this new function where we'll first scale the rook operator back and then return the err.

pkg/restore/restore_crd.go
subhamkrai: so, let's say if we get an error we don't want to scale the rook operator first, directly return the error and the user will need to scale back the rook operator pod.

subhamkrai:
@travisn @Madhu-1 Let's discuss the error in this thread.

So, currently, the flow is if we get any error after scaling down to the rook operator pod, and before scaling up the rook operator pod, I'm scaling back the rook operator first and then returning the error.

What I'm gathering from the comments now are,

1. We first want to return the error first and scale back the rook operator
2.  We'll leave it to the users to scale back the rook operator.

Let me know your thoughts.

pkg/restore/restore_crd.go
subhamkrai: actual because calling this restore command, we call `clientsets := GetClientsets(cmd.Context())` to check rook version and there we first check if rook operator pod is running or not. We do this for all command
```
Warning: rook version 'rook: v1.12.0-alpha.0.227.g3e248bf9a' is running a pre-release version of Rook.
```
This is the first line of any command operation. So,we are safe we don't need to think about case where rook operator pod is already down

Madhu-1:
> @travisn @Madhu-1 Let's discuss the error in this thread.
> 
> So, currently, the flow is if we get any error after scaling down to the rook operator pod, and before scaling up the rook operator pod, I'm scaling back the rook operator first and then returning the error.
> 
> What I'm gathering from the comments now are,
> 
> 1. We first want to return the error first and scale back the rook operator
> 2. We'll leave it to the users to scale back the rook operator.
> 
> Let me know your thoughts.

IMHO scaling back the operator makes sense only if it is scaled down by the tool not by the user, the user might have scaled it down for some other reasons.

subhamkrai:
> > @travisn @Madhu-1 Let's discuss the error in this thread.
> > So, currently, the flow is if we get any error after scaling down to the rook operator pod, and before scaling up the rook operator pod, I'm scaling back the rook operator first and then returning the error.
> > What I'm gathering from the comments now are,
> > 
> > 1. We first want to return the error first and scale back the rook operator
> > 2. We'll leave it to the users to scale back the rook operator.
> > 
> > Let me know your thoughts.
> 
> IMHO scaling back the operator makes sense only if it is scaled down by the tool not by the user, the user might have scaled it down for some other reasons.

Yes, for this command tool scale down the operator

pkg/restore/restore_crd.go
Madhu-1: Do you mean the tool will fail early if the rook operator pod is not running before coming here?

subhamkrai: it will initially print the log
```
"waiting for pod with label %q in namespace %q to be running
```
and after 60 iteration it will fail

pkg/restore/crd.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/debug"
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var getCrName = `
+	kubectl -n %s get %s -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.deletionGracePeriodSeconds}{"\n"}{end}' | awk '$2=="0" {print $1}'
```
travisn, line 33: > currently, the tool is doesn't have the capability of finding if it kubectl or oc. But I think it will be better if in the future if the tool is able to find the differences.

If we can't autodetect it yet, we need a CLI arg that lets them override it to `oc`.

pkg/restore/crd.go
```diff
@@ -0,0 +1,225 @@
...
+	kubectl -n %s patch %s/%s --type json --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'
+`
+
+var recreateCr = `
+	kubectl create -f %s
+`
+
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+	crd := args[0]
+	command := fmt.Sprintf(getCrName, clusterNamespace, crd)
```
travisn, line 54: IMO if the commands are only used once, the code is more readable to define them as a local variable here instead of up above in the file. There's no need for a package var.

pkg/restore/restore_crd.go
travisn: So if the operator is not running, it blocks all plugin commands? Instead, how about if we just print a warning that the version could not be detected because the operator is down. Then the commands won't be blocked. In some disaster recovery scenarios, it will likely be common that the admin could scale down the operator while they are troubleshooting, before running these commands.

pkg/restore/crd.go
```diff
@@ -0,0 +1,225 @@
...
+	logging.Info("Backed up crd %s named %s in file %s.yaml", crd, crName, crd)
+
+	webhookConfigName := "rook-ceph-webhook"
+	logging.Info("deleting validating webhook %s if present", webhookConfigName)
+	err = k8sclientset.Kube.AdmissionregistrationV1().ValidatingWebhookConfigurations().Delete(ctx, webhookConfigName, v1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		logging.Fatal(fmt.Errorf("failed to delete validating webhook %s. %v", webhookConfigName, err))
+	}
+
+	logging.Info("Fetching %s %s uid", crd, crName)
```
travisn, line 81:
```suggestion
	logging.Info("Fetching the UID for %s/%s", crd, crName)
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,225 @@
...
+	command = fmt.Sprintf(getCrUID, clusterNamespace, crd, crName)
+	uid := exec.ExecuteBashCommand(command)
+	logging.Info("Successfully fetched uid %s from %s %s", uid, crd, crName)
+
+	removeOwnerRefOfUID(ctx, k8sclientset, operatorNamespace, clusterNamespace, uid)
+
+	logging.Info("removing finalizers from %s %s", crd, crName)
```
travisn, line 88: nit: Some of these log statements start with capital letter and some with lower case. How about starting with upper case for consistency?
```suggestion
	logging.Info("Removing finalizers from %s %s", crd, crName)
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,225 @@
...
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("re-creating the crd %s from file %s.yaml created above", crd, crName)
+	command = fmt.Sprintf(recreateCr, crd+".yaml")
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Scaling up the operator to 1")
+	err = debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		scaleBackOperatorOnError(ctx, k8sclientset, operatorNamespace, err)
```
travisn, line 99: Line 97 already tries to scale up the operator, now it's attempting again on error?

pkg/restore/restore_crd.go
travisn: Right, we can print the error and a message about the operator pod still being scaled down so they can investigate the issue. We don't want the operator to start up if there is some intermediate state that might wipe the cluster.

travisn (changes requested):

pkg/restore/crd.go
subhamkrai: agreed, updated

.github/workflows/go-test.yaml
```diff
@@ -112,6 +112,22 @@ jobs:
           kubectl -n rook-ceph scale deployment rook-ceph-osd-0 --replicas 0
           kubectl rook-ceph rook purge-osd 0 --force
 
+      - name: Restore CRD without CRName
+        run: |
+          # First let's delete the cephCluster
+          kubectl -n rook-ceph delete cephcluster my-cluster --timeout 3s || true
+
+          kubectl rook-ceph -n rook-ceph restore cephcluster
```
travisn, line 120: Need documentation in the readme for this new command

pkg/restore/crd.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+	crd := args[0]
+	getCrName := `kubectl -n %s get %s -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.deletionGracePeriodSeconds}{"\n"}{end}' | awk '$2=="0" {print $1}'`
```
travisn, line 35: Is this command getting the deleted resource(s)? What if there are no resources? Will it detect that nothing needs to be done?

travisn: How about logging something like "detecting which resources to restore"?

pkg/restore/crd.go
```diff
@@ -0,0 +1,203 @@
...
+	if len(args) == 2 {
+		crName = args[1]
+	}
+
+	logging.Info("Restoring crd %s name %s", crd, crName)
+
+	logging.Info("Scaling down the operator to 0")
+	err := debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 0)
+	if err != nil {
+		logging.Fatal(err)
```
travisn, line 48: > So if the operator is not running, it blocks all plugin commands? Instead, how about if we just print a warning that the version could not be detected because the operator is down. Then the commands won't be blocked. In some disaster recovery scenarios, it will likely be common that the admin could scale down the operator while they are troubleshooting, before running these commands.

Did you see this comment? thanks

.github/workflows/go-test.yaml
```diff
@@ -112,6 +112,22 @@ jobs:
...
+          tests/github-action-helper.sh wait_for_crd_to_be_ready_default
```
travisn, line 121: What state is the CR in during this action? Will it require a successful reconcile to get back to Ready state? Looks like the CI is taking over a minute to wait for the Ready state. That seems longer than I would expect, but maybe that's normal.

travisn (changes requested):

pkg/restore/crd.go
subhamkrai: Do you want to do the above change for this command or all the controls in the plugin? I missed reading your comment earlier

.github/workflows/go-test.yaml
subhamkrai: in my local testing, local I have seen it takes around a minute or full reconciliation to mark this CR ready at least for the cephCluster cr

subhamkrai: I'll test again and see

pkg/restore/crd.go
subhamkrai: I'll test this but for now adding the logs

.github/workflows/go-test.yaml
subhamkrai: added, copied the docs from rook docs

.github/workflows/go-test.yaml
subhamkrai:
```
~/go/src/github.com/kubectl-rook-ceph
srai@192 ~ (restore-crd) $ kc get cephcluster
NAME         DATADIRHOSTPATH   MONCOUNT   AGE   PHASE         MESSAGE                   HEALTH   EXTERNAL   FSID
my-cluster   /var/lib/rook     1          61s   Progressing   Configuring Ceph Mgr(s)                       
~/go/src/github.com/kubectl-rook-ceph
srai@192 ~ (restore-crd) $ kc get cephcluster
NAME         DATADIRHOSTPATH   MONCOUNT   AGE   PHASE   MESSAGE                        HEALTH        EXTERNAL   FSID
my-cluster   /var/lib/rook     1          64s   Ready   Cluster created successfully   HEALTH_WARN              5dc8aca4-cb12-47ee-a83e-a9f773b39a0a
```

it took around 1 min

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	logging.Info("Backing up kubernetes and crd resources")
+	getCrYamlContent := `kubectl -n %s get %s %s -oyaml > %s.yaml`
```
subhamkrai, line 58:
```suggestion
	getCrYamlContent := `kubectl -n %s get %s %s -oyaml > %s.yaml`
```

This command will create the crd file which is used to re-create the cr later. Currently, this file is left there not removed after re-creating the cr. should we remove it after successful cr creation or leave the file as it is?

README.md
```diff
@@ -101,6 +101,7 @@ Visit docs below for complete details about each command and their flags uses.
 1. [Debug OSDs and Mons](docs/debug.md)
 1. [Restore mon quorum](docs/mons.md#restore-quorum)
 1. [Disaster Recovery](docs/dr-health.md)
+1. [Restore CRD](docs/crd.md)
```
travisn, line 104:
```suggestion
1. [Restore deleted CRs](docs/crd.md)
```

docs/crd.md
```diff
@@ -0,0 +1,39 @@
+# Restore the CRD stuck in delete mode
+
+When the Rook CRDs are deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data appears present in the cluster, Rook will refuse to allow the resources to be deleted since the operator will refuse to remove the finalizer on the CRs until the underlying data is deleted.
```
travisn, line 3:
```suggestion
When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
```

docs/crd.md
```diff
@@ -0,0 +1,39 @@
...
+
+While it is good that the CRs will not be deleted and the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, below command will allow you to restore the CRs to their prior state without even necessarily suffering cluster downtime.
```
travisn, line 5: The state will show as "Terminating", right?
```suggestion
While it is good that the CRs will not be deleted and the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Terminating state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
```

docs/crd.md
```diff
@@ -0,0 +1,39 @@
...
+
+```bash
+kubectl rook-ceph restore cephcluster [cephCRDName]
```
travisn, line 8:
```suggestion
kubectl rook-ceph restore cephcluster [cephClusterName]
```

docs/crd.md
```diff
@@ -0,0 +1,39 @@
+# Restore the CRD stuck in delete mode
+
+When the Rook CRDs are deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data appears present in the cluster, Rook will refuse to allow the resources to be deleted since the operator will refuse to remove the finalizer on the CRs until the underlying data is deleted.
+
+While it is good that the CRs will not be deleted and the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, below command will allow you to restore the CRs to their prior state without even necessarily suffering cluster downtime.
+
```
travisn, line 6: How about a separate section for the main syntax, and then for the example?
```
## Restore Command

[Show/describe the general syntax]

## CephCluster Restore Example

Sentence that describes the cluster example below...
```

docs/crd.md
```diff
@@ -0,0 +1,39 @@
+# Restore the CRD stuck in delete mode
```
travisn, line 1: Keep the L1 header short
```suggestion
# Restoring Deleted CRs
```

docs/crd.md
```diff
@@ -0,0 +1,39 @@
...
+
+When the Rook CRDs are deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data appears present in the cluster, Rook will refuse to allow the resources to be deleted since the operator will refuse to remove the finalizer on the CRs until the underlying data is deleted.
+
+While it is good that the CRs will not be deleted and the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, below command will allow you to restore the CRs to their prior state without even necessarily suffering cluster downtime.
+
+```bash
+kubectl rook-ceph restore cephcluster [cephCRDName]
+
+Warning: rook version 'rook: v1.12.0-alpha.0.241.ga64f9bccf' is running a pre-release version of Rook.
```
travisn, line 10: In the example output, let's pretend it's not pre-release :)
```suggestion
```

cmd/commands/restore.go
```diff
@@ -0,0 +1,33 @@
...
+package command
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/restore"
+	"github.com/spf13/cobra"
+)
+
+// RestoreCmd represents the restore commands
+var RestoreCmd = &cobra.Command{
+	Use:   "restore",
```
travisn, line 26: We could have multiple types of restore operations, already we have a `mons restore-quorum` command. So let's make this command more descriptive to avoid conflict with future types of restore.
```suggestion
	Use:   "restore-deleted",
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+	crd := args[0]
+	logging.Info("Detecting which ceph crd resources to restore")
```
travisn, line 35:
```suggestion
	logging.Infof("Detecting which ceph crd resources to restore for crd %q", crd)
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+	getCrName := `kubectl -n %s get %s -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.deletionGracePeriodSeconds}{"\n"}{end}' | awk '$2=="0" {print $1}'`
+	command := fmt.Sprintf(getCrName, clusterNamespace, crd)
+	crName := strings.TrimSpace(exec.ExecuteBashCommand(command))
+
+	if crName == "" {
+		logging.Info("Nothing to do here, no crd resources with deletionGracePeriodSeconds 0")
```
travisn, line 41:
```suggestion
		logging.Info("Nothing to do here, no %q resources in deleted state", crd)
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+		return
+	}
+
+	if len(args) == 2 {
```
travisn, line 45:
```suggestion
	if len(args) >= 2 {
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+		crName = args[1]
+	}
+
+	logging.Info("Restoring crd %s name %s", crd, crName)
```
travisn, line 49:
```suggestion
	logging.Info("Restoring CR %s", crName)
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+	crd := args[0]
+	logging.Info("Detecting which ceph crd resources to restore")
+	getCrName := `kubectl -n %s get %s -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.deletionGracePeriodSeconds}{"\n"}{end}' | awk '$2=="0" {print $1}'`
+	command := fmt.Sprintf(getCrName, clusterNamespace, crd)
+	crName := strings.TrimSpace(exec.ExecuteBashCommand(command))
```
travisn, line 38: What if there are multiple deleted resources? We wouldn't expect this for a CephCluster, but it could happen for other CRs.

pkg/restore/crd.go
travisn: Any command that doesn't require the operator should not be blocked if the operator is not running. So I'd suggest:
1. The version check should only print a message if it couldn't detect the version. 
2. Add a new helper that waits for the operator to start, which is only called by commands that require the operator. 
3. If this call to scale down the operator fails because it's already not running, then we should continue since the intended state of stopping the operator is already in the desired state.

travisn: And this could certainly be a separate PR, but I'd suggest we implement it before releasing this feature.

pkg/restore/crd.go
travisn: If the restore is successful, seems like we should deleted it. It's really a temporary file. But if there is any failure, we should not delete it.

travisn: Let's have a more unique filename that is the `crd-crname.yaml`, in case there are multiple CRs to restore.
```suggestion
	getCrYamlContent := `kubectl -n %s get %s %s -oyaml > %s-%s.yaml`
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+	err = k8sclientset.Kube.AdmissionregistrationV1().ValidatingWebhookConfigurations().Delete(ctx, webhookConfigName, v1.DeleteOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		logging.Fatal(fmt.Errorf("failed to delete validating webhook %s. %v", webhookConfigName, errors.Wrapf(err, "Operator pod still being scaled down")))
+	}
+
+	logging.Info("Fetching the UID for %s/%s", crd, crName)
+	getCrUID := `kubectl -n %s get %s %s -o 'jsonpath={.metadata.uid}'`
+	command = fmt.Sprintf(getCrUID, clusterNamespace, crd, crName)
+	uid := exec.ExecuteBashCommand(command)
+	logging.Info("Successfully fetched uid %s from %s %s", uid, crd, crName)
```
travisn, line 74: Wherever we have the crd and cr name together, let's separate with a `/`
```suggestion
	logging.Info("Successfully fetched uid %s from %s/%s", uid, crd, crName)
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+
+	removeOwnerRefOfUID(ctx, k8sclientset, operatorNamespace, clusterNamespace, uid)
+
+	logging.Info("Removing finalizers from %s %s", crd, crName)
+	removeFinalizers := `kubectl -n %s patch %s/%s --type json --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'`
+	command = fmt.Sprintf(removeFinalizers, clusterNamespace, crd, crName)
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Re-creating the crd %s from file %s.yaml created above", crd, crName)
+	recreateCr := `kubectl create -f %s`
```
travisn, line 84: Let's use CR instead of Cr in var names
```suggestion
	recreateCR := `kubectl create -f %s`
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+	command = fmt.Sprintf(recreateCr, crd+".yaml")
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Scaling up the operator to 1")
+	err = debug.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 1)
+	if err != nil {
+		logging.Fatal(errors.Wrapf(err, "Operator pod still being scaled down"))
```
travisn, line 91: The operator is scaling up not down?

pkg/restore/crd.go
```diff
@@ -0,0 +1,209 @@
...
+		for _, owner := range secret.OwnerReferences {
+			if string(owner.UID) == targetUID {
+				logging.Info("Removing owner references for secret %s", secret.Name)
+				// Remove ownerReferences.
+				secret.OwnerReferences = nil
+
+				// Update the Secret without ownerReferences.
+				_, err := k8sclientset.Kube.CoreV1().Secrets(clusterNamespace).Update(ctx, &secret, v1.UpdateOptions{})
+				if err != nil {
+					logging.Fatal(errors.Wrapf(err, "Operator pod still being scaled down"))
```
travisn, line 115: Scaling the operator down? Seems like this same error was copied many places.

travisn (changes requested):

pkg/restore/crd.go
subhamkrai: Good question, I'll need to test this but currently, it will get all the names. It requires in command

pkg/restore/crd.go
subhamkrai: why `>=`? with the above conditions means the user has passed some extra arguments although cobra will fail while validating the command it will not reach here.

docs/crd.md
subhamkrai: Last I remember it was saying `Deleting ....`, I'll confirm again and update

subhamkrai:
depends #180

.github/workflows/go-test.yaml
```diff
@@ -112,6 +112,22 @@ jobs:
           kubectl -n rook-ceph scale deployment rook-ceph-osd-0 --replicas 0
           kubectl rook-ceph rook purge-osd 0 --force
 
+      - name: Restore CRD without CRName
+        run: |
+          # First let's delete the cephCluster
+          kubectl -n rook-ceph delete cephcluster my-cluster --timeout 3s || true
```
BlaineEXE, line 118: Is this equivalent?

```suggestion
          kubectl -n rook-ceph delete cephcluster my-cluster --timeout 3s --wait=false
```

cmd/commands/restore.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/restore"
+	"github.com/spf13/cobra"
+)
+
+// RestoreCmd represents the restore commands
+var RestoreCmd = &cobra.Command{
+	Use:   "restore-deleted",
+	Short: "Reads the required arg CRD and optional CR name to restore. Ex: restore cephcluster <my-cluster>",
```
travisn, line 27: Let's describe what the command does. The arguments are shown by the help command anyway, right? No need to describe them here?
```suggestion
	Short: "Restores a CR that was accidentally deleted and is still in terminating state.",
```

docs/crd.md
```diff
@@ -0,0 +1,50 @@
+# Restoring Deleted CRs
+
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While it is good that the CRs will not be deleted and the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
```
travisn, line 5:
```suggestion
While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
```

docs/crd.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Restore Command
+
+```bash
+kubectl rook-ceph restore-deleted <CRDName> [CRName]
```
travisn, line 11:
```suggestion
kubectl rook-ceph restore-deleted <CRD> [CRName]
```

docs/crd.md
```diff
@@ -0,0 +1,50 @@
...
+
+# `restore-deleted`: is the command to restore the deleted CR
+# `<CRDName>`: the CRD deleting you want to restore. Example: cephcluster, cephfilesystem, cephblockpool and so on.
```
travisn, line 14:
```suggestion
# `<CRD>`: the CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
```

docs/crd.md
```diff
@@ -0,0 +1,50 @@
...
+# `[CRName]`: is the name of the specific CR which you want to restore since there can be multiple CRs under the same CRD. Example let's say you multiple cephfilesystem stuck in deleting state in your cluster named `filesystem-1`, `filesystem-2` and you want to restore to store only `filesystem-2`. In this case you can pass the name of the specifc CR, like this: `restore-deleted cephfilesystem filesystem-2`.
```
travisn, line 15: Shorter can be more concise to understand.
```suggestion
# `[CRName]`: is the name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/rook/kubectl-rook-ceph/pkg/exec"
+	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+	crd := args[0]
+	logging.Info("Detecting which ceph crd resources to restore for crd %q", crd)
```
travisn, line 36:
```suggestion
	logging.Info("Detecting which resources to restore for crd %q", crd)
```

pkg/restore/crd.go
travisn: Ok sounds good if it doesn't reach here with more args anyway

pkg/restore/crd.go
```diff
@@ -0,0 +1,215 @@
...
+	logging.Info("Successfully fetched uid %s from %s/%s", uid, crd, crName)
+
+	removeOwnerRefOfUID(ctx, k8sclientset, operatorNamespace, clusterNamespace, uid)
+
+	logging.Info("Removing finalizers from %s/%s", crd, crName)
+	removeFinalizers := `kubectl -n %s patch %s/%s --type json --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'`
+	command = fmt.Sprintf(removeFinalizers, clusterNamespace, crd, crName)
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Re-creating the crd %s from file %s created above", crd, crFileName)
```
travisn, line 85:
```suggestion
	logging.Info("Re-creating the CR %s from file %s created above", crd, crFileName)
```

travisn (changes requested):

cmd/commands/restore.go
subhamkrai: correct

subhamkrai: actually, what we describe in `Short:` is printed in help command

.github/workflows/go-test.yaml
subhamkrai: seems like yes

docs/crd.md
```diff
@@ -0,0 +1,49 @@
...
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
+
+## Restore Command
+
+```bash
+kubectl rook-ceph restore-deleted <CRD> [CRName]
+
+# `restore-deleted`: is the command to restore the deleted CR
```
travisn, line 12: No need to describe this command here, it was already described above
```suggestion
```

docs/crd.md
```diff
@@ -0,0 +1,49 @@
...
+# `<CRD>`: the CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
```
travisn, line 13: Put these argument descriptions in bullets instead of inside the code block. The code block is generally just used for commands and their output. Also, I'd suggest moving these argument descriptions above the command example (before line 10)
```suggestion
- `<CRD>`: the CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
```

pkg/restore/crd.go
```diff
@@ -0,0 +1,220 @@
...
+	logging.Info("Restoring CR %s", crName)
+
+	logging.Info("Scaling down the operator to 0")
+	err := k8sutil.SetDeploymentScale(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator", 0)
+	if err != nil {
+		deploy, er := k8sutil.GetDeployment(ctx, k8sclientset.Kube, operatorNamespace, "rook-ceph-operator")
+		if er != nil && !apierrors.IsNotFound(er) {
+			logging.Fatal(er)
+		}
+		if deploy.Status.ReadyReplicas != 0 {
```
travisn, line 58: If `er != nil` and the deployment was not found, `deploy` could be nil, right?

pkg/restore/crd.go
```diff
@@ -0,0 +1,220 @@
...
+
+	logging.Info("Removing finalizers from %s/%s", crd, crName)
+	removeFinalizers := `kubectl -n %s patch %s/%s --type json --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'`
+	command = fmt.Sprintf(removeFinalizers, clusterNamespace, crd, crName)
+	logging.Info(exec.ExecuteBashCommand(command))
+
+	logging.Info("Re-creating the CR %s from file %s created above", crd, crFileName)
+	recreateCR := `kubectl create -f %s`
+	command = fmt.Sprintf(recreateCR, crFileName)
+	logging.Info(exec.ExecuteBashCommand(command))
```
travisn, line 93: Can this command fail? If so, is it fatal? We wouldn't want to continue and start the operator if it did fail.

docs/crd.md
```diff
@@ -0,0 +1,49 @@
...
+# `[CRName]`: is the name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
+```
+
+## CephCluster Restore Example
+
+```bash
+kubectl rook-ceph restore-deleted cephcluster [cephClusterName]
+
+Info: Detecting which ceph crd resources to restore
+Info: Restoring crd cephcluster name my-cluster
```
travisn, line 23: After we detect which resource(s) to restore, I'm wondering if we should pause and confirm. There is some risk to this operation, so it felt very sudden to continue without any prompt. For example, prompt:
 "The resource %s was found deleted. Do you want to restore it? yes | no". Only if they type `y` or `yes` then we would continue. It's not super risky so we don't need a longer confirmation string, but something simple like this seems best. Thoughts?

travisn (changes requested):

pkg/restore/crd.go
subhamkrai: if the above commands, it will exit from their only. Method `ExecuteBashCommand` has check 
```
    stdout, err := cmd.Output()
	if err != nil {
		logging.Fatal(err)
	}
```

I think we are good here?

docs/crd.md
subhamkrai: that's a good point and I agree this is a kind of destructive command better confirmed by user input.

And `yes` is better than just `y` for input.

pkg/restore/crd.go
subhamkrai: right, added the check `notFound`

docs/crd.md
```diff
@@ -0,0 +1,87 @@
+# Restoring Deleted CRs
+
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
+
+## Restore Command
+
```
travisn, line 8: An intro sentence for the section would be helpful, such as:
The `restore-deleted` command has one required and one optional parameter:

docs/crd.md
```diff
@@ -0,0 +1,87 @@
...
+- `<CRD>`: the CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
+- `[CRName]`: is the name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
+
+```bash
+kubectl rook-ceph restore-deleted <CRD> [CRName]
+
```
travisn, line 14: The next section has the example output. Seems like you can delete lines 14-43 and just have this section focus on the parameters and basic description.

docs/crd.md
```diff
@@ -0,0 +1,87 @@
...
+Info: Detecting which resources to restore for crd "cephcluster"
+Info: Restoring CR my-cluster
+Warning: The resource my-cluster was found deleted. Do you want to restore it? yes | no
+
+Info: skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true
+Info: Scaling down the operator to 0
+Info: Backing up kubernetes and crd resources
+---
```
travisn, line 22: Why are there so many separators? `---`

pkg/restore/crd.go
travisn: Get a chance to look at this?

pkg/mons/restore_quorum.go
```diff
@@ -278,13 +278,15 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 	return badMons, goodMonPublicIp, goodMonPort, nil
 }
 
-func promptToContinueOrCancel(answer string) (string, error) {
+func PromptToContinueOrCancel(inputSource, answer string) (string, error) {
```
travisn, line 281: How about the expected answer as a parameter? Then you don't need an if statement to hard code the answer inside the method
```suggestion
func PromptToContinueOrCancel(inputSource, expectedAnswer, answer string) (string, error) {
```

travisn (changes requested):

pkg/restore/crd.go
subhamkrai: somehow it was skipped with all the comments, I'll look this tomorrow

docs/crd.md
travisn: Ah, I would expect `...` to represent removing some lines

pkg/mons/restore_quorum.go

docs/crd.md
subhamkrai: removed

docs/crd.md
subhamkrai: added

pkg/restore/crd.go
subhamkrai: Good catch, I need to add `| head -n 1`  in the end to get only the first line.

subhamkrai: it is working now

cmd/commands/restore.go
```diff
@@ -0,0 +1,34 @@
...
+package command
+
+import (
+	"github.com/rook/kubectl-rook-ceph/pkg/restore"
+	"github.com/spf13/cobra"
+)
+
+// RestoreCmd represents the restore commands
+var RestoreCmd = &cobra.Command{
+	Use:   "restore-deleted",
```
subhamkrai, line 26:
```suggestion
	Use:   "restore-deleted",
```

any suggestion on the name?

travisn: `undelete` is another option. `restore-deleted` still sounds better to me, but we could take a vote. Other thoughts?

docs/crd.md
```diff
@@ -0,0 +1,60 @@
+# Restoring Deleted CRs
+
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
+
+## Restore Command
```
travisn, line 7:
```suggestion
## restore-deleted Command
```

docs/crd.md
```diff
@@ -0,0 +1,60 @@
...
+
+The `restore-deleted` command has one required and one optional parameter:
+
+- `<CRD>`: the CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
+- `[CRName]`: is the name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
```
travisn, line 12:
```suggestion
- `[CRName]`: The name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
```

docs/crd.md
```diff
@@ -0,0 +1,60 @@
...
+
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
+
+## Restore Command
+
+The `restore-deleted` command has one required and one optional parameter:
+
+- `<CRD>`: the CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
```
travisn, line 11: It's not a sentence, but it seems better capitalized.
```suggestion
- `<CRD>`: The CRD type that is to be restored, such as CephCluster, CephFilesystem, CephBlockPool and so on.
```

docs/crd.md
```diff
@@ -0,0 +1,60 @@
...
+- `[CRName]`: is the name of the specific CR which you want to restore since there can be multiple instances under the same CRD. For example, if there are multiple CephFilesystems stuck in deleting state, a specific filesystem can be restored: `restore-deleted cephfilesystem filesystem-2`.
+
+```bash
+kubectl rook-ceph restore-deleted <CRD> [CRName]
+```
+
+## CephCluster Restore Example
+
+```bash
+kubectl rook-ceph restore-deleted cephcluster [cephClusterName]
```
travisn, line 21: For the example, let's show the command exactly as executed
```suggestion
kubectl rook-ceph restore-deleted cephcluster
```

docs/crd.md
```diff
@@ -0,0 +1,60 @@
...
+Info: Deleting validating webhook rook-ceph-webhook if present
+Info: Fetching the UID for cephcluster/my-cluster
+Info: Successfully fetched uid 9575a4c9-7d0f-4b98-93d5-0b7dab7efeb4 from cephcluster/my-cluster
+Info: Removing ownerreferences from resources with matching uid 9575a4c9-7d0f-4b98-93d5-0b7dab7efeb4
+Info: Removing owner references for secret cluster-peer-token-my-cluster
+Info: Removed ownerReference for Secret: cluster-peer-token-my-cluster
+
+Info: Removing owner references for secret rook-ceph-admin-keyring
+Info: Removed ownerReference for Secret: rook-ceph-admin-keyring
+---
```
travisn, line 41: If these are to show removed text, I'd suggest changing to dots and only one line of them is needed. But perhaps for this example we don't need to remove any text since it's not terribly long for this command.
```suggestion
...
```

pkg/mons/restore_quorum.go
```diff
@@ -278,13 +279,15 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 	return badMons, goodMonPublicIp, goodMonPort, nil
 }
 
-func promptToContinueOrCancel(answer string) (string, error) {
+func PromptToContinueOrCancel(inputSource, expectedAnswer, answer string) (string, error) {
```
travisn, line 282: Just return an error. The caller can choose to log something specific, no need to return it here.
```suggestion
func PromptToContinueOrCancel(inputSource, expectedAnswer, answer string) error {
```

pkg/mons/restore_quorum.go
```diff
@@ -278,13 +279,15 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
 	if skip, ok := os.LookupEnv("ROOK_PLUGIN_SKIP_PROMPTS"); ok && skip == "true" {
 		return "skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true", nil
```
travisn, line 284:
```suggestion
		logger.Info("skipped prompt since ROOK_PLUGIN_SKIP_PROMPTS=true")
		return nil
```

pkg/mons/restore_quorum.go
```diff
@@ -278,13 +279,15 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
 	}
 
-	if answer == "yes-really-restore" {
-		return "proceeding", nil
+	if inputSource == "restore-quorum" && answer == expectedAnswer {
```
travisn, line 287: The input source shouldn't matter when comparing the answer
```suggestion
	if answer == expectedAnswer {
```

pkg/restore/crd.go
travisn: So if there are multiple deleted resources and they don't specify the CR name, they need to run the command again to restore all of them, right? Maybe mention that in the doc.

travisn (changes requested):

cmd/commands/restore.go
subhamkrai: I'm okay with  `restore-deleted`

pkg/mons/restore_quorum.go
subhamkrai: right, removed

docs/crd.md
subhamkrai: updated

travisn (changes requested):
Just a few small suggestions

docs/crd.md
```diff
@@ -0,0 +1,112 @@
+# Restoring Deleted CRs
+
+When a Rook CR is deleted, the Rook operator will respond to the deletion event to attempt to clean up the cluster resources. If any data is still present in the cluster, Rook will refuse to delete the CR to ensure data is not lost. The operator will refuse to remove the finalizer on the CR until the underlying data is deleted.
+
+While the underlying Ceph data and daemons continue to be available, the CRs will be stuck indefinitely in a Deleting state in which the operator will not continue to ensure cluster health. Upgrades will be blocked, further updates to the CRs are prevented, and so on. Since Kubernetes does not allow undeleting resources, the command below will allow repairing the CRs without even necessarily suffering cluster downtime.
+
+> [!NOTE]
+> If there are multiple deleted resources in the cluster and no specific resource is mentioned, in that case the first resource will be restored and in order to restore all resources we need to re-run the command.
```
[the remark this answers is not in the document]
subhamkrai: Yes, we could loop through all the resources but IMO it's better to restore one resource at a time. It's better for monitoring and debugging if something goes wrong. Also, users may get overwhelmed with all the log messages.

Also, we later decide to restore all the resources at once. In that case, it will be better if we do so after Javier's PR since he is adding the code to list resources from dynamic API, it will help in looping over the resource. And currently we are getting the resource from the command and then reading it as a string.

travisn (approved):

docs/crd.md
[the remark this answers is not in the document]
travisn: Ok we can do this for now, it's really a corner case anyway.

PR #180, opened 2023-09-19
command: skip operator pod check when not required

subhamkrai:
we should not be waiting or doing any operation realted to operator when it is not really required. Also, in some cases user may have intentionally scale down the operator. That's why we should skip doing operator operation when not required.

### DIFF
--- a/cmd/commands/ceph.go
+++ b/cmd/commands/ceph.go
@@ -30,6 +30,7 @@ var CephCmd = &cobra.Command{
 	Args:               cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		clientsets := GetClientsets(cmd.Context())
+		VerifyOperatorPodIsRunning(cmd.Context(), clientsets, OperatorNamespace, CephClusterNamespace)
 		logging.Info("running 'ceph' command with args: %v", args)
 		exec.RunCommandInOperatorPod(cmd.Context(), clientsets, cmd.Use, args, OperatorNamespace, CephClusterNamespace, false, true)
 	},
--- a/cmd/commands/operator.go
+++ b/cmd/commands/operator.go
@@ -35,6 +35,7 @@ var restartCmd = &cobra.Command{
 	Args:  cobra.NoArgs,
 	Run: func(cmd *cobra.Command, _ []string) {
 		clientsets := GetClientsets(cmd.Context())
+		VerifyOperatorPodIsRunning(cmd.Context(), clientsets, OperatorNamespace, CephClusterNamespace)
 		k8sutil.RestartDeployment(cmd.Context(), clientsets.Kube, OperatorNamespace, "rook-ceph-operator")
 	},
 }
--- a/cmd/commands/root.go
+++ b/cmd/commands/root.go
@@ -116,7 +116,9 @@ func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, o
 	if err != nil {
 		logging.Fatal(fmt.Errorf("CephCluster namespace '%s' does not exist. %v", cephClusterNamespace, err))
 	}
+}
 
+func VerifyOperatorPodIsRunning(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string) {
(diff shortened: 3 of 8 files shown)


cmd/commands/root.go
```diff
@@ -102,12 +102,12 @@ func GetClientsets(ctx context.Context) *k8sutil.Clientsets {
...
 	}
 
-	PreValidationCheck(ctx, clientsets, OperatorNamespace, CephClusterNamespace)
+	PreValidationCheck(ctx, clientsets, OperatorNamespace, CephClusterNamespace, checkRookVersion)
 
 	return clientsets
 }
 
-func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string) {
+func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string, checkRookVersion bool) {
```
travisn, line 110: Is this parameter really about whether we require the operator pod to be running? While we implement that check by checking the rook version, I just want to be clear when `false` or `true` should be passed to this method.
```suggestion
func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string, requireOperatorPodRunning bool) {
```

cmd/commands/ceph.go
```diff
@@ -29,7 +29,7 @@ var CephCmd = &cobra.Command{
 	DisableFlagParsing: true,
 	Args:               cobra.MinimumNArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		clientsets := GetClientsets(cmd.Context())
+		clientsets := GetClientsets(cmd.Context(), true)
```
travisn, line 32: Could we separate this into two API calls such as this? To me, it would be more readable.
```suggestion
		clientsets := GetClientsets(cmd.Context())
		CheckOperatorPod(cmd.Context())
```

cmd/commands/root.go
subhamkrai: removed the check

cmd/commands/ceph.go
subhamkrai: agreed

cmd/commands/operator.go
```diff
@@ -45,6 +46,7 @@ var setCmd = &cobra.Command{
 	Args:  cobra.ExactArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
 		clientsets := GetClientsets(cmd.Context())
+		CheckOperatorPod(cmd.Context(), clientsets, OperatorNamespace, CephClusterNamespace)
```
travisn, line 49: Why does the operator pod need to be running? Aren't we just updating the configmap?

cmd/commands/root.go
```diff
@@ -116,7 +116,9 @@ func PreValidationCheck(ctx context.Context, k8sclientset *k8sutil.Clientsets, o
 	if err != nil {
 		logging.Fatal(fmt.Errorf("CephCluster namespace '%s' does not exist. %v", cephClusterNamespace, err))
 	}
+}
 
+func CheckOperatorPod(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string) {
```
travisn, line 121: How about this?
```suggestion
func VerifyOperatorPodIsRunning(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, cephClusterNamespace string) {
```

travisn (changes requested):

cmd/commands/operator.go
subhamkrai: we are making a change in configmap `rook-operator...` and if the pod is not running there will be no effect right? To update in values in the configmap operator pod should be running otherwise values will be updated but operator will not reconcile those

travisn: Correct, the operator wouldn't reconcile them if it's down. I can imagine someone would want to update a setting and then starting the operator separately, for example to see different log levels at startup.

subhamkrai: okay, I have removed the check

travisn (approved):

Issue #181, opened 2023-09-19
Allow multiple OSDs to be purged in a single operation


jhoblitt:
I often need to remove multiple OSDs at the same time when a node is/was removed from the cluster.  This morning, I resorted to a bash loop:

```bash
$ for O in 103 108 11 113 118 123 128 133 14 17 2 21 25 29 33 38 43 48 5 53 58 63 68 73 78 8 83; do kubectl rook-ceph rook purge-osd --force $O; done
```

It would be convenient if the `purge-osd` sub-command accepted a list of OSD IDs.

subhamkrai:
@jhoblitt did you try `kubectl rook-ceph purge-osd $OSD_ID,$OSD_ID... --force`

jhoblitt:
@subhamkrai I didn't... I looked at the `-h` but I apparently didn't read the first line of the output...

```
$ k rook-ceph rook purge-osd -h
Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.

Usage:
  rook-ceph rook purge-osd [flags]

Flags:
      --force   force deletion of an OSD if the OSD still contains data
  -h, --help    help for purge-osd

Global Flags:
      --context string              Kubernetes context to use
      --kubeconfig string           kubernetes config path
  -n, --namespace string            Kubernetes namespace where CephCluster is created (default "rook-ceph")
      --operator-namespace string   Kubernetes namespace where rook operator is running
```

I don't have any cobra-fu.  Is it possible to add examples to the "usage" string?

subhamkrai:
> @subhamkrai I didn't... I looked at the `-h` but I apparently didn't read the first line of the output...
> 
> ```
> $ k rook-ceph rook purge-osd -h
> Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
> 
> Usage:
>   rook-ceph rook purge-osd [flags]
> 
> Flags:
>       --force   force deletion of an OSD if the OSD still contains data
>   -h, --help    help for purge-osd
> 
> Global Flags:
>       --context string              Kubernetes context to use
>       --kubeconfig string           kubernetes config path
>   -n, --namespace string            Kubernetes namespace where CephCluster is created (default "rook-ceph")
>       --operator-namespace string   Kubernetes namespace where rook operator is running
> ```
> 
> I don't have any cobra-fu. Is it possible to add examples to the "usage" string?

[...]

PR #182, opened 2023-09-22
usages: Clear how to use purge-osd command

subhamkrai:
It was unclear in the help section of purge-osd
on how to remove multiple osd at once. So adding
an example command to remove multiple osd.

Fixes: #181

### DIFF
--- a/cmd/commands/rook.go
+++ b/cmd/commands/rook.go
@@ -47,7 +47,7 @@ var versionCmd = &cobra.Command{
 
 var purgeCmd = &cobra.Command{
 	Use:   "purge-osd",
-	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.",
+	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs, for example, purge-osd 0,1",
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		clientsets := GetClientsets(cmd.Context())


subhamkrai:
@jhoblitt now is it clear on the usages
```yaml
/bin/kubectl-rook-ceph rook purge-osd -h
Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs. For example: kubectl rook-ceph rook purge-osd 0,1

Usage:
  rook-ceph rook purge-osd [flags]

Flags:
      --force   force deletion of an OSD if the OSD still contains data
  -h, --help    help for purge-osd

Global Flags:
      --context string              Kubernetes context to use
      --kubeconfig string           kubernetes config path
  -n, --namespace string            Kubernetes namespace where CephCluster is created (default "rook-ceph")
      --operator-namespace string   Kubernetes namespace where rook operator is running
```

parth-gr (approved):

cmd/commands/rook.go
```diff
@@ -47,7 +47,7 @@ var versionCmd = &cobra.Command{
 
 var purgeCmd = &cobra.Command{
 	Use:   "purge-osd",
-	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.",
+	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs. For example: kubectl rook-ceph rook purge-osd 0,1",
```
travisn, line 50: I might suggest a slightly simpler example for the CLI help, but I'm ok if you want to keep your text too.
```suggestion
	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs, for example, purge-osd 0,1",
```

travisn (approved):

cmd/commands/rook.go
subhamkrai: updated, will merge after the CI

PR #183, opened 2023-09-27
mons: adding missing info to start operator

subhamkrai:
somehow log to ask user to press enter to
start the operator and continue starting
mon quorum was missing, adding now. It was
mentioned in the doc but missing the command.


travisn (approved):
The behavior before this fix was that you had to press enter, but nothing told the user to press enter? I guess it looked hung?

subhamkrai:
> The behavior before this fix was that you had to press enter, but nothing told the user to press enter? I guess it looked hung?

yess

PR #186, opened 2023-10-02
docs: improve wording in rook command documentation

obnoxxx:
just some rewording that occurred to me when reading the docs

### DIFF
--- a/docs/rook.md
+++ b/docs/rook.md
@@ -1,17 +1,21 @@
 # Rook
 
-Rook is parent command which requires sub-command to get working. Currently, rook supports these sub-commands:
+The `rook` command supports the following sub-commands:
 
-1. `purge-osd <osd-id> [--force]` : [purge osd](#purge-osd) permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
+1. `purge-osd <osd-id> [--force]` : [purge osd](#purge-osd) permanently remove an OSD from the cluster. Multiple OSDs can be removed in a single command with a comma-separated list of IDs.
 2. `version`: [version](#version) prints the rook version.
 3. `status`      : [status](#status) print the phase and conditions of the CephCluster CR
 4. `status all`  : [status all](#status-all) print the phase and conditions of all CRs
 5. `status <CR>` : [status  cr](#status-cr-name) print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc
 
-## Purge OSD
+## Purge an OSD
 
-Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
+Permanently remove an OSD from the cluster. 
 
+
+!!! warning
+    Data loss is possible when passing the --force flag if the PGs are not healthy on other OSDs.
+    
 ```bash
 kubectl rook-ceph rook purge-osd 0 --force
 
@@ -23,6 +27,8 @@ kubectl rook-ceph rook purge-osd 0 --force
 # 2022-09-14 08:58:29.421345 I | cephosd: validating status of osd.0
 # 2022-09-14 08:58:29.421371 I | cephosd: osd.0 is healthy. It cannot be removed unless it is 'down'
 ```
+
+Multiple OSDs can be removed in one invocation with a comma-separated list of IDs.
 
 ## Version
 


docs/rook.md
```diff
@@ -1,16 +1,16 @@
 # Rook
 
-Rook is parent command which requires sub-command to get working. Currently, rook supports these sub-commands:
+Rook is a parent command and  requires sub-command to get working. Currently, rook supports the following  sub-commands:
```
travisn, line 3: How about simplifying?
```suggestion
The `rook` command supports the following  sub-commands:
```

docs/rook.md
```diff
@@ -1,16 +1,16 @@
...
 
-1. `purge-osd <osd-id> [--force]` : [purge osd](#purge-osd) permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
+1. `purge-osd <osd-id> [--force]` : [purge osd](#purge-osd) permanently remove an OSD from the cluster. Multiple OSDs can be removed in a single command with a comma-separated list of IDs.
 2. `version`: [version](#version) prints the rook version.
 3. `status`      : [status](#status) print the phase and conditions of the CephCluster CR
 4. `status all`  : [status all](#status-all) print the phase and conditions of all CRs
 5. `status <CR>` : [status  cr](#status-cr-name) print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc
 
-## Purge OSD
+## Purge OSDs
```
travisn, line 11: This section shows an example of purging a single OSD. How about keeping it singular?
```suggestion
## Purge an OSD
```

docs/rook.md
```diff
@@ -1,16 +1,16 @@
...
 
-Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
+Permanently remove an OSD from the cluster. Multiple OSDs can be removed in one invocation with a comma-separated list of IDs.
```
travisn, line 13: How about moving the comment about multiple OSDs below the example? The example is only for a single OSD.
```suggestion
Permanently remove an OSD from the cluster. 

(below the output)
Multiple OSDs can be removed in one invocation with a comma-separated list of IDs.
```

docs/rook.md
```diff
@@ -1,16 +1,16 @@
...
```
travisn, line 14: Let's add a warning
```suggestion
!!! warning
    Data loss is possible when passing the --force flag if the PGs are not healthy on other OSDs.
```

docs/rook.md
```diff
@@ -62,7 +62,7 @@ kubectl rook-ceph rook status
 #       "lastHeartbeatTime": "2022-09-14T08:57:02Z",
 #       "lastTransitionTime": "2022-09-14T08:56:59Z",
 #       "message": "Cluster created successfully",
-#       "reason": "ClusterCreated",
+#       "reason": "ClusterCreated",`
```
travisn, line 65: revert?

travisn (changes requested):

obnoxxx:
@travisn, I think I addressed all your requests  and/or committed and squashed all your suggestions

docs/rook.md
obnoxxx: done

travisn (changes requested):

docs/rook.md
```diff
@@ -1,17 +1,21 @@
...
 4. `status all`  : [status all](#status-all) print the phase and conditions of all CRs
 5. `status <CR>` : [status  cr](#status-cr-name) print the phase and conditions of CRs of a specific type, such as 'cephobjectstore', 'cephfilesystem', etc
 
-## Purge OSD
+## Purge an OSD
 
-Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
+Permanently remove an OSD from the cluster. 
 
+Multiple OSDs can be removed in one invocation with a comma-separated list of IDs.
```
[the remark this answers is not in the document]
obnoxxx: done!

travisn (approved):

docs/rook.md
```diff
@@ -1,17 +1,21 @@
...
 
-## Purge OSD
+## Purge an OSD
 
-Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs.
+Permanently remove an OSD from the cluster. 
 
+
+!!! warning
+    Data loss is possible when passing the --force flag if the PGs are not healthy on other OSDs.
```
BlaineEXE, line 17: I'm not sure that all users will know what a PG is. It might be good to link to ceph docs.

BlaineEXE (approved):
Approving with a minor suggestion

docs/rook.md
obnoxxx: good point, @BlaineEXE ! would you consider it sufficient to add that in a follow-up PR?

BlaineEXE: My experience is that most follow-up PRs don't actually happen. This is already merged, so it'll have to be in follow-up

PR #187, opened 2023-10-02
build: fix `make clean`

obnoxxx:
Signed-off-by: Michael Adam <obnox@samba.org>

`make clean` was broken because the rm command pointed to a non-existing file location.

This fixes this by using the correct path.


travisn (changes requested):
The change looks good. In the commit message, please just move the line with `Signed-off-by` after the commit description.

obnoxxx:
@travisn  wrote:
> The change looks good. In the commit message, please just move the line with `Signed-off-by` after the commit description.

Done, thanks!

travisn (approved):

PR #201, opened 2023-11-17
osd: get ack from user before osd removal

subhamkrai:
To avoid accidental removal of osd and making
sure user wants to remova the osd, adding changes
to get ack from user where user have to give ack
by writing `yes-i-really-mean-it`.

Resolves #

**Checklist:**

### DIFF
--- a/cmd/commands/rook.go
+++ b/cmd/commands/rook.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	"github.com/rook/kubectl-rook-ceph/pkg/mons"
 	"github.com/rook/kubectl-rook-ceph/pkg/rook"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


cmd/commands/rook.go
```diff
@@ -50,10 +51,17 @@ var purgeCmd = &cobra.Command{
 	Short: "Permanently remove an OSD from the cluster. Multiple OSDs can be removed with a comma-separated list of IDs, for example, purge-osd 0,1",
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
+		var answer string
+		osdID := args[0]
+		logging.Warning("Are you sure you want to purge osd ID %s? If so, enter 'yes-i-really-mean-it'\n", osdID)
+		fmt.Scanf("%s", &answer)
+		if err := mons.PromptToContinueOrCancel("yes-i-really-mean-it", answer); err != nil {
```
travisn, line 58: Can we only prompt in the case safe-to-destroy returns false? If it is safe to destroy, then seems like no need for the prompt.

travisn (changes requested):

cmd/commands/rook.go
subhamkrai: @travisn we check for `safe-to-destroy` in rook code, so how we add the prompt there? We can do the check `safe-to-destroy` here also but it will duplicate the check or we can keep the prompt in the start only like it is now. 
Thoughts?

travisn: Today the API call is:
```go
func RemoveOSDs(context *clusterd.Context, clusterInfo *client.ClusterInfo, osdsToRemove []string, preservePVC, forceOSDRemoval bool) error {
```

We could change this to have a callback such as:
```go
func RemoveOSDs(context *clusterd.Context, clusterInfo *client.ClusterInfo, osdsToRemove []string, preservePVC string, forceRemovalCallback func(osdID int) bool) error {
```

Then the implementation of upstream callers could look at the force flag to decide whether to force remove the OSD. Or if desired (e.g. downstream), the callback could prompt for whether to continue.

subhamkrai: I'm not if I understand the advantage of using the callback function as today also we checked for the force flag and asked for the prompt.

BlaineEXE: I think the intent of the callback is to allow for the `RemoveOSDs()` function to implement interactivity in the callback itself, while the `RemoveOSDs()` routine is running. I must admit though that I think the callback API is confusing. It may be easier to understand to break the API up into 2 parts:

1. SafeToRemoveOSDs( ..., osdsToRemove []string) (bool, error)
2. RemoveOSDs( ..., osdsToRemove []string, preservePVC, forceOSDRemoval bool) error

I think this follows the single responsibility principle better and allows callers more flexibility with how to handle different Ceph states.

travisn: The behavior that we want falls in two categories:
1. If an OSD is not safe to destroy, show a CLI prompt asking them to enter a response such as: `yes-really-destroy-osd`
2. If an OSD is not safe to destroy, destroy it anyway if the force flag is true. 

From an API perspective for readability it could be simpler to avoid the callback. If we go that route, what about this to get the two behaviors above?
1. RemoveOSDsPromptIfNotSafe( ..., osdsToRemove []string) error
2. RemoveOSDsForceIfNotSafe( ..., osdsToRemove []string, forceOSDRemoval bool) error

I would still vote for the callback instead of this, but I could go either way.

subhamkrai: I think that we can close this PR and open the PR on Rook ?

subhamkrai: https://github.com/rook/rook/pull/13281

Issue #211, opened 2023-12-11
Address the unchecked scenarios for stale subvolume command


yati1998:
In reference to the PR: https://github.com/rook/kubectl-rook-ceph/pull/206,
there are few comments which needs to be addressed and some unchecked scenarios:
1. Print the output in proper format
2. include the type(stale/bound) while listing the subvolumes
3. delete the omap values for the stale subvolumes
4. check for pending clones and external ceph cluster presence

Checkout the document for more details:
https://docs.google.com/document/d/1leSMY06IDkQD0weJ62mjgcJXAbppbNgR3u7nU-dSifo/edit

yati1998:
Hey guys, adding a document for the steps that needs to be followed to connect to external cluster. https://docs.google.com/document/d/18fAS4zBgKAEnmESzm38AP2mVgY0AkCgjrlf2OFNoseQ/edit . In my opinion, rather than adding these steps to the script, we can rather add pre-requisites before running the subvolume command in external cluster. Let me know your thoughts on it. cc @travisn @Madhu-1

subhamkrai:
@yati1998 what's the status on this?

subhamkrai:
do we need to keep this issue open or closing this one and opening new issue for formatting sounds good

> yes, its done. Just a little formatting is required

PR #233, opened 2024-02-02
Allow to delete subvolume with retain-snapshot feature

yati1998:
This PR introduces the following enhancement to the stale subvolume cleanup process:
* Allow to delete the stale subvolumes that have snapshots with retain-snapshot feature
* Correct the order of arguments passed while deleting the subvolume
* Allow users to delete only one subvolume at a time.

### DIFF
--- a/cmd/commands/subvolume.go
+++ b/cmd/commands/subvolume.go
@@ -40,17 +40,19 @@ var listCmd = &cobra.Command{
 }
 
 var deleteCmd = &cobra.Command{
-	Use:                "delete",
-	Short:              "Deletes a stale subvolume.",
-	DisableFlagParsing: true,
-	Args:               cobra.ExactArgs(3),
-	Example:            "kubectl rook-ceph delete <subvolumes> <filesystem> <subvolumegroup>",
+	Use:     "delete",
+	Short:   "Deletes a stale subvolume.",
+	Args:    cobra.RangeArgs(2, 3),
+	Example: "kubectl rook-ceph delete <filesystem> <subvolume> [subvolumegroup]",
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx := cmd.Context()
-		subList := args[0]
-		fs := args[1]
-		svg := args[2]
-		subvolume.Delete(ctx, clientSets, operatorNamespace, cephClusterNamespace, subList, fs, svg)
+		fs := args[0]
+		subvol := args[1]
+		svg := "csi"
+		if len(args) > 2 {
+			svg = args[2]
+		}
+		subvolume.Delete(ctx, clientSets, operatorNamespace, cephClusterNamespace, fs, subvol, svg)
 	},
 }
 
--- a/docs/subvolume.md
+++ b/docs/subvolume.md
@@ -9,12 +9,11 @@ and delete them without impacting other resources and attached volumes.
 The subvolume command will require the following sub commands:
 * `ls` : [ls](#ls) lists all the subvolumes
     * `--stale`: lists only stale subvolumes
-* `delete <subvolumes> <filesystem> <subvolumegroup>`:
-    [delete](#delete) stale subvolumes as per user's input.
-    It will list and delete only the stale subvolumes to prevent any loss of data.
-    * subvolumes: comma-separated list of subvolumes of same filesystem and subvolumegroup.
-    * filesystem: filesystem name to which the subvolumes belong. 
-    * subvolumegroup: subvolumegroup name to which the subvolumes belong.
+* `delete <filesystem> <subvolume> [subvolumegroup]`:
+    [delete](#delete) a stale subvolume.
+    * subvolume: subvolume name.
+    * filesystem: filesystem name to which the subvolume belongs.
+    * subvolumegroup: subvolumegroup name to which the subvolume belong(default is "csi")
 ## ls
 
 ```bash
@@ -23,8 +22,8 @@ kubectl rook-ceph subvolume ls
 # Filesystem  Subvolume  SubvolumeGroup  State
 # ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi in-use
 # ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110006 csi in-use
 # ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale-with-snapshot
 
 ```
 
@@ -33,23 +32,15 @@ kubectl rook-ceph subvolume ls --stale
 
 # Filesystem  Subvolume  SubvolumeGroup state
 # ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi stale
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi stale
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi stale-with-snapshot
 
 ```
 
 ## delete
 
 ```bash
-kubectl rook-ceph subvolume delete csi-vol-427774b4-340b-11ed-8d66-0242ac110004 ocs-storagecluster csi
+kubectl rook-ceph subvolume delete ocs-storagecluster csi-vol-427774b4-340b-11ed-8d66-0242ac110004
 
-# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
-
-```
-
-```bash
-kubectl rook-ceph subvolume delete csi-vol-427774b4-340b-11ed-8d66-0242ac110004,csi-vol-427774b4-340b-11ed-8d66-0242ac110005 ocs-storagecluster csi
-
-# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
-# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
+# Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110004" deleted
 
 ```
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
@@ -34,9 +33,16 @@ type fsStruct struct {
 }
 
 type subVolumeInfo struct {
-	svg string
-	fs  string
+	svg   string
+	fs    string
+	state string
 }
+
+const (
+	inUse             = "in-use"
+	stale             = "stale"
+	staleWithSnapshot = "stale-with-snapshot"
+)
 
 func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, includeStaleOnly bool) {
 
@@ -52,7 +58,7 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	}
 	subvolumeNames := make(map[string]subVolumeInfo)
 	for _, pv := range pvList.Items {
-		if pv.Spec.CSI != nil && pv.Spec.CSI.VolumeAttributes["subvolumeName"] != "" {
+		if pv.Spec.CSI != nil {
 			subvolumeNames[pv.Spec.CSI.VolumeAttributes["subvolumeName"]] = subVolumeInfo{}
 		}
 	}
@@ -90,26 +96,59 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
+				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+
 				// Assume the volume is stale unless proven otherwise
-				stale := true
+				stalevol := true
 				// lookup for subvolume in list of the PV references
 				_, ok := subvolumeNames[sv.Name]
-				if ok || checkSnapshot(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name) {
-					// The volume is not stale if a PV was found, or it has a snapshot
-					stale = false
+				if ok {
+					// The volume is not stale if a PV was found
+					stalevol = false
 				}
-				status := "stale"
-				if !stale {
+				status := stale
+				if !stalevol {
 					if includeStaleOnly {
 						continue
 					}
-					status = "in-use"
+					status = inUse
+				} else {
+					// check the state of the stale subvolume
+					// if it is snapshot-retained then skip listing it.
+					if state == "snapshot-retained" {
+						status = state
+						continue
+					}
+					// check if the stale subvolume has snapshots.
+					if checkSnapshot(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name) {
+						status = staleWithSnapshot
+					}
+
 				}
-				subvolumeNames[sv.Name] = subVolumeInfo{fs.Name, svg.Name}
+				subvolumeNames[sv.Name] = subVolumeInfo{fs.Name, svg.Name, state}
 				fmt.Println(fs.Name, sv.Name, svg.Name, status)
 			}
 		}
 	}
+}
+
+// getSubvolumeState returns the state of the subvolume
+func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
+	subVolumeInfo, errvol := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup}, operatorNamespace, clusterNamespace, true)
+	if errvol != nil {
+		logging.Error(errvol, "failed to get filesystems")
+		return ""
+	}
+	var info map[string]interface{}
+	err := json.Unmarshal([]byte(subVolumeInfo), &info)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to unmarshal: %q", err))
+	}
+	state, ok := info["state"].(string)
+	if !ok {
+		logging.Fatal(fmt.Errorf("failed to get the state of subvolume: %q", SubVol))
+	}
+	return state
 }
 
 // gets list of filesystem
@@ -166,36 +205,13 @@ func unMarshaljson(list string) []fsStruct {
 	return unmarshal
 }
 
-func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subList, fs, svg string) {
-	subvollist := strings.Split(subList, ",")
+func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvol, svg string) {
 	k8sSubvolume := getK8sRefSubvolume(ctx, clientsets)
-	for _, subvolume := range subvollist {
-		check := checkStaleSubvolume(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg, k8sSubvolume)
-		if check {
-			_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvolume, svg}, OperatorNamespace, CephClusterNamespace, true)
-			if err != nil {
-				logging.Error(err, "failed to delete stale subvolume %q", subvolume)
-				continue
-			}
-			logging.Info("subvolume %q deleted", subvolume)
-		} else {
-			logging.Info("subvolume %q is not stale", subvolume)
-		}
+	_, check := k8sSubvolume[subvol]
+	if !check {
+		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
+		logging.Info("subvolume %q deleted", subvol)
+	} else {
+		logging.Info("subvolume %q is not stale", subvol)
 	}
-}
-
-// checkStaleSubvolume checks if there are any stale subvolume to be deleted
-func checkStaleSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg string, k8sSubvolume map[string]subVolumeInfo) bool {
-	_, ok := k8sSubvolume[subvolume]
-	if !ok {
-		snapshot := checkSnapshot(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg)
-		if snapshot {
-			logging.Error(fmt.Errorf("subvolume %s has snapshots", subvolume))
-			return false
-		} else {
-			return true
-		}
-	}
-	logging.Error(fmt.Errorf("Subvolume %s is referenced by a PV", subvolume))
-	return false
 }
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -79,9 +79,11 @@ jobs:
         run: |
           set -ex
           kubectl rook-ceph ceph fs subvolume create myfs test-subvol group-a
+          kubectl rook-ceph ceph fs subvolume create myfs test-subvol-1 group-a
           kubectl rook-ceph subvolume ls
           kubectl rook-ceph subvolume ls --stale
-          kubectl rook-ceph subvolume delete test-subvol myfs group-a
+          kubectl rook-ceph subvolume delete myfs test-subvol group-a
+          kubectl rook-ceph subvolume delete myfs test-subvol-1
 
       - name: Get mon endpoints
         run: |
@@ -234,10 +236,12 @@ jobs:
       - name: Subvolume command
         run: |
           set -ex
+          kubectl rook-ceph --operator-namespace test-operator -n test-cluster ceph fs subvolume create myfs test-subvol-1 group-a
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster ceph fs subvolume create myfs test-subvol group-a
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume ls
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume ls --stale
-          kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete test-subvol myfs group-a
+          kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs test-subvol group-a
+          kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs test-subvol-1 
 
       - name: Get mon endpoints
         run: |


cmd/commands/subvolume.go
```diff
@@ -40,22 +40,23 @@ var listCmd = &cobra.Command{
...
 var deleteCmd = &cobra.Command{
-	Use:                "delete",
-	Short:              "Deletes a stale subvolume.",
-	DisableFlagParsing: true,
-	Args:               cobra.ExactArgs(3),
-	Example:            "kubectl rook-ceph delete <subvolumes> <filesystem> <subvolumegroup>",
+	Use:     "delete",
+	Short:   "Deletes a stale subvolume.",
+	Args:    cobra.ExactArgs(3),
+	Example: "kubectl rook-ceph delete <subvolumes> <filesystem> <subvolumegroup>",
```
Madhu-1, line 46: You need to swap the filesystem and subvolume, IMHO allowing one at a time also fine.

cmd/commands/subvolume.go
```diff
@@ -40,22 +40,23 @@ var listCmd = &cobra.Command{
...
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx := cmd.Context()
 		subList := args[0]
 		fs := args[1]
 		svg := args[2]
-		subvolume.Delete(ctx, clientSets, operatorNamespace, cephClusterNamespace, subList, fs, svg)
+		retainSnapshots, _ := cmd.Flags().GetBool("retain-snapshots")
```
Madhu-1, line 52: why do we need to have this flag?

cmd/commands/subvolume.go
```diff
@@ -40,22 +40,23 @@ var listCmd = &cobra.Command{
...
+		subvolume.Delete(ctx, clientSets, operatorNamespace, cephClusterNamespace, subList, fs, svg, retainSnapshots)
 	},
 }
 
 func init() {
 	SubvolumeCmd.AddCommand(listCmd)
 	SubvolumeCmd.PersistentFlags().Bool("stale", false, "List only stale subvolumes")
 	SubvolumeCmd.AddCommand(deleteCmd)
+	SubvolumeCmd.PersistentFlags().Bool("retain-snapshots", false, "Delete subvolume but retain the snapshot")
```
Madhu-1, line 61: This should be the default behaviour, we dont need to have a new flag from user, because you can never delete a subvolume which is having snapshots without `retain-snapshot` flag in ceph CLI.

docs/subvolume.md
```diff
@@ -23,8 +24,9 @@ kubectl rook-ceph subvolume ls
 # Filesystem  Subvolume  SubvolumeGroup  State
 # ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi in-use
 # ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110006 csi in-use
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110006 csi snapshot-retained
```
Madhu-1, line 27: You can skip listing subvolume which is already deleted but retained only due to snapshot, WDYT?

pkg/filesystem/subvolume.go
```diff
@@ -53,7 +53,9 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	subvolumeNames := make(map[string]subVolumeInfo)
 	for _, pv := range pvList.Items {
 		if pv.Spec.CSI != nil && pv.Spec.CSI.VolumeAttributes["subvolumeName"] != "" {
-			subvolumeNames[pv.Spec.CSI.VolumeAttributes["subvolumeName"]] = subVolumeInfo{}
+			if pv.Spec.CSI.VolumeAttributes["subvolumeName"] != "" {
```
Madhu-1, line 56: in which case the subvolumeName can be empty?

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,35 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
 
+// getSubVolumeState returns the state of the subvolume
+func getSubVolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
+	subVolumeInfo := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup}, operatorNamespace, clusterNamespace, true, false)
+	var info map[string]interface{}
+	err := json.Unmarshal([]byte(subVolumeInfo), &info)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to unmarshal: %q", err))
+	}
+	return info["state"].(string)
```
Madhu-1, line 125: typecast with `ok` flag or else this can panic

pkg/filesystem/subvolume.go
```diff
@@ -143,31 +168,25 @@ func unMarshaljson(list string) []fsStruct {
...
-		} else {
-			logging.Info("subvolume %q is not stale", subvolume)
-		}
+	check := checkStaleSubvolume(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs, subvol, svg, k8sSubvolume)
+	if check && !retainSnapshots {
+		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg}, OperatorNamespace, CephClusterNamespace, true, false)
+		logging.Info("subvolume %q deleted", subvol)
+	} else if check && retainSnapshots {
+		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true, false)
+		logging.Info("subvolume %q deleted: snapshot: %q", subvol, retainSnapshots)
```
Madhu-1, line 179: we dont need to consider `--retain-snapshots` as a special case, any subvolume can be deleted with this flag.

travisn (commented):
Please rebase to pick up the recent CI fix.

cmd/commands/subvolume.go
yati1998: ohh yeah, forgot to change here.

cmd/commands/subvolume.go
yati1998: The idea was to list the subvolmes along with the status(if it has snapshots or not), so user can pass this flag and delete the stale subvolumes, even if it has snapshots.

docs/subvolume.md
yati1998: Yeah seems a better option to be. I will update it.

yati1998:
@mergifyio rebase

docs/subvolume.md
```diff
@@ -9,12 +9,13 @@ and delete them without impacting other resources and attached volumes.
 The subvolume command will require the following sub commands:
 * `ls` : [ls](#ls) lists all the subvolumes
     * `--stale`: lists only stale subvolumes
-* `delete <subvolumes> <filesystem> <subvolumegroup>`:
+* `delete <filesystem> <subvolumes> <subvolumegroup>`:
     [delete](#delete) stale subvolumes as per user's input.
     It will list and delete only the stale subvolumes to prevent any loss of data.
-    * subvolumes: comma-separated list of subvolumes of same filesystem and subvolumegroup.
+    * subvolumes: subvolume name.
     * filesystem: filesystem name to which the subvolumes belong.
```
subhamkrai, line 16:
```suggestion
    * filesystem: filesystem name to which the subvolumes belong.
```

pkg/filesystem/subvolume.go
```diff
@@ -79,12 +81,14 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
+				state := getSubVolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
```
subhamkrai, line 84:
```suggestion
				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
```

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,41 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
-				subvolumeNames[sv.Name] = subVolumeInfo{fs.Name, svg.Name}
+				subvolumeNames[sv.Name] = subVolumeInfo{fs.Name, svg.Name, state}
 				fmt.Println(fs.Name, sv.Name, svg.Name, status)
 			}
 		}
 	}
 }
 
+// getSubVolumeState returns the state of the subvolume
+func getSubVolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
```
subhamkrai, line 120:
```suggestion
func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
```

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,41 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
+	err := json.Unmarshal([]byte(subVolumeInfo), &info)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to unmarshal: %q", err))
+	}
+	state, ok := info["state"].(string)
+	if !ok {
+		logging.Error(fmt.Errorf("failed to get the state of subvol: %q", SubVol))
+	}
+	return state
+
```
subhamkrai, line 132:
```suggestion
```

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,41 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
+func getSubVolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
+	subVolumeInfo := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup}, operatorNamespace, clusterNamespace, true, false)
+	var info map[string]interface{}
+	err := json.Unmarshal([]byte(subVolumeInfo), &info)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to unmarshal: %q", err))
+	}
+	state, ok := info["state"].(string)
+	if !ok {
+		logging.Error(fmt.Errorf("failed to get the state of subvol: %q", SubVol))
```
subhamkrai, line 129:
```suggestion
		logging.Error(fmt.Errorf("failed to get the state of subvolume: %q", SubVol))
```

subhamkrai (changes requested):

cmd/commands/subvolume.go
Madhu-1: IMHO we dont need any flag this should be the default behaviour

docs/subvolume.md
```diff
@@ -9,12 +9,13 @@ and delete them without impacting other resources and attached volumes.
 The subvolume command will require the following sub commands:
 * `ls` : [ls](#ls) lists all the subvolumes
     * `--stale`: lists only stale subvolumes
-* `delete <subvolumes> <filesystem> <subvolumegroup>`:
+* `delete <filesystem> <subvolumes> <subvolumegroup>`:
```
Madhu-1, line 12:
```suggestion
* `delete <filesystem> <subvolume> <subvolumegroup>`:
```

docs/subvolume.md
```diff
@@ -9,12 +9,13 @@ and delete them without impacting other resources and attached volumes.
...
     [delete](#delete) stale subvolumes as per user's input.
```
Madhu-1, line 13:
```suggestion
[delete](#delete) stale subvolume as per user's input.
```

pkg/filesystem/subvolume.go
```diff
@@ -143,31 +174,22 @@ func unMarshaljson(list string) []fsStruct {
...
+		logging.Info("subvolume %q deleted", subvol)
+	} else {
+		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
 
 // checkStaleSubvolume checks if there are any stale subvolume to be deleted
 func checkStaleSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg string, k8sSubvolume map[string]subVolumeInfo) bool {
 	_, ok := k8sSubvolume[subvolume]
 	if !ok {
```
Madhu-1, line 191: just return `ok` from this function? and remove all the input parameters and can you change function name to `isStableSubvolume()`

cmd/commands/subvolume.go
```diff
@@ -40,11 +40,10 @@ var listCmd = &cobra.Command{
...
 var deleteCmd = &cobra.Command{
-	Use:                "delete",
-	Short:              "Deletes a stale subvolume.",
-	DisableFlagParsing: true,
-	Args:               cobra.ExactArgs(3),
-	Example:            "kubectl rook-ceph delete <subvolumes> <filesystem> <subvolumegroup>",
+	Use:     "delete",
+	Short:   "Deletes a stale subvolume.",
+	Args:    cobra.ExactArgs(3),
+	Example: "kubectl rook-ceph delete <filesystem> <subvolume> <subvolumegroup>",
```
travisn, line 46: It will be most intuitive if we specify the arguments from highest level to lowest level. How about subvolume last?
```suggestion
	Example: "kubectl rook-ceph delete <filesystem> <subvolumegroup> <subvolume>",
```

cmd/commands/subvolume.go
```diff
@@ -40,11 +40,10 @@ var listCmd = &cobra.Command{
 }
 
 var deleteCmd = &cobra.Command{
-	Use:                "delete",
-	Short:              "Deletes a stale subvolume.",
-	DisableFlagParsing: true,
-	Args:               cobra.ExactArgs(3),
-	Example:            "kubectl rook-ceph delete <subvolumes> <filesystem> <subvolumegroup>",
```
travisn, line 47: Was deleting multiple subvolumes working before? Is that still a goal?

travisn (changes requested):
Please note the CI failures

cmd/commands/subvolume.go
yati1998: this is the usual format that we follow to create and remove subvolume, it would be good in my opinion to keep it as it is.

cmd/commands/subvolume.go
yati1998: yes it was working before, but as @Madhu-1 mentioned, it's better to follow single deletion at a time.

yati1998:
> Please note the CI failures

````
Run set -ex
+ set -ex
+ kubectl rook-ceph rbd ls replicapool
Warning: rook version 'rook: v1.13.0-alpha.0.220.gf5fcf[9](https://github.com/rook/kubectl-rook-ceph/actions/runs/7781414479/job/21215804058?pr=233#step:11:10)6e0' is running a pre-release version of Rook.```

It says failure while listing the rbd pools, doesn't seems to be related to my changes. Any idea on what's the issue?

File: cmd/commands/subvolume.go (hunk 16 above)
Comment (travisn) [reply cmd/commands/subvolume.go:46]: We don't create it in the tool or in rook, I assume just in the csi driver? Sounds good to keep it consistent, can you just show an example of the creation for my curiosity?

File: cmd/commands/subvolume.go (hunk 17 above)
Comment (travisn) [reply cmd/commands/subvolume.go:47]: The only way it really would make sense to support deletion of multiple subvolumes would be if the subvolumes were the last param, but I agree single deletion is sufficient.

**Review by travisn — COMMENTED** [review]
> > Please note the CI failures
> 
> ```
> Run set -ex
> + set -ex
> + kubectl rook-ceph rbd ls replicapool
> Warning: rook version 'rook: v1.13.0-alpha.0.220.gf5fcf[9](https://github.com/rook/kubectl-rook-ceph/actions/runs/7781414479/job/21215804058?pr=233#step:11:10)6e0' is running a pre-release version of Rook.```
> 
> It says failure while listing the rbd pools, doesn't seems to be related to my changes. Any idea on what's the issue?
> ```

Looks like no OSDs were created, it seems related to other CI issues we are seeing related to #234. You might try a force push to see if a fresh run might succeed.

**Comment by yati1998** [comment]
https://github.com/rook/kubectl-rook-ceph/blob/master/.github/workflows/go-test.yaml#L81 
cc @travisn

File: cmd/commands/subvolume.go (hunk 16 above)
Comment (Madhu-1) [reply cmd/commands/subvolume.go:46]:
```bash
sh-4.4$ ceph fs subvolume create
Invalid command: missing required parameter vol_name(<string>)
fs subvolume create <vol_name> <sub_name> [<size:int>] [<group_name>] [<pool_layout>] [<uid:int>] [<gid:int>] [<mode>] [--namespace-isolated] :  Create a CephFS subvolume in a volume, and optionally, with a specific size (in bytes), a specific data pool layout, a specific mode, in a specific subvolume group and in separate RADOS namespace
```

@travisn this is how it looks in ceph CLI

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,40 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 						continue
 					}
 					status = "in-use"
+				} else {
+					// check if the stale subvolume has snapshots.
+					if checkSnapshot(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name) {
+						status = "stale and has snapshot"
```
Madhu-1, line 103: can we have some pre-defined states in CLI output? the current one looks to be of more detailed one.

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,40 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
+					}
+					// check the state of the stale subvolume
+					// if it is snapshot-retained then skip listing it.
+					if state == "snapshot-retained" {
+						status = state
+						continue
+					}
```
Madhu-1, line 110: Move this above, we dont need to check for snapshots in its a retained one.

pkg/filesystem/subvolume.go
```diff
@@ -93,14 +97,40 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
+func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
+	subVolumeInfo := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup}, operatorNamespace, clusterNamespace, true, false)
+	var info map[string]interface{}
+	err := json.Unmarshal([]byte(subVolumeInfo), &info)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to unmarshal: %q", err))
+	}
+	state, ok := info["state"].(string)
+	if !ok {
+		logging.Error(fmt.Errorf("failed to get the state of subvolume: %q", SubVol))
```
Madhu-1, line 129: need to log it as `Fatal` isnt it?

pkg/filesystem/subvolume.go
yati1998: what pre-defined states do you mean? the one in the subvolume info just says it's completed in this state. We are displaying the detailed state as the user could know that the subvolume has snapshot before deleting it.

Madhu-1: You need to have some const for it like below

```go

const (
InUse = "In-Use"
Stale = "Stale"
...
) and use it

Comment (yati1998) [reply pkg/filesystem/subvolume.go:103]: ohh okay, misunderstood, My bad. Cool will update it.

File: cmd/commands/subvolume.go (hunk 16 above)
Comment (travisn) [inline cmd/commands/subvolume.go:46]: Can the `subvolumegroup` be optional and we default to `csi`?

File: pkg/filesystem/subvolume.go [hunk 21]
```diff
@@ -34,10 +33,17 @@ type fsStruct struct {
...
-	fs  string
+	svg   string
+	fs    string
+	state string
 }
 
+const (
+	InUse            = "In-Use"
+	Stale            = "Stale"
+	StaleAndSnapshot = "Stale And Has Snapshot"
```
travisn, line 44: If this is printed as a column, I would suggest it's nice to have a single word to make the output more parsable
```suggestion
	StaleAndSnapshot = "Stale-with-Snapshot"
```

cmd/commands/subvolume.go
yati1998: Mostly the subvolumegroup is named csi only, but IMO, let the user pass the subvolumegroup.  What's your opinion on this @Madhu-1 ?

Madhu-1: default to `csi` looks good and make this optional

pkg/filesystem/subvolume.go
yati1998: It was added as a check when there will be no CSI subvolumes, but yeah with your last PR to fix this issue, got to understand the correct check.
if there is a CSI volume, don't think the subvolumename will be empty. I will remove the check, it's not required.

cmd/commands/subvolume.go
```diff
@@ -40,17 +40,13 @@ var listCmd = &cobra.Command{
...
+	Use:     "delete",
+	Short:   "Deletes a stale subvolume.",
+	Args:    cobra.MinimumNArgs(2),
+	Example: "kubectl rook-ceph delete <filesystem> <subvolume> <subvolumegroup>",
```
travisn, line 46: Let's use square brackets around optional args
```suggestion
	Example: "kubectl rook-ceph delete <filesystem> <subvolume> [subvolumegroup]",
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
 The subvolume command will require the following sub commands:
 * `ls` : [ls](#ls) lists all the subvolumes
     * `--stale`: lists only stale subvolumes
-* `delete <subvolumes> <filesystem> <subvolumegroup>`:
-    [delete](#delete) stale subvolumes as per user's input.
-    It will list and delete only the stale subvolumes to prevent any loss of data.
-    * subvolumes: comma-separated list of subvolumes of same filesystem and subvolumegroup.
-    * filesystem: filesystem name to which the subvolumes belong. 
-    * subvolumegroup: subvolumegroup name to which the subvolumes belong.
+* `delete <filesystem> <subvolume> <subvolumegroup>`:
```
travisn, line 12:
```suggestion
* `delete <filesystem> <subvolume> [subvolumegroup]`:
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
...
+    [delete](#delete) stale subvolume as per user's input.
+    It will delete only the stale subvolume to prevent any loss of data.
+    * subvolume: subvolume name.
+    * filesystem: filesystem name to which the subvolume belong.
+    * subvolumegroup(optional): subvolumegroup name to which the subvolume belong.
+      * By default the subvolumegroup is "csi".
```
travisn, line 18:
```suggestion
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
...
 * `ls` : [ls](#ls) lists all the subvolumes
     * `--stale`: lists only stale subvolumes
-* `delete <subvolumes> <filesystem> <subvolumegroup>`:
-    [delete](#delete) stale subvolumes as per user's input.
-    It will list and delete only the stale subvolumes to prevent any loss of data.
-    * subvolumes: comma-separated list of subvolumes of same filesystem and subvolumegroup.
-    * filesystem: filesystem name to which the subvolumes belong. 
-    * subvolumegroup: subvolumegroup name to which the subvolumes belong.
+* `delete <filesystem> <subvolume> <subvolumegroup>`:
+    [delete](#delete) stale subvolume as per user's input.
```
travisn, line 13:
```suggestion
    [delete](#delete) a stale subvolume
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
...
+    It will delete only the stale subvolume to prevent any loss of data.
```
travisn, line 14: How is this preventing loss of data? Can we just keep it simple and remove this line?
```suggestion
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
...
+    * subvolume: subvolume name.
+    * filesystem: filesystem name to which the subvolume belong.
```
travisn, line 16:
```suggestion
    * filesystem: filesystem name to which the subvolume belongs.
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
...
+    * subvolumegroup(optional): subvolumegroup name to which the subvolume belong.
```
travisn, line 17: How about just including the default on this line?
```suggestion
    * subvolumegroup (optional): subvolumegroup name to which the subvolume belong (default is "csi")
```

docs/subvolume.md
```diff
@@ -9,47 +9,47 @@ and delete them without impacting other resources and attached volumes.
...
 
 # Filesystem  Subvolume  SubvolumeGroup  State
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110006 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi In-Use
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi In-Use
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi Stale
+# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi Stale-with-Snapshot
```
travisn, line 28: nit: How about if we make `in-use`, `stale`, and `stale-with-snapshot` all lower case? It seems more consistent with other items.
```suggestion
# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale-with-snapshot
```

travisn (changes requested):

docs/subvolume.md
yati1998: Since the field is option, IMO its good to mention the default value??

docs/subvolume.md
```diff
@@ -33,23 +32,22 @@ kubectl rook-ceph subvolume ls --stale
...
-kubectl rook-ceph subvolume delete csi-vol-427774b4-340b-11ed-8d66-0242ac110004 ocs-storagecluster csi
+kubectl rook-ceph subvolume delete ocs-storagecluster csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi
 
-# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
+# Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110004" deleted
 
 ```
 
 ```bash
-kubectl rook-ceph subvolume delete csi-vol-427774b4-340b-11ed-8d66-0242ac110004,csi-vol-427774b4-340b-11ed-8d66-0242ac110005 ocs-storagecluster csi
```
Madhu-1, line 50: add a lone about to mention that you are deleting without subvolumegroup or else no one will notice it and we will have two delete commands in the doc

pkg/filesystem/subvolume.go
```diff
@@ -143,32 +178,30 @@ func unMarshaljson(list string) []fsStruct {
...
-			logging.Info("subvolume %q is not stale", subvolume)
-		}
+	fs := args[0]
+	subvol := args[1]
+	var svg string
+	if len(args) == 2 {
+		svg = "csi"
+	} else {
+		svg = args[2]
+	}
```
Madhu-1, line 190:
```suggestion
svg:="csi"
if len(args) > 2 {
	svg = args[2]
} 
```

pkg/filesystem/subvolume.go
```diff
@@ -143,32 +178,30 @@ func unMarshaljson(list string) []fsStruct {
...
+
+	check := isStaleSubvolume(subvol, k8sSubvolume)
+
```
Madhu-1, line 193: remove extra line

pkg/filesystem/subvolume.go
```diff
@@ -79,28 +85,57 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
+					}
+					// check if the stale subvolume has snapshots.
+					if checkSnapshot(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name) {
+						status = StaleWithSnapshot
+					}
+
 				}
-				subvolumeNames[sv.Name] = subVolumeInfo{fs.Name, svg.Name}
+				subvolumeNames[sv.Name] = subVolumeInfo{fs.Name, svg.Name, state}
 				fmt.Println(fs.Name, sv.Name, svg.Name, status)
```
Madhu-1, line 118: Println is not a formatted output when we have large number of subvolumes it difficult to read, can you please open an issue to print as formatted like `kubectl output`

Madhu-1 (commented):
small nits, LGTM

pkg/filesystem/subvolume.go
```diff
@@ -34,10 +33,17 @@ type fsStruct struct {
...
-	fs  string
+	svg   string
+	fs    string
+	state string
 }
 
+const (
+	InUse             = "in-use"
+	Stale             = "stale"
+	StaleWithSnapshot = "stale-with-snapshot"
```
Madhu-1, line 44: no need to be exported variables as they are not used anywhere else outside this file?

pkg/filesystem/subvolume.go
yati1998: https://github.com/rook/kubectl-rook-ceph/issues/211 , it is already added here, I will check on how we can have a proper format for it.

docs/subvolume.md
```diff
@@ -9,12 +9,11 @@ and delete them without impacting other resources and attached volumes.
...
-    [delete](#delete) stale subvolumes as per user's input.
-    It will list and delete only the stale subvolumes to prevent any loss of data.
-    * subvolumes: comma-separated list of subvolumes of same filesystem and subvolumegroup.
-    * filesystem: filesystem name to which the subvolumes belong. 
-    * subvolumegroup: subvolumegroup name to which the subvolumes belong.
+* `delete <filesystem> <subvolume> [subvolumegroup]`:
+    [delete](#delete) a stale subvolume.
+    * subvolume: subvolume name.
+    * filesystem: filesystem name to which the subvolume belongs.
+    * subvolumegroup(optional): subvolumegroup name to which the subvolume belong(default is "csi")
```
travisn, line 16: nit: optional is implied by the default
```suggestion
    * subvolumegroup: subvolumegroup name to which the subvolume belong (default is "csi")
```

pkg/filesystem/subvolume.go
```diff
@@ -143,32 +178,26 @@ func unMarshaljson(list string) []fsStruct {
...
 	k8sSubvolume := getK8sRefSubvolume(ctx, clientsets)
-	for _, subvolume := range subvollist {
-		check := checkStaleSubvolume(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg, k8sSubvolume)
-		if check {
-			exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvolume, svg}, OperatorNamespace, CephClusterNamespace, true, false)
-			logging.Info("subvolume %q deleted", subvolume)
-		} else {
-			logging.Info("subvolume %q is not stale", subvolume)
-		}
+	fs := args[0]
```
travisn, line 183: Let's keep the args evaluation in the caller, instead of inside this method. So I'd suggest to move these lines 183-188 back inside `cmd/commands/subvolume.go`. This keeps the `Delete()` as a proper API, and all CLI processing is done in the `cmd` package.

cmd/commands/subvolume.go
```diff
@@ -40,17 +40,19 @@ var listCmd = &cobra.Command{
...
 
 var deleteCmd = &cobra.Command{
-	Use:                "delete",
-	Short:              "Deletes a stale subvolume.",
-	DisableFlagParsing: true,
-	Args:               cobra.ExactArgs(3),
-	Example:            "kubectl rook-ceph delete <subvolumes> <filesystem> <subvolumegroup>",
+	Use:     "delete",
+	Short:   "Deletes a stale subvolume.",
+	Args:    cobra.MinimumNArgs(2),
```
subhamkrai, line 45:
```suggestion
	Args:    cobra.RangeArgs(2,3),
```

pkg/filesystem/subvolume.go
```diff
@@ -166,36 +205,20 @@ func unMarshaljson(list string) []fsStruct {
...
+		logging.Info("subvolume %q deleted", subvol)
+	} else {
+		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
 
-// checkStaleSubvolume checks if there are any stale subvolume to be deleted
-func checkStaleSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg string, k8sSubvolume map[string]subVolumeInfo) bool {
+// isStaleSubvolume checks if there are any stale subvolume to be deleted
+func isStaleSubvolume(subvolume string, k8sSubvolume map[string]subVolumeInfo) bool {
```
subhamkrai, line 220:
```suggestion
func isStaleSubvolume(subvolume string, k8sSubvolume map[string]subVolumeInfo) bool {
```

do we need this function? I see this is a one-liner method and only called once. So, will suggested to not include in new function

subhamkrai (dismissed):

yati1998:
@travisn @subhamkrai @Madhu-1 I have addressed all the comments, please do review it once you are free.

pkg/filesystem/subvolume.go
```diff
@@ -166,36 +205,13 @@ func unMarshaljson(list string) []fsStruct {
 	return unmarshal
 }
 
-func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subList, fs, svg string) {
-	subvollist := strings.Split(subList, ",")
+func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvol, svg string) {
 	k8sSubvolume := getK8sRefSubvolume(ctx, clientsets)
```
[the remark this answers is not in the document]
yati1998: It's not about the subvolume being present, this is a check added to make sure we are deleting only stale subvolumes. We are listing all the subvolumes with `ls` command, but still, the user may enter an incorrect subvolume name, which may lead to the deletion of the non-stale subvolume. Hence, it's good to check before directly deleting.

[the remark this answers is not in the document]
travisn: Ok I see now that it is checking to confirm the stale volume. But the perf implication is still concerning. In a test cluster with a handful of PVs it is fine. But what if there are 100 or 1K or 10K PVs in the cluster? How long will this take? Or will these queries for all the PVs even fail if there are too many. I'll open an issue to test it.

travisn (approved):

PR #235, opened 2024-02-05
mon: fetch good mon from socket command

subhamkrai:
let's validate the mon id passed by user based on which
quourm is restored. We don't want to restore quorum which is
not either 'leader' or 'peon'.

Resolves #

**Checklist:**

### DIFF
--- a/docs/mons.md
+++ b/docs/mons.md
@@ -44,6 +44,7 @@ For example, here is the output in a test cluster up to the point of starting th
 
 ```
 $ kubectl rook-ceph mons restore-quorum c
+Info: mon c state is leader
 mon=b, endpoint=192.168.64.168:6789
 mon=c, endpoint=192.168.64.167:6789
 mon=a, endpoint=192.168.64.169:6789
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -95,7 +95,6 @@ func RunCommandInLabeledPod(ctx context.Context, clientsets *k8sutil.Clientsets,
 		return "", fmt.Errorf("failed to get rook mon pod where the command could be executed. %w", err)
 	}
 	var stdout, stderr bytes.Buffer
-
 	err = execCmdInPod(ctx, clientsets, cmd, list.Items[0].Name, container, list.Items[0].Namespace, clusterNamespace, args, &stdout, &stderr, returnOutput)
 	if err != nil {
 		err := fmt.Errorf("failed to run command. %w", err)
@@ -117,13 +116,17 @@ func execCmdInPod(ctx context.Context, clientsets *k8sutil.Clientsets,
 	command, podName, containerName, podNamespace, clusterNamespace string,
 	args []string, stdout, stderr io.Writer, returnOutput bool) error {
 
+	if len(args) < 1 {
+		return fmt.Errorf("no arg passed to exec with %q command", command)
+	}
+
 	cmd := []string{}
 	cmd = append(cmd, command)
 	cmd = append(cmd, args...)
 
 	if containerName == "rook-ceph-tools" {
 		cmd = append(cmd, "--connect-timeout=10")
-	} else if cmd[0] == "ceph" {
+	} else if cmd[0] == "ceph" && args[0] != "daemon" {
 		cmd = append(cmd, "--connect-timeout=10", fmt.Sprintf("--conf=/var/lib/rook/%s/%s.config", clusterNamespace, clusterNamespace))
 	} else if cmd[0] == "rbd" {
 		cmd = append(cmd, fmt.Sprintf("--conf=/var/lib/rook/%s/%s.config", clusterNamespace, clusterNamespace))
--- a/pkg/mons/restore_quorum.go
+++ b/pkg/mons/restore_quorum.go
@@ -18,6 +18,7 @@ package mons
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -34,7 +35,12 @@ import (
 )
 
 func RestoreQuorum(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, goodMon string) {
-	err := restoreQuorum(ctx, clientsets, operatorNamespace, clusterNamespace, goodMon)
+	err := validateMonIsUp(ctx, clientsets, clusterNamespace, goodMon)
+	if err != nil {
+		logging.Fatal(err)
+	}
+
+	err = restoreQuorum(ctx, clientsets, operatorNamespace, clusterNamespace, goodMon)
 	if err != nil {
 		logging.Fatal(err)
 	}
@@ -302,6 +308,31 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 		logging.Info("mon=%s, endpoints=%s\n", monName, monEndpoint)
 	}
 	return badMons, goodMonPublicIp, goodMonPort, nil
+}
+
+func validateMonIsUp(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace, monID string) error {
+	args := []string{"daemon", fmt.Sprintf("mon.%s", monID), "mon_status"}
+	out, err := exec.RunCommandInLabeledPod(ctx, clientsets, fmt.Sprintf("mon=%s", monID), "mon", "ceph", args, clusterNamespace, true)
+	if err != nil {
+		return fmt.Errorf("failed to run command ceph daemon mon.%s mon_status %v", monID, err)
+	}
+
+	type monStatus struct {
+		State string `json:"state"`
+	}
+	var monStatusOut monStatus
+	err = json.Unmarshal([]byte(out), &monStatusOut)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal mon stat output %v", err)
+	}
+
+	logging.Info("mon %q state is %q", monID, monStatusOut.State)
+
+	if monStatusOut.State == "leader" || monStatusOut.State == "peon" {
+		return nil
+	}
+
+	return fmt.Errorf("mon %q in %q state but must be in leader/peon state", monID, monStatusOut.State)
 }
 
 func PromptToContinueOrCancel(expectedAnswer, answer string) error {


pkg/mons/restore_quorum.go
```diff
@@ -215,6 +217,20 @@ func updateMonMap(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNa
...
+	type monStat struct {
+		Leader string `json:"leader"`
+	}
+
+	var monStatOut monStat
+	err := json.Unmarshal([]byte(monStatOutput), &monStatOut)
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to unmarshal mon stat output: %v", err))
+	}
+	return monStatOut.Leader
```
sp98, line 231: Can this be empty?

subhamkrai: I have not seen this value empty, but it's worth adding a check to see if it is empty or not. Thanks

sp98 (approved):

README.md
```diff
@@ -50,15 +50,14 @@ These are args currently supported:
...
 
-
 ### Commands
 
 - `ceph <args>` : Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](https://docs.ceph.com/en/pacific/start/intro/) for more.
 
 - `rbd <args>` : Call a 'rbd' CLI command with arbitrary args
 
 - `mons` : Print mon endpoints
-  - `restore-quorum <mon-name>` : Restore the mon quorum based on a single healthy mon since quorum was lost with the other mons
```
travisn, line 61: I imagine there are scenarios where the `ceph mon stat` would fail, but the user still wants to reset the quorum. I'd suggest instead of removing the mon name param, let's make it optional. Then we run the `ceph mon stat` and if it fails or doesn't match the mon (if they specify it), then we ask if they really want to continue.

travisn (changes requested):

README.md
subhamkrai: I see your point, I have made it optional. Thanks

cmd/commands/mons.go
```diff
@@ -39,12 +39,16 @@ var MonCmd = &cobra.Command{
 // RestoreQuorum represents the mons command
 var RestoreQuorum = &cobra.Command{
 	Use:                "restore-quorum",
-	Short:              "When quorum is lost, restore quorum to the remaining healthy mon",
+	Short:              "When quorum is lost, restore quorum to the remaining healthy mon.Optional arg 'mon-name'",
```
travisn, line 42: I'd suggest keep the message simple, no need to say it's an optional arg.
```suggestion
	Short:              "When quorum is lost, restore quorum to the remaining healthy mon.",
```

docs/mons.md
```diff
@@ -15,127 +15,80 @@ kubectl rook-ceph mons
...
    - The Rook operator log shows timeout errors and continuously fails to reconcile
    - All commands in the toolbox are unresponsive
    - Multiple mon pods are likely down
-2. Identify which mon has good state.
+2. Optionally, when Ceph status is `Error` identify which mon has good state.
    - Exec to a mon pod and run the following command
      - `ceph daemon mon.<name> mon_status`
      - For example, if connecting to mon.a, run: `ceph daemon mon.a mon_status`
-   - If multiple mons respond, find the mon with the highest `election_epoch`
+   - If multiple mons respond, f
```
travisn, line 31:
```suggestion
   - If multiple mons respond, ?
```

pkg/mons/restore_quorum.go
```diff
@@ -215,6 +229,24 @@ func updateMonMap(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNa
 	exec.RunCommandInLabeledPod(ctx, clientsets, labelSelector, "mon", "monmaptool", []string{"--print", monmapPath}, clusterNamespace, false, true)
 }
 
+func fetchGoodMon(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) string {
+	monStatOutput := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"mon", "stat", "--format", "json-pretty"}, operatorNamespace, clusterNamespace, true, false)
```
travisn, line 233: The scenario for resetting the mon quorum is that quorum has been lost. The challenge here is that `ceph mon stat` will hang if quorum is lost, so it won't do much good to call it if quorum is lost. Where did the suggestion to use `ceph mon stat` come from? I just don't see it helping this scenario.

Instead of `ceph mon stat`, perhaps we should connect to a mon pod and query the admin socket as suggested in this comment:
```
To fetch the leader mon run the command `ceph daemon mon.<name> mon_status` inside mon pod and check for the `state: leader`
```

travisn (changes requested):

pkg/mons/restore_quorum.go
subhamkrai: I was working on downstream, and I though, what if happens if someone passes the wrong mon id, it will try to restore based on the wrong mon which is not good mon. Then I was exploring the command and found `ceph mon stat` which has the leader mon. 

do you think we fetch the good mon with `ceph daemon mon.<id> mon_status` command and not depend on the user or this is too much automation?

subhamkrai: agreed, this way we are validating the monID passed by user and also brings less code change.

cmd/commands/mons.go
subhamkrai: not valid anymore.

pkg/mons/restore_quorum.go
```diff
@@ -304,6 +310,29 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
+	err = json.Unmarshal([]byte(out), &monStatusOut)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal mon stat output")
+	}
+
+	if monStatusOut.State == "leader" || monStatusOut.State == "peon" {
+		return nil
+	}
+
+	return fmt.Errorf("mon %s state is not leader/peon", monStatusOut.State)
```
travisn, line 333: Have you observed other states? Seems like if we get any state, at least the mon is responding, and perhaps we can still proceed.

subhamkrai: no I have seen/tried to other states also based on [this](https://docs.ceph.com/en/latest/rados/troubleshooting/troubleshooting-mon/#most-common-monitor-issues) ceph doc

 ```doc
If the Monitor is out of the quorum, then its state will be one of the following: probing, electing or synchronizing. If the state of the Monitor is leader or peon, then the Monitor believes itself to be in quorum but the rest of the cluster believes that it is not in quorum.
 ```
seems like checking this should be the right thing. Thoughts?

pkg/mons/restore_quorum.go
```diff
@@ -304,6 +310,29 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 	return badMons, goodMonPublicIp, goodMonPort, nil
 }
 
+func fetchGoodMon(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace, monID string) error {
```
travisn, line 313:
```suggestion
func validateMonIsUp(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace, monID string) error {
```

pkg/mons/restore_quorum.go
travisn: Got it, based on that doc, checking just for leader/peon sounds good.

pkg/mons/restore_quorum.go
```diff
@@ -304,6 +310,29 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
+		State string `json:"state"`
+	}
+	var monStatusOut monStatus
+	err = json.Unmarshal([]byte(out), &monStatusOut)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal mon stat output")
+	}
+
+	if monStatusOut.State == "leader" || monStatusOut.State == "peon" {
+		return nil
```
travisn, line 330: How about info logging for the state of the mon that was detected?

travisn (changes requested):
just a couple small suggestions

pkg/mons/restore_quorum.go
subhamkrai: done

pkg/mons/restore_quorum.go

docs/mons.md
```diff
@@ -44,6 +44,7 @@ For example, here is the output in a test cluster up to the point of starting th
 
 ```
 $ kubectl rook-ceph mons restore-quorum c
+Info: mon a state is leader
```
travisn, line 47: This example should print for mon `c`, right?
```suggestion
Info: mon c state is leader
```

pkg/mons/restore_quorum.go
```diff
@@ -304,6 +310,31 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
+	type monStatus struct {
+		State string `json:"state"`
+	}
+	var monStatusOut monStatus
+	err = json.Unmarshal([]byte(out), &monStatusOut)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal mon stat output")
+	}
+
+	logging.Info("mon %s state is %s", monID, monStatusOut.State)
```
travisn, line 329: The logs are easier to read vars when quotes around them
```suggestion
	logging.Info("mon %q state is %q", monID, monStatusOut.State)
```

pkg/mons/restore_quorum.go
```diff
@@ -304,6 +310,31 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
+		return fmt.Errorf("failed to run command ceph daemon mon.%s mon_status %v", monID, err)
+	}
+
+	type monStatus struct {
+		State string `json:"state"`
+	}
+	var monStatusOut monStatus
+	err = json.Unmarshal([]byte(out), &monStatusOut)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal mon stat output")
```
Madhu-1, line 326: wrap the error or return the actual error you are getting?

pkg/mons/restore_quorum.go
Madhu-1: add mon id to err as well?

pkg/mons/restore_quorum.go

pkg/mons/restore_quorum.go
subhamkrai: right, missed that added now

pkg/exec/exec.go
```diff
@@ -123,7 +122,7 @@ func execCmdInPod(ctx context.Context, clientsets *k8sutil.Clientsets,
 
 	if containerName == "rook-ceph-tools" {
 		cmd = append(cmd, "--connect-timeout=10")
-	} else if cmd[0] == "ceph" {
+	} else if cmd[0] == "ceph" && args[0] != "daemon" {
```
Madhu-1, line 125: dont we need to check for `args` length before accessing it?

pkg/mons/restore_quorum.go
Madhu-1:
```suggestion
return fmt.Errorf("mon %s in %s state but not in leader/peon state", monID,monStatusOut.State)
```
Do you meant to log like this or similar?

subhamkrai: yeah, your suggestion is more descriptive.

pkg/exec/exec.go
subhamkrai: we don't want to run the command `ceph daemon *` with the extra flags like `--config` which is required for other commands. That's why I have added this check so that command that starts with `ceph daemon * ` should not have the extra flags

Madhu-1: my suggestion was to do len(agrs)>1 and return error if its less than 1 to ensure that no panic happens as we are accessing it.

subhamkrai: I don't why I misinterpreted. We check the length in some places but better to add the check so I added now

pkg/exec/exec.go
```diff
@@ -117,13 +116,17 @@ func execCmdInPod(ctx context.Context, clientsets *k8sutil.Clientsets,
 	command, podName, containerName, podNamespace, clusterNamespace string,
 	args []string, stdout, stderr io.Writer, returnOutput bool) error {
 
+	if len(args) < 1 {
+		return fmt.Errorf("no arg passed to exec with %s command", command)
```
travisn, line 120:
```suggestion
		return fmt.Errorf("no arg passed to exec with %q command", command)
```

pkg/mons/restore_quorum.go
```diff
@@ -304,6 +310,31 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
...
+		return fmt.Errorf("failed to unmarshal mon stat output %v", err)
+	}
+
+	logging.Info("mon %q state is %q", monID, monStatusOut.State)
+
+	if monStatusOut.State == "leader" || monStatusOut.State == "peon" {
+		return nil
+	}
+
+	return fmt.Errorf("mon %q in %q state but not in leader/peon state", monID, monStatusOut.State)
```
travisn, line 335:
```suggestion
	return fmt.Errorf("mon %q in %q state but must be in leader/peon state", monID, monStatusOut.State)
```

travisn (approved):
Approved with a couple small wording suggestions

pkg/mons/restore_quorum.go

PR #237, opened 2024-02-06
clean cephFS omap details

yati1998:
Remove the omap object for the given subvolume.

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -77,11 +77,13 @@ jobs:
         run: |
           set -ex
           kubectl rook-ceph ceph fs subvolume create myfs test-subvol group-a
-          kubectl rook-ceph ceph fs subvolume create myfs test-subvol-1 group-a
           kubectl rook-ceph subvolume ls
           kubectl rook-ceph subvolume ls --stale
           kubectl rook-ceph subvolume delete myfs test-subvol group-a
-          kubectl rook-ceph subvolume delete myfs test-subvol-1
+          tests/github-action-helper.sh create_sc_with_retain_policy
+          tests/github-action-helper.sh create_stale_subvolume
+          subVol=$(kubectl rook-ceph subvolume ls --stale | awk '{print $2}' | grep csi-vol)
+          kubectl rook_ceph subvolume delete myfs $subVol
 
       - name: Get mon endpoints
         run: |
@@ -232,12 +234,14 @@ jobs:
       - name: Subvolume command
         run: |
           set -ex
-          kubectl rook-ceph --operator-namespace test-operator -n test-cluster ceph fs subvolume create myfs test-subvol-1 group-a
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster ceph fs subvolume create myfs test-subvol group-a
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume ls
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume ls --stale
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs test-subvol group-a
-          kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs test-subvol-1
+          tests/github-action-helper.sh create_sc_with_retain_policy_custom_ns test-operator test-cluster
+          tests/github-action-helper.sh create_stale_subvolume
+          subVol=$(kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume ls --stale | awk '{print $2}' | grep csi-vol)
+          kubectl rook_ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs $subVol
 
       - name: Get mon endpoints
         run: |
--- a/docs/subvolume.md
+++ b/docs/subvolume.md
@@ -17,30 +17,37 @@ The subvolume command will require the following sub commands:
 ## ls
 
 ```bash
-kubectl rook-ceph subvolume ls
+$ kubectl rook-ceph subvolume ls
 
-# Filesystem  Subvolume  SubvolumeGroup  State
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale-with-snapshot
-
+Filesystem  Subvolume  SubvolumeGroup  State
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi in-use
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale-with-snapshot
 ```
 
 ```bash
-kubectl rook-ceph subvolume ls --stale
+$ kubectl rook-ceph subvolume ls --stale
 
-# Filesystem  Subvolume  SubvolumeGroup state
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi stale
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi stale-with-snapshot
-
+Filesystem  Subvolume  SubvolumeGroup state
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi stale
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi stale-with-snapshot
 ```
 
 ## delete
 
 ```bash
-kubectl rook-ceph subvolume delete ocs-storagecluster csi-vol-427774b4-340b-11ed-8d66-0242ac110004
+$ kubectl rook-ceph subvolume delete myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005
 
-# Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110004" deleted
+Info: Deleting the omap object and key for subvolume "csi-vol-0c91ba82-5a63-4117-88a4-690acd86cbbd"
+Info: omap object:"csi.volume.0c91ba82-5a63-4117-88a4-690acd86cbbd" deleted
+Info: omap key:"csi.volume.pvc-78abf81c-5381-42ee-8d75-dc17cd0cf5de" deleted
+Info: subvolume "csi-vol-0c91ba82-5a63-4117-88a4-690acd86cbbd" deleted
+```
 
+```bash
+$ kubectl rook-ceph subvolume delete myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+
+Info: No omapvals found for subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110005" deleted
 ```
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
@@ -28,8 +29,9 @@ import (
 )
 
 type fsStruct struct {
-	Name string
-	data []string
+	Name         string
+	MetadataName string `json:"metadata_pool"`
+	data         []string
 }
 
 type subVolumeInfo struct {
@@ -201,7 +203,6 @@ func unMarshaljson(list string) []fsStruct {
 	if errg != nil {
 		logging.Fatal(errg)
 	}
-
 	return unmarshal
 }
 
@@ -209,9 +210,100 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
 	k8sSubvolume := getK8sRefSubvolume(ctx, clientsets)
 	_, check := k8sSubvolume[subvol]
 	if !check {
-		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
-		logging.Info("subvolume %q deleted", subvol)
+		deleteOmapForSubvolume(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to delete subvolume of %s/%s/%s", fs, svg, subvol)
+		}
+		logging.Info("subvolume %s/%s/%s deleted", fs, svg, subvol)
+
 	} else {
-		logging.Info("subvolume %q is not stale", subvol)
+		logging.Info("subvolume %s/%s/%s is not stale", fs, svg, subvol)
 	}
+}
+
+func getMetadataPoolName(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs string) (string, error) {
+	fsstruct, err := getFileSystem(ctx, clientsets, OperatorNamespace, CephClusterNamespace)
+	if err != nil {
+		return "", err
+	}
+
+	for _, pool := range fsstruct {
+		if pool.Name == fs {
+			return pool.MetadataName, nil
+		}
+	}
+	return "", fmt.Errorf("metadataPool not found for %q filesystem", fs)
+}
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmapForSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
+	logging.Info("Deleting the omap object and key for subvolume %q", subVol)
+	omapkey := getOmapKey(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs)
+	omapval := getOmapVal(subVol)
+	poolName, err := getMetadataPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	if omapval != "" {
+		// remove omap object.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to remove omap object for subvolume %q", subVol)
+		}
+		logging.Info("omap object:%q deleted", omapval)
+
+	}
+	if omapkey != "" {
+		// remove omap key.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapkey, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to remove omap key for subvolume %q", subVol)
+		}
+		logging.Info("omap key:%q deleted", omapkey)
+
+	}
+}
+
+// getOmapKey gets the omap key and value details for a given subvolume.
+// Deletion of omap object required the subvolumeName which is of format
+// csi.volume.subvolume, where subvolume is the name of subvolume that needs to be
+// deleted.
+// similarly to delete of omap key requires csi.volume.ompakey, where
+// omapkey is the pv name which is extracted the omap object.
+func getOmapKey(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) string {
+
+	poolName, err := getMetadataPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	omapval := getOmapVal(subVol)
+
+	cmd := []string{"getomapval", omapval, "csi.volname", "-p", poolName, "--namespace", "csi", "/dev/stdout"}
+	pvname, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", cmd, OperatorNamespace, CephClusterNamespace, true)
+	if err != nil || pvname == "" {
+		logging.Info("No PV found for subvolume %s: %s", subVol, err)
+		return ""
+	}
+	// omap key is for format csi.volume.pvc-fca205e5-8788-4132-979c-e210c0133182
+	// hence, attaching pvname to required prefix.
+	omapkey := "csi.volume." + pvname
+
+	return omapkey
+}
+
+// func getOmapVal is used to get the omapval from the given subvolume
+// omapval is of format csi.volume.427774b4-340b-11ed-8d66-0242ac110005
+// which is similar to volume name csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+// hence, replacing 'csi-vol-' to 'csi.volume.' to get the omapval
+func getOmapVal(subVol string) string {
+
+	splitSubvol := strings.SplitAfterN(subVol, "-", 3)
+	if len(splitSubvol) < 3 {
+		return ""
+	}
+	subvol_id := splitSubvol[len(splitSubvol)-1]
+	omapval := "csi.volume." + subvol_id
+
+	return omapval
 }
--- a/pkg/filesystem/subvolume_test.go
+++ b/pkg/filesystem/subvolume_test.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2024 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subvolume
+
+import "testing"
+
+func TestGetOmapVal(t *testing.T) {
+
+	tests := []struct {
+		name string
+		want string
+	}{
+		{
+			name: "csi-vol-427774b4-340b-11ed-8d66-0242ac110005",
+			want: "csi.volume.427774b4-340b-11ed-8d66-0242ac110005",
+		},
+		{
+			name: "nfs-export-427774b4-340b-11ed-8d66-0242ac110005",
+			want: "csi.volume.427774b4-340b-11ed-8d66-0242ac110005",
+		},
+		{
+			name: "",
+			want: "",
+		},
+		{
+			name: "csi-427774b4-340b-11ed-8d66-0242ac11000",
+			want: "csi.volume.340b-11ed-8d66-0242ac11000",
+		},
+		{
+			name: "csi-427774b440b11ed8d660242ac11000",
+			want: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := getOmapVal(tt.name); got != tt.want {
+				t.Errorf("getOmapVal() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/tests/github-action-helper.sh
+++ b/tests/github-action-helper.sh
@@ -76,6 +76,37 @@ deploy_rook_in_custom_namespace() {
   deploy_csi_driver_custom_ns "$1" "$2"
 }
 
+create_sc_with_retain_policy(){
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml -o storageclass.yaml
+  sed -i "s|name: rook-cephfs|name: rook-cephfs-retain|g" storageclass.yaml
+  sed -i "s|reclaimPolicy: Delete|reclaimPolicy: Retain|g" storageclass.yaml
+  kubectl create -f storageclass.yaml 
+}
+
+create_sc_with_retain_policy_custom_ns(){
+  export OPERATOR_NS=$1
+  export CLUSTER_NS=$2
+
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml -o storageclass.yaml
+  sed -i "s|name: rook-cephfs|name: rook-cephfs-retain|g" storageclass.yaml
+  sed -i "s|reclaimPolicy: Delete|reclaimPolicy: Retain|g" storageclass.yaml
+  sed -i "s|provisioner: rook-ceph.cephfs.csi.ceph.com |provisioner: test-operator.cephfs.csi.ceph.com |g" storageclass.yaml
+  deploy_with_custom_ns $OPERATOR_NS $CLUSTER_NS storageclass.yaml
+}
+
+create_stale_subvolume() {
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml -o pvc.yaml
+  sed -i "s|name: cephfs-pvc|name: cephfs-pvc-retain|g" pvc.yaml
+  sed -i "s|storageClassName: rook-cephfs|storageClassName: rook-cephfs-retain|g" pvc.yaml
+  kubectl create -f pvc.yaml
+  kubectl get pvc cephfs-pvc-retain
+  : "${PVNAME:=$(kubectl get pvc cephfs-pvc-retain -o=jsonpath='{.spec.volumeName}')}"
+  wait_for_pvc_to_be_bound_state_default
+  kubectl get pvc cephfs-pvc-retain
+  kubectl delete pvc cephfs-pvc-retain
+  kubectl delete pv "$PVNAME"
+}
+
 deploy_with_custom_ns() {
   sed -i "s|rook-ceph # namespace:operator|$1 # namespace:operator|g" "$3"
   sed -i "s|rook-ceph # namespace:cluster|$2 # namespace:cluster|g" "$3"
@@ -107,6 +138,16 @@ deploy_csi_driver_custom_ns() {
   curl -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/subvolumegroup.yaml -o subvolumegroup.yaml
   deploy_with_custom_ns "$1" "$2" subvolumegroup.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml
+}
+
+wait_for_pvc_to_be_bound_state_default() {
+  timeout 100 bash <<-'EOF'
+    until [ $(kubectl get pvc cephfs-pvc-retain -o jsonpath='{.status.phase}') == "Bound" ]; do
+      echo "waiting for the pvc to be in bound state"
+      sleep 1
+    done
+EOF
+  timeout_command_exit_code
 }
 
 wait_for_pod_to_be_ready_state_default() {
--- a/pkg/exec/exec.go
+++ b/pkg/exec/exec.go
@@ -130,6 +130,8 @@ func execCmdInPod(ctx context.Context, clientsets *k8sutil.Clientsets,
 		cmd = append(cmd, "--connect-timeout=10", fmt.Sprintf("--conf=/var/lib/rook/%s/%s.config", clusterNamespace, clusterNamespace))
 	} else if cmd[0] == "rbd" {
 		cmd = append(cmd, fmt.Sprintf("--conf=/var/lib/rook/%s/%s.config", clusterNamespace, clusterNamespace))
+	} else if cmd[0] == "rados" {
+		cmd = append(cmd, fmt.Sprintf("--conf=/var/lib/rook/%s/%s.config", clusterNamespace, clusterNamespace))
 	}
 
 	// Prepare the API URL used to execute another process within the Pod.  In


yati1998:
```Error: command terminated with exit code 1
did not load config file, using default settings.
2024-02-06T10:08:53.016+0000 7fa11474ae00 -1 Errors while parsing config file!
2024-02-06T10:08:53.016+0000 7fa11474ae00 -1 can't open ceph.conf: (2) No such file or directory
unable to get monitor info from DNS SRV with service name: ceph-mon
2024-02-06T10:08:53.018+0000 7fa11474ae00 -1 failed for service _ceph-mon._tcp
2024-02-06T10:08:53.018+0000 7fa11474ae00 -1 monclient: get_monmap_and_config cannot identify monitors to contact
failed to fetch mon config (--no-mon-config to skip)```

**Comment by yati1998** [comment]
cc: @travisn @Madhu-1

**Comment by travisn** [comment]
> https://github.com/rook/kubectl-rook-ceph/actions/runs/7799941151/job/21272736952?pr=237#step:12:20
> 
> ```
> did not load config file, using default settings.
> 2024-02-06T10:08:53.016+0000 7fa11474ae00 -1 Errors while parsing config file!
> 2024-02-06T10:08:53.016+0000 7fa11474ae00 -1 can't open ceph.conf: (2) No such file or directory
> unable to get monitor info from DNS SRV with service name: ceph-mon
> 2024-02-06T10:08:53.018+0000 7fa11474ae00 -1 failed for service _ceph-mon._tcp
> 2024-02-06T10:08:53.018+0000 7fa11474ae00 -1 monclient: get_monmap_and_config cannot identify monitors to contact
> failed to fetch mon config (--no-mon-config to skip)```
> ```

The plugin may not understand that it needs to append the mon and cluster commands to the `rados` command. Looks like you need to add the `rados` command [here](https://github.com/rook/kubectl-rook-ceph/blob/8a410dfd141c6e120f9ea5e2250b7b85a2452728/pkg/exec/exec.go#L109-L115).

**Comment by yati1998** [comment]
> > https://github.com/rook/kubectl-rook-ceph/actions/runs/7799941151/job/21272736952?pr=237#step:12:20
> > ```
> > did not load config file, using default settings.
> > 2024-02-06T10:08:53.016+0000 7fa11474ae00 -1 Errors while parsing config file!
> > 2024-02-06T10:08:53.016+0000 7fa11474ae00 -1 can't open ceph.conf: (2) No such file or directory
> > unable to get monitor info from DNS SRV with service name: ceph-mon
> > 2024-02-06T10:08:53.018+0000 7fa11474ae00 -1 failed for service _ceph-mon._tcp
> > 2024-02-06T10:08:53.018+0000 7fa11474ae00 -1 monclient: get_monmap_and_config cannot identify monitors to contact
> > failed to fetch mon config (--no-mon-config to skip)```
> > ```
> 
> The plugin may not understand that it needs to append the mon and cluster commands to the `rados` command. Looks like you need to add the `rados` command [here](https://github.com/rook/kubectl-rook-ceph/blob/8a410dfd141c6e120f9ea5e2250b7b85a2452728/pkg/exec/exec.go#L109-L115).

That works @travisn . Thanks.

File: .github/workflows/go-test.yaml [hunk 1]
```diff
@@ -77,7 +77,7 @@ jobs:
 
       - name: Subvolume command
         run: |
-          set -ex
+          set -e
```
subhamkrai, line 80:
```suggestion
          set -e
```
?

pkg/filesystem/subvolume.go
```diff
@@ -178,12 +183,28 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
 			logging.Info("subvolume %q deleted", subvolume)
+			DeleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvolume, fs)
 		} else {
 			logging.Info("subvolume %q is not stale", subvolume)
 		}
 	}
 }
 
+// DeleteOmap deletes omap object and key for the given subvolume.
+func DeleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
```
subhamkrai, line 194:
```suggestion
func deleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
```

pkg/filesystem/subvolume.go
```diff
@@ -178,12 +183,28 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
 				continue
 			}
 			logging.Info("subvolume %q deleted", subvolume)
+			DeleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvolume, fs)
```
subhamkrai, line 186:
```suggestion
			deleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvolume, fs)
```

pkg/filesystem/subvolume.go
```diff
@@ -178,12 +183,28 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
 			logging.Info("subvolume %q is not stale", subvolume)
 		}
 	}
 }
 
+// DeleteOmap deletes omap object and key for the given subvolume.
+func DeleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
+	omapdetails, err := getOmapDetails(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs)
+	if err != nil {
+		logging.Fatal(err)
```
subhamkrai, line 197:
```suggestion
		logging.Fatal(err)
```

it's better to good error message along with error

subhamkrai (commented):
Also, please update commit/PR title

travisn:
The CI is failing:

```
+ kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs test-subvol group-a
Info: subvolume "test-subvol" deleted
Error: error getting omap keys myfs-metadata/test.subvolume: (2) No such file or directory
. failed to run command. command terminated with exit code 1%!(EXTRA string=failed to list omapval)
Error: failed to get omap details: "error getting omap keys myfs-metadata/test.subvolume: (2) No such file or directory\n. failed to run command. command terminated with exit code 1"
```

yati1998:
> The CI is failing:
> 
> ```
> + kubectl rook-ceph --operator-namespace test-operator -n test-cluster subvolume delete myfs test-subvol group-a
> Info: subvolume "test-subvol" deleted
> Error: error getting omap keys myfs-metadata/test.subvolume: (2) No such file or directory
> . failed to run command. command terminated with exit code 1%!(EXTRA string=failed to list omapval)
> Error: failed to get omap details: "error getting omap keys myfs-metadata/test.subvolume: (2) No such file or directory\n. failed to run command. command terminated with exit code 1"
> ```

Yes, this error is expected. Updated the test to include error in output.

.github/workflows/go-test.yaml
```diff
@@ -77,7 +77,7 @@ jobs:
 
       - name: Subvolume command
         run: |
-          set -ex
+          set +e -x
```
travisn, line 80: This change is causing other tests to fail.

pkg/filesystem/subvolume.go
```diff
@@ -44,6 +45,11 @@ const (
 	staleWithSnapshot = "stale-with-snapshot"
 )
 
+type OmapDetails struct {
```
Madhu-1, line 48: why this need to be exported?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,73 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
 		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
 		logging.Info("subvolume %q deleted", subvol)
+		deleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
 	} else {
 		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
```
Madhu-1, line 227: what will happen to the omap if the subvolumes are deleted but the omap are not deleted/cleaned up?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,73 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+			return false
+		} else {
+			return true
+		}
+	}
+	logging.Error(fmt.Errorf("Subvolume %s is referenced by a PV", subvolume))
+	return false
+}
+
+func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) (OmapDetails, error) {
```
Madhu-1, line 257: add function comments exactly what are we doing on each helper function as there are string operations it is highly difficult to understand what we are doing?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,73 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+	var omapval1, pv string
+	poolName := fs + "-metadata"
+	omapval := strings.Replace(subVol, "-", ".", 2)
+	checknfs := strings.Contains(subVol, "nfs-export")
+	if checknfs {
+		omapval1 = strings.Replace(omapval, "export", "volume", 1)
+		omapval1 = strings.Replace(omapval1, "nfs", "csi", 1)
+	} else {
+		omapval1 = strings.Replace(omapval, "vol", "volume", 1)
+
```
Madhu-1, line 269: extra line

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,73 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+		}
+	}
+	logging.Error(fmt.Errorf("Subvolume %s is referenced by a PV", subvolume))
+	return false
+}
+
+func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) (OmapDetails, error) {
+
+	var details OmapDetails
+	var omapval1, pv string
```
Madhu-1, line 260: use meaningful variable names, IMO we should not use `omapval1`

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,73 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+
+	ls, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"listomapvals", omapval1, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+	if err != nil {
+		logging.Error(err, "failed to list omapval")
+		return OmapDetails{}, err
+	}
+	if ls != "" {
+		volname := strings.Split(ls, "|")
+		if checknfs {
+			pv = "csi.volume." + volname[12] + volname[14] + volname[16]
```
Madhu-1, line 280: do we have length check for `volname` to ensure we don't panic, can add comments for this kind of operations

Madhu-1 (changes requested):
i think we should start testing all these functionalities in the CI to ensure it works, Please add test cases for it

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,73 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+	// remove omap object.
+	exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapdetails.omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+	logging.Info("omap object:%q deleted", omapdetails.omapval)
+	// remove omap key.
+	exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapdetails.omapkey, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+	logging.Info("omap key:%q deleted", omapdetails.omapkey)
+}
+
+// checkStaleSubvolume checks if there are any stale subvolume to be deleted
+func checkStaleSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg string, k8sSubvolume map[string]subVolumeInfo) bool {
```
Madhu-1, line 242: dont we already have a helper for this one? i though we added one in the last PR.

docs/subvolume.md
```diff
@@ -40,16 +40,20 @@ kubectl rook-ceph subvolume ls --stale
...
 ```bash
-kubectl rook-ceph subvolume delete csi-vol-427774b4-340b-11ed-8d66-0242ac110004,csi-vol-427774b4-340b-11ed-8d66-0242ac110005 ocs-storagecluster csi
+kubectl rook-ceph subvolume delete ocs-storagecluster-cephfilesystem csi-vol-test csi
 
-# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
-# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
+# Info: subvolume "csi-vol-test" deleted
+# Error: error getting omap keys ocs-storagecluster-cephfilesystem-metadata/test.subvolume:
+# (2) No such file or directory
+# . failed to run command. command terminated with exit code 1%!(EXTRA string=failed to list omapval)
```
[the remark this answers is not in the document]
yati1998: Incase the subvolume is created from backend and it's stale, there will be no omap created for it.  In that case if we try to delete the omap, it will through this error.

pkg/filesystem/subvolume.go
yati1998: not required, will update it.

pkg/filesystem/subvolume.go
yati1998: this is the case when someone deletes the subvolume from the backend. Here we are deleting the stale subvolume and just after that removing it's entry from Omap. 
Here, we are deleting omap of only stale subvolumes, as that's the input we are asking from the user.  In case we  want to cleanup the omap to remove all the stale omaps whose subvolume and pv both are deleted , we need to have separate command, where it will list all the omaps and pv's and verify the stale omap.

pkg/filesystem/subvolume.go
yati1998: in last PR it was removed as not required, maybe after rebase it's considered as a new change, I will remove it.

docs/subvolume.md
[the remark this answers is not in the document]
subhamkrai: let's document the happy case output.

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,97 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+	_, ok := k8sSubvolume[subvolume]
+	if !ok {
+		snapshot := checkSnapshot(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs, subvolume, svg)
+		if snapshot {
+			logging.Error(fmt.Errorf("subvolume %s has snapshots", subvolume))
+			return false
+		} else {
+			return true
+		}
```
subhamkrai, line 251:
```suggestion
		} 
		return true
```
else is dead code here since we have return in if

subhamkrai (changes requested):

yati1998:
@travisn @Madhu-1 I have addressed all the comments above, please do have a look once you are free.
@Madhu-1 about adding the test for stale rados omap, we need to create sc with retain feature and a few other steps need to be followed, I am working on it to check how we can add that here in ci.

pkg/filesystem/subvolume.go
yati1998: will open and link one.

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,89 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
 		logging.Info("subvolume %q deleted", subvol)
+		deleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
 	} else {
 		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
+	omapdetails := getOmapDetails(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs)
```
travisn, line 228: At the start of this method, how about adding an info log that omap details are to be deleted?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,89 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+	checknfs := strings.Contains(subVol, "nfs-export")
+	// check if it is cephfs or nfs subvolume
+	if checknfs {
+		omapval = strings.Replace(replace, "export", "volume", 1)
+		omapval = strings.Replace(omapval, "nfs", "csi", 1)
+	} else {
+		omapval = strings.Replace(replace, "vol", "volume", 1)
+
+	}
+	ls, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"listomapvals", omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
```
travisn, line 263: Will the omap values be in the same subvolumegroup as the pv that was deleted, or will it always be in "csi"?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,89 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+	if err != nil {
+		if !strings.Contains(err.Error(), "No such file or directory") {
+			logging.Fatal(fmt.Errorf("failed to list omapvals: %q", err))
+		} else {
```
travisn, line 267: No need for "else" since the previous condition will exit after: `logging.Fatal()`
```suggestion
		}
```

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,89 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+		//csi.volname
+		//value (40 bytes) :
+		//00000000  70 76 63 2d 65 63 62 62  34 34 38 61 2d 30 35 37  |pvc-ecbb448a-057|
+		//00000010  36 2d 34 35 30 35 2d 62  32 39 35 2d 30 32 39 33  |6-4505-b295-0293|
+		//00000020  34 37 37 36 34 65 61 64                           |47764ead|
+		//0000002
+		volname := strings.Split(ls, "|")
+		if checknfs {
+			if len(volname) > 16 {
+				pv = "csi.volume." + volname[12] + volname[14] + volname[16]
```
travisn, line 292: This looks very brittle. If anything changes in the output of the command or if the volume name could be a different length, this can easily break. Is there no way to get the csi.imagename without the binary output?

travisn (changes requested):

pkg/filesystem/subvolume.go
yati1998: I am not sure, I initially tried to get the value for csi.imagename only, will see if there is a way possible

pkg/filesystem/subvolume.go
yati1998: But what if the condition is not true and there are no omap details, in that case, it will execute the else statement. I have tested this condition.

travisn: As a general pattern, it is simpler to have fewer code blocks. If we return from an `if` block, there is no need to have an `else` block. For example, this code is the same without the `else` block, because `logging.Fatal()` inside the `if` exits the program immediately.
```go
	if err != nil {
		if !strings.Contains(err.Error(), "No such file or directory") {
			// exit immediately
			logging.Fatal(fmt.Errorf("failed to list omapvals: %q", err))
		}
		logging.Info("No omapvals found for subvolume %s", subVol)
		return omapDetails{}
	}
```
But since it's equivalent in functionality, it's just a nit.

yati1998:
@subhamkrai https://github.com/rook/kubectl-rook-ceph/actions/runs/8060623451/job/22031771871?pr=237#step:22:79 any idea why the ci is failing?

tests/github-action-helper.sh
```diff
@@ -31,9 +31,21 @@ deploy_rook() {
...
+  sed -i "s|reclaimPolicy: Delete|reclaimPolicy: Retain|g" storageclass.yaml
+  kubectl create -f storageclass.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/pvc.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml -o pvc.yaml
+  sed -i "s|name: cephfs-pvc|name: cephfs-pvc-retain|g" pvc.yaml
+  sed -i "s|storageClassName: rook-cephfs|storageClassName: rook-cephfs-retain|g" pvc.yaml
+  kubectl create -f pvc.yaml
+  kubectl delete pvc cephfs-pvc-retain
+  "${pvname:=$(kubectl get pv | grep cephfs-pvc-retain | awk '{print $1}')}"
```
subhamkrai, line 47:
```suggestion
  PVNAME:=$(kubectl get pv | grep cephfs-pvc-retain | awk '{print $1}')
```
or 
```
:  "${PVNAME:=$(kubectl get pv | grep cephfs-pvc-retain | awk '{print $1}')}"
```

yati1998: Initially I tried with the first suggestion, but even that is failing, will go for the next.

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+		} else {
+			if len(volname) > 11 {
+				pv = "csi.volume." + volname[7] + volname[9] + volname[11]
+			}
+		}
+
+		details = omapDetails{omapval, pv}
+	} else {
+		logging.Fatal(fmt.Errorf("omap details not found"))
+	}
```
subhamkrai, line 304: let's keep
```
if ls == "" {
 logging.Fatal(fmt.Errorf("omap details not found"))
} else {
 happy path code
}

it's better to keep the condition first which return the code

File: tests/github-action-helper.sh [hunk 21]
```diff
@@ -31,9 +31,24 @@ deploy_rook() {
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/filesystem-test.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/subvolumegroup.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/toolbox.yaml
+  kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/storageclass-test.yaml
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml -o storageclass.yaml
+  sed -i "s|name: rook-cephfs|name: rook-cephfs-retain|g" storageclass.yaml
+  sed -i "s|reclaimPolicy: Delete|reclaimPolicy: Retain|g" storageclass.yaml
+  kubectl create -f storageclass.yaml
```
subhamkrai, line 39:
```suggestion
  
  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml -o storageclass.yaml
  sed -i "s|name: rook-cephfs|name: rook-cephfs-retain|g" storageclass.yaml
  sed -i "s|reclaimPolicy: Delete|reclaimPolicy: Retain|g" storageclass.yaml
  kubectl create -f storageclass.yaml
  
```

When not' directly creating let's add a space to make it more readable

tests/github-action-helper.sh
```diff
@@ -31,9 +31,24 @@ deploy_rook() {
...
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml -o pvc.yaml
+  sed -i "s|name: cephfs-pvc|name: cephfs-pvc-retain|g" pvc.yaml
+  sed -i "s|storageClassName: rook-cephfs|storageClassName: rook-cephfs-retain|g" pvc.yaml
+  kubectl create -f pvc.yaml
+  kubectl get pvc cephfs-pvc-retain
+  wait_for_pvc_to_be_bound_state_default
+  : "${PVNAME:=$(kubectl get pvc cephfs-pvc-retain -o=jsonpath='{.spec.volumeName}')}"
+  kubectl get pvc cephfs-pvc-retain
+  kubectl delete pvc cephfs-pvc-retain
+  kubectl delete pv "$PVNAME"
```
subhamkrai, line 51:
```suggestion

  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml -o pvc.yaml
  sed -i "s|name: cephfs-pvc|name: cephfs-pvc-retain|g" pvc.yaml
  sed -i "s|storageClassName: rook-cephfs|storageClassName: rook-cephfs-retain|g" pvc.yaml
  kubectl create -f pvc.yaml

  kubectl get pvc cephfs-pvc-retain
  wait_for_pvc_to_be_bound_state_default
  : "${PVNAME:=$(kubectl get pvc cephfs-pvc-retain -o=jsonpath='{.spec.volumeName}')}"
  kubectl get pvc cephfs-pvc-retain
  kubectl delete pvc cephfs-pvc-retain
  kubectl delete pv "$PVNAME"
```

subhamkrai (changes requested):

pkg/filesystem/subvolume.go
BlaineEXE: Up to subham's preference, a form that I see more often in golang code is this, which keeps code easier to read in the happy-path case:

```go
if ls == "" {
    logging.Fatal(/* blah */)
    return omapDetails{} // early return
}
// happy path on the left
```

subhamkrai: so in our code
`logging.Fatal `: log error and return
`logging.Error`: just return the error

BlaineEXE: Oh, I even missed the `Fatal` part. I think that means the `else` isn't necessary from your initial example.

subhamkrai: right

pkg/filesystem/subvolume.go
nixpanic: A much cleaner way is to write to a temporary file:
```console
sh-5.1$ rados --pool ocs-storagecluster-cephblockpool getomapval csi.volume.df42d7a3-8f3a-4630-835e-ed264a088777 csi.imagename /tmp/imagename 
Writing to /tmp/imagename
sh-5.1$ grep '' /tmp/imagename
csi-vol-df42d7a3-8f3a-4630-835e-ed264a088777
```
Note that the contents of `/tmp/imagename` does not contain a `\n` at the end of the line.

You can also write to `/dev/stdout` so that you do not need to create/remove a temporary file:
```console
sh-5.1$ IMAGENAME=$(rados --pool ocs-storagecluster-cephblockpool getomapval csi.volume.df42d7a3-8f3a-4630-835e-ed264a088777 csi.imagename /dev/stdout 2> /dev/null)
sh-5.1$ echo ${IMAGENAME}
csi-vol-df42d7a3-8f3a-4630-835e-ed264a088777
```

A big _**NO**_ against the parsing of the output the way you do now, sorry!

yati1998: Yeah Yeah., I am working on it. I was bit busy testing the ci tests, will work on other comments too.

.github/workflows/go-test.yaml
```diff
@@ -82,6 +82,8 @@ jobs:
           kubectl rook-ceph subvolume ls --stale
           kubectl rook-ceph subvolume delete myfs test-subvol group-a
           kubectl rook-ceph subvolume delete myfs test-subvol-1
+          SubVol=$(kubectl rook-ceph subvolume ls --stale | awk '{print $2}' | grep csi-vol)
```
Madhu-1, line 85:
```suggestion
         subVol=$(kubectl rook-ceph subvolume ls --stale | awk '{print $2}' | grep csi-vol)
```

docs/subvolume.md
```diff
@@ -39,8 +39,18 @@ kubectl rook-ceph subvolume ls --stale
...
 
-# Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110004" deleted
+# Info: subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110004 deleted
+# Info: omap object:"csi.volume.427774b4-340b-11ed-8d66-0242ac110005" deleted
+# Info: omap key:"csi.volume.pvc-fca205e5-8788-4132-979c-e210c0133182" deleted
+
+```
+
+```bash
+kubectl rook-ceph subvolume delete ocs-storagecluster-cephfilesystem csi-vol-test
```
Madhu-1, line 51: `csi-vol-test` is not a proper format use `csi-vol-427774b4-340b-11ed-8d66-0242ac112345`

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
 }
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs, svg string) {
+	logging.Info("Deleting the omap object and key for subvolume %q", subVol)
+	omapdetails := getOmapDetails(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs, svg)
+	poolName := fs + "-metadata"
+	if omapdetails.omapval != "" {
+		// remove omap object.
+		exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapdetails.omapval, "-p", poolName, "--namespace", svg}, OperatorNamespace, CephClusterNamespace, true)
```
Madhu-1, line 233: `--namespace` is always `csi` we cannot pass svg value to it

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
 	} else {
 		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs, svg string) {
+	logging.Info("Deleting the omap object and key for subvolume %q", subVol)
+	omapdetails := getOmapDetails(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs, svg)
+	poolName := fs + "-metadata"
```
Madhu-1, line 230: AFAIK this is not the right way, we need to get the metadata pool name from the filesystem CR

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+	if omapdetails.omapval != "" {
+		// remove omap object.
+		exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapdetails.omapval, "-p", poolName, "--namespace", svg}, OperatorNamespace, CephClusterNamespace, true)
+		logging.Info("omap object:%q deleted", omapdetails.omapval)
+	}
+	if omapdetails.omapkey != "" {
+		// remove omap key.
+		exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapdetails.omapkey, "-p", poolName, "--namespace", svg}, OperatorNamespace, CephClusterNamespace, true)
```
Madhu-1, line 238: `exec.RunCommandInOperatorPod` can fail due to other reason we should not skip error check and consider all the cases as success isn't it?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+		logging.Info("omap key:%q deleted", omapdetails.omapkey)
+	}
+}
+
+// getOmapDetails gets the omap key and value details for a given subvolume.
+// Deletion of omap object required the rbd-image-name which is of format
```
Madhu-1, line 244: `rbd-image-name`?

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+// deleted.
+// similarly to delete of omap key requires csi.volume.ompakey, where
+// omapkey is the pv name which is extracted the omap object.
+func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs, svg string) omapDetails {
+
+	var details omapDetails
+	var omapval, pv string
+	poolName := fs + "-metadata"
+	replace := strings.Replace(subVol, "-", ".", 2)
+	checknfs := strings.Contains(subVol, "nfs-export")
```
Madhu-1, line 255: i think this assumption is wrong, anyone can create subvolume for nfs without this as well :)

pkg/filesystem/subvolume.go
```diff
@@ -211,7 +217,90 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
...
+			}
+		} else {
+			if len(volname) > 11 {
+				pv = "csi.volume." + volname[7] + volname[9] + volname[11]
+			}
+		}
+
+		details = omapDetails{omapval, pv}
+	} else {
+		logging.Fatal(fmt.Errorf("omap details not found"))
```
Madhu-1, line 303: This also seems also not correct, if not found should be considered as success as it might have been already deleted.

Madhu-1 (changes requested):

pkg/filesystem/subvolume.go
yati1998: Will send nfs changes in another PR. Currently, removing it from this comit to make it simpler.

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,100 @@ func unMarshaljson(list string) []fsStruct {
...
+	// which is similar to volume name csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+	// hence, replacing '-' to '.' to get the omapval
+	replace := strings.Replace(subVol, "-", ".", 2)
+	omapval = strings.Replace(replace, "vol", "volume", 1)
+
+	cmd := []string{"getomapval", omapval, "csi.volname", "-p", poolName, "--namespace", "csi", "/dev/stdout"}
+	pvname, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", cmd, OperatorNamespace, CephClusterNamespace, true)
+	if err != nil || pvname == "" {
+		if !strings.Contains(err.Error(), "No such file or directory") {
+			logging.Fatal(fmt.Errorf("failed to list omapvals: %q", err))
```
nixpanic, line 296: You are not listing omap values anymore, you now only get a single one. Include the missing value and error?

Is this really fatal? What if the values were cleaned up some way already?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,100 @@ func unMarshaljson(list string) []fsStruct {
...
 	return unmarshal
 }
 
 func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvol, svg string) {
 	k8sSubvolume := getK8sRefSubvolume(ctx, clientsets)
 	_, check := k8sSubvolume[subvol]
 	if !check {
-		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
+		deleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
```
travisn, line 219: Why retain snapshots? If we're deleting the stale subvolumes, seems like their snapshots are also not needed.

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,100 @@ func unMarshaljson(list string) []fsStruct {
...
+		if err != nil {
+			logging.Error(err, "failed to delete subvolume of %q/%q", fs, subvol)
+		}
```
travisn, line 222: The logging.Info needs to be in the else block
```suggestion
		} else {
```

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,100 @@ func unMarshaljson(list string) []fsStruct {
...
 		logging.Info("subvolume %q deleted", subvol)
```
travisn, line 223: All logging statements that format an arg need to use the `f` method
```suggestion
		logging.Infof("subvolume %q deleted", subvol)
```

travisn (changes requested):
@yati1998 Please review https://github.com/rook/rook/issues/13869 for the discussion about cleaning up subvolumegroups and omap, perhaps you have some input there, thanks.

pkg/filesystem/subvolume.go
yati1998: if snapshots are present we can't simply delete the subvolumes, we first need to delete the snapshots and then the subvolume. And I am not very sure above deleting the snapshots from admin side. Hence, best way to delete the subvolume would be with retain-snapshots.
cc @nixpanic  @Madhu-1 what do you think?

pkg/filesystem/subvolume.go
yati1998: ahh, missed this.

pkg/filesystem/subvolume.go
travisn: Oh nevermind, we implemented this differently in this repo from Rook. :)

pkg/filesystem/subvolume.go
nixpanic: Ceph-CSI uses a reference counter in OMAP, I think... Only when the reference counter is 0, the subvolume/snapshot should be deleted. Maybe that is only for snapshots, not sure it is used for (plain) subvolumes too.

travisn: What is the behavior of `--retain-snapshots`? If the subvolume is being deleted, are the snapshots somehow still useful? I thought the point of this feature was to clean up all the stale resources, and snapshots seem stale in this case too.

Madhu-1: Yes snapshots are meant for backup , even if the subvolume is stale for some reason, the user can restore from the snapshots later

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
 }
 
 func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs, subvol, svg string) {
 	k8sSubvolume := getK8sRefSubvolume(ctx, clientsets)
 	_, check := k8sSubvolume[subvol]
 	if !check {
-		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
-		logging.Info("subvolume %q deleted", subvol)
+		deleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
```
Madhu-1, line 219: first delete the subvolume and than delete the rados object?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+		if err != nil {
+			logging.Error(err, "failed to delete subvolume of %q/%q", fs, subvol)
```
Madhu-1, line 221: log should be fetal isnt it?
log in format of `fsName/groupName/subVolName`

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+		} else {
+			logging.Info("subvolume %q deleted", subvol)
```
Madhu-1, line 223: follow same format for all the cases

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
 	} else {
 		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
+
+func getPoolName(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs string) (string, error) {
+	// getFilesystem gets the filesystem
+	fsstruct, err := getFileSystem(ctx, clientsets, OperatorNamespace, CephClusterNamespace)
+	if err != nil {
+		logging.Error(err, "failed to get filesystem")
```
Madhu-1, line 234: lets not log in the internal functions, log it from the caller

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+			logging.Error(err, "failed to delete subvolume of %q/%q", fs, subvol)
+		} else {
+			logging.Info("subvolume %q deleted", subvol)
+		}
 	} else {
 		logging.Info("subvolume %q is not stale", subvol)
 	}
 }
+
+func getPoolName(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs string) (string, error) {
```
Madhu-1, line 230: it should be get metadatPool Name not just pool name isn't it?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+	omapdetails := getOmapDetails(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs)
+	poolName, err := getPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	if omapdetails.omapval != "" {
+		// remove omap object.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapdetails.omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Error(err, "failed to remove omap object for subvolume %q", subVol)
```
Madhu-1, line 259: should be logged as fatal or return from here or else logging.Info at line 261 will also gets logged

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+		}
+		logging.Info("omap object:%q deleted", omapdetails.omapval)
+	}
+	if omapdetails.omapkey != "" {
+		// remove omap key.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapdetails.omapkey, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Error(err, "failed to remove omap key for subvolume %q", subVol)
+		}
+		logging.Info("omap key:%q deleted", omapdetails.omapkey)
```
Madhu-1, line 269: same here as well

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+
+	var details omapDetails
+	var omapval, omapkey string
+	poolName, err := getPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	// omapval is of format csi.volume.427774b4-340b-11ed-8d66-0242ac110005
+	// which is similar to volume name csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+	// hence, replacing '-' to '.' to get the omapval
```
Madhu-1, line 289: instead of replace why not remove `csi-vol-` and use it? and this might not work if someone is using https://github.com/ceph/ceph-csi/blob/c90f7ed7776e16c07a07c62106b7643a2819279e/examples/cephfs/storageclass.yaml#L48 volumePrefix

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+// getOmapDetails gets the omap key and value details for a given subvolume.
+// Deletion of omap object required the subvolumeName which is of format
+// csi.volume.subvolume, where subvolume is the name of subvolume that needs to be
+// deleted.
+// similarly to delete of omap key requires csi.volume.ompakey, where
+// omapkey is the pv name which is extracted the omap object.
+func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) omapDetails {
+
+	var details omapDetails
+	var omapval, omapkey string
```
Madhu-1, line 282: we dont need var declaration right? we can declare these variables where its required.

pkg/filesystem/subvolume.go
yati1998: Incase the subvolume is deleted and rados omap deletion gets stuck, we will loose the info to delete the rados omap too. So from my knowledge and testing I would consider deleting the omap first and then subvolume.

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
 	_, check := k8sSubvolume[subvol]
 	if !check {
-		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
-		logging.Info("subvolume %q deleted", subvol)
+		deleteOmap(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to delete subvolume of %q/%q/%q", fs, svg, subvol)
+		} else {
+			logging.Info("subvolume %q/%q/%q deleted", fs, svg, subvol)
```
travisn, line 223: With the separating slashes, better not to include the quotes or else it will be difficult to read
```suggestion
			logging.Info("subvolume %s/%s/%s deleted", fs, svg, subvol)
```

tests/github-action-helper.sh
```diff
@@ -31,9 +31,27 @@ deploy_rook() {
...
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/storageclass-test.yaml
+
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml -o storageclass.yaml
+  sed -i "s|name: rook-cephfs|name: rook-cephfs-retain|g" storageclass.yaml
+  sed -i "s|reclaimPolicy: Delete|reclaimPolicy: Retain|g" storageclass.yaml
+  kubectl create -f storageclass.yaml
+
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/rbd/pvc.yaml
   kubectl create -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/pvc.yaml -o pvc.yaml
```
travisn, line 44: Any specific tests like purging these stale subvolumes should be done as a new step instead of as part of the basic cluster setup. Please remove these steps here and put them in a new Github action step in go-test.yaml.

travisn: Note that these steps are also currently failing the CI.

tests/github-action-helper.sh
```diff
@@ -66,6 +84,24 @@ deploy_rook_in_custom_namespace() {
 
   curl -f https://raw.githubusercontent.com/rook/rook/master/deploy/examples/toolbox.yaml -o toolbox.yaml
   deploy_with_custom_ns "$1" "$2" toolbox.yaml
+
+  curl https://raw.githubusercontent.com/rook/rook/master/deploy/examples/csi/cephfs/storageclass.yaml -o storageclass.yaml
```
travisn, line 88: Move these steps to a Github action step

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
+// deleted.
+// similarly to delete of omap key requires csi.volume.ompakey, where
+// omapkey is the pv name which is extracted the omap object.
+func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) omapDetails {
+
+	poolName, err := getMetadataPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	// omapval is of format csi.volume.427774b4-340b-11ed-8d66-0242ac110005
```
travisn, line 286: Please refactor lines 286-291 to a separate method and then add a unit test. String parsing is always a case for a unit test.

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,98 @@ func unMarshaljson(list string) []fsStruct {
...
 	}
 }
+
+// getOmapDetails gets the omap key and value details for a given subvolume.
+// Deletion of omap object required the subvolumeName which is of format
+// csi.volume.subvolume, where subvolume is the name of subvolume that needs to be
+// deleted.
+// similarly to delete of omap key requires csi.volume.ompakey, where
+// omapkey is the pv name which is extracted the omap object.
+func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) omapDetails {
```
travisn, line 280: I'd suggest just returning the two strings. No need for a struct. Multiple return values is common in golang.
```suggestion
func getOmapDetails(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) (string, string) {
```

travisn (changes requested):
Have you tested locally in minikube that the actions are working to purge the stale subvolumes? The CI is failing with these steps.

yati1998:
@travisn @subhamkrai @sp98 can anyone please check if there is some issue with rbd command, ci is getting stuck there.
Also @travisn have addressed your comments.

sp98:
> @travisn @subhamkrai @sp98 can anyone please check if there is some issue with rbd command, ci is getting stuck there. Also @travisn have addressed your comments.

in the tmate session the same command is working. Only difference is that it doesn't return anything. But I don't think that should cause the CI to stuck!

``` 
runner@fv-az527-340:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ kubectl rook-ceph  rbd ls replicapool
Warning: rook version 'rook: v1.13.0-alpha.0.389.gc409a6ae4' is running a pre-release version of Rook.

runner@fv-az527-340:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ 
```

travisn:
The subvolume delete step is failing in the CI now. Likely it is related to this error since if the storage class isn't as expected, the PVC stays in pending state.
```
+ kubectl create -f storageclass.yaml
Error from server (AlreadyExists): error when creating "storageclass.yaml": storageclasses.storage.k8s.io "rook-cephfs-retain" already exists
```

pkg/filesystem/subvolume.go
```diff
@@ -44,6 +46,11 @@ const (
 	staleWithSnapshot = "stale-with-snapshot"
 )
 
+type omapDetails struct {
```
travisn, line 49: not used anymore?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +208,106 @@ func unMarshaljson(list string) []fsStruct {
...
+	for _, pool := range fsstruct {
+		if pool.Name == fs {
+			return pool.MetadataName, nil
+		}
 	}
+	return "", fmt.Errorf("pool not found")
+}
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmap(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
```
travisn, line 247:
```suggestion
func deleteOmapFromSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
```

travisn (changes requested):

pkg/filesystem/subvolume.go
yati1998: forgot to remove, thanks.

pkg/filesystem/subvolume.go
yati1998: we are not deleting omap from subvolume. In csi on creation of subvolume, it is stored in omap for record. I don't think this could be a suitable name.

travisn: The omap is only related to this subvolume. There are other types of omaps related to rados namespaces, right? Since this method is only related to subvolumes, the goal is to make the code more readable that it's related to subvolumes. Do you have a different name suggested?

yati1998: got it, will update it

subhamkrai:
```
persistentvolumeclaim "cephfs-pvc-retain" deleted
+ kubectl delete pv ''
error: resource name may not be empty
```

Pv is deleted or not present

yati1998:
```runner@fv-az1154-938:~/work/kubectl-rook-ceph/kubectl-rook-ceph$ oc get pvc
NAME                 STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS      AGE
cephfs-pvc           Pending                                      rook-cephfs       18m
cephfs-pvc-retain1   Pending                                      rook-cephfs       108s
rbd-pvc              Pending                                      rook-ceph-block   18m
```

I see none of the pvcs are getting created and the error being `volumeID already exists`. Seems we need to have some cleanup steps before we create the resources . Anyone has any idea why is it so?
cc @subhamkrai @travisn @Madhu-1

yati1998:
> ```
> persistentvolumeclaim "cephfs-pvc-retain" deleted
> + kubectl delete pv ''
> error: resource name may not be empty
> ```
> 
> Pv is deleted or not present

The pv is not present as the pvc is not in the bound state. And the reason for it in mentioned in my last comment.

pkg/filesystem/subvolume_test.go
```diff
@@ -0,0 +1,43 @@
...
+package subvolume
+
+import "testing"
+
+func TestGetOmapVal(t *testing.T) {
+
+	tests := []struct {
+		name string
+		want string
+	}{
```
nixpanic, line 26: Maybe consider adding negative tests as well? An empty string, a string with only no, one or two `-` in there probably are problematic and cause issues. It may be unlikely that these get passed to the function, but you should be prepared for weird issues in a cleanup process, some values may be unexpectedly missing or somehow got corrupted.

yati1998:
@Madhu-1 waiting for your review here. Please feel free to add your comments.

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,109 @@ func unMarshaljson(list string) []fsStruct {
...
+			logging.Info("subvolume %s/%s/%s deleted", fs, svg, subvol)
+		}
 	} else {
-		logging.Info("subvolume %q is not stale", subvol)
+		logging.Info("subvolume %s/%s/%s is not stale", fs, svg, subvol)
+	}
+}
+
+func getMetadataPoolName(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, fs string) (string, error) {
+	// getFilesystem gets the filesystem
```
Madhu-1, line 226: can we remove this comment?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,109 @@ func unMarshaljson(list string) []fsStruct {
...
+	fsstruct, err := getFileSystem(ctx, clientsets, OperatorNamespace, CephClusterNamespace)
+	if err != nil {
+		return "", err
+	}
+
+	// this iterates over the filesystems and subvolumegroup to get the list of subvolumes that exist
```
Madhu-1, line 232: this comments also need to be removed, its talking about group,subvolumes etc.

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,109 @@ func unMarshaljson(list string) []fsStruct {
...
+	for _, pool := range fsstruct {
+		if pool.Name == fs {
+			return pool.MetadataName, nil
+		}
+	}
+	return "", fmt.Errorf("pool not found")
```
Madhu-1, line 238: return something like `fmt.Errorf("metadataPool not found for %q filesystem",fs)`

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,109 @@ func unMarshaljson(list string) []fsStruct {
...
+}
+
+// deleteOmap deletes omap object and key for the given subvolume.
+func deleteOmapForSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) {
+	logging.Info("Deleting the omap object and key for subvolume %q", subVol)
+	omapkey := getOmapKey(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs)
+	omapval := getOmapVal(subVol)
+	poolName, err := getMetadataPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
```
Madhu-1, line 248: we dont return err as nil and empty PoolName isn't it?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,109 @@ func unMarshaljson(list string) []fsStruct {
...
+// Deletion of omap object required the subvolumeName which is of format
+// csi.volume.subvolume, where subvolume is the name of subvolume that needs to be
+// deleted.
+// similarly to delete of omap key requires csi.volume.ompakey, where
+// omapkey is the pv name which is extracted the omap object.
+func getOmapKey(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespace, CephClusterNamespace, subVol, fs string) string {
+
+	poolName, err := getMetadataPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
```
Madhu-1, line 280: same here as well

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,109 @@ func unMarshaljson(list string) []fsStruct {
...
+	}
+	omapval := getOmapVal(subVol)
+
+	cmd := []string{"getomapval", omapval, "csi.volname", "-p", poolName, "--namespace", "csi", "/dev/stdout"}
+	pvname, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", cmd, OperatorNamespace, CephClusterNamespace, true)
+	if err != nil || pvname == "" {
+		logging.Info("No pvname found for subvolume %s: %s", subVol, err)
```
Madhu-1, line 287:
```suggestion
logging.Info("No PV found for subvolume %s: %s", subVol, err)
```

Madhu-1 (commented):
LGTM,
small nits

pkg/filesystem/subvolume.go
yati1998: yess

pkg/filesystem/subvolume.go
yati1998: yes, it's not required.

Madhu-1 (approved):

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,107 @@ func unMarshaljson(list string) []fsStruct {
...
 	if !check {
-		exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
-		logging.Info("subvolume %q deleted", subvol)
+		deleteOmapForSubvolume(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to delete subvolume of %s/%s/%s", fs, svg, subvol)
+		} else {
+			logging.Info("subvolume %s/%s/%s deleted", fs, svg, subvol)
+		}
```
subhamkrai, line 219:
```suggestion
		} 
			logging.Info("subvolume %s/%s/%s deleted", fs, svg, subvol)
		
```
this else is not required as `logging.Fatal` return and exit the code

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,107 @@ func unMarshaljson(list string) []fsStruct {
...
+	omapval := getOmapVal(subVol)
+	poolName, err := getMetadataPoolName(ctx, clientsets, OperatorNamespace, CephClusterNamespace, fs)
+	if err != nil || poolName == "" {
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	if omapval != "" {
+		// remove omap object.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Error(err, "failed to remove omap object for subvolume %q", subVol)
```
subhamkrai, line 252:
```suggestion
			logging.Error(err, "failed to remove omap object for subvolume %q", subVol)
```
`logging.Error` will only log and not return and exit and since this method doesn't return anything do you think we should use `logging.Fatal`?

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,107 @@ func unMarshaljson(list string) []fsStruct {
...
+		} else {
+			logging.Info("omap object:%q deleted", omapval)
+		}
+	}
+	if omapkey != "" {
+		// remove omap key.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapkey, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Error(err, "failed to remove omap key for subvolume %q", subVol)
```
subhamkrai, line 261:
```suggestion
			logging.Error(err, "failed to remove omap key for subvolume %q", subVol)
```
same here

tests/github-action-helper.sh
```diff
@@ -109,6 +140,15 @@ deploy_csi_driver_custom_ns() {
...
 }
 
+wait_for_pvc_to_be_bound_state_default() {
+  timeout 100 bash <<-'EOF'
+    until [ $(kubectl get pvc cephfs-pvc-retain -o jsonpath='{.status.phase}') == "Bound" ]; do
+      echo "waiting for the pvc to be in bound state"
+      sleep 1
+    done
+EOF
+}
```
subhamkrai, line 150:
```suggestion
EOF
    timeout_command_exit_code
}
```

add above function call to exit if timeout reached

subhamkrai (changes requested):

subhamkrai (changes requested):
Also, since this clean cephfs OMAP please update the commit title and PR title to mention cephFS OMAP

docs/subvolume.md
```diff
@@ -17,30 +17,41 @@ The subvolume command will require the following sub commands:
...
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale-with-snapshot
+Filesystem  Subvolume  SubvolumeGroup  State
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi in-use
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi in-use
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110007 csi stale-with-snapshot
 
 ```
```
subhamkrai, line 28:
```suggestion
```
```

remove extra line

File: docs/subvolume.md [hunk 63]
```diff
@@ -17,30 +17,41 @@ The subvolume command will require the following sub commands:
...
+$ kubectl rook-ceph subvolume ls --stale
 
-# Filesystem  Subvolume  SubvolumeGroup state
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi stale
-# ocs-storagecluster-cephfilesystem csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi stale-with-snapshot
+Filesystem  Subvolume  SubvolumeGroup state
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110004 csi stale
+myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi stale-with-snapshot
 
 ```
```
subhamkrai, line 37: remove extra line

docs/subvolume.md
```diff
@@ -17,30 +17,41 @@ The subvolume command will require the following sub commands:
...
+
+```
+
+```bash
+$ kubectl rook-ceph subvolume delete myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+
+Info: No omapvals found for subvolume csi-vol-427774b4-340b-11ed-8d66-0242ac110005
+Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110005" deleted
 
 ```
```
subhamkrai, line 57:
```suggestion
Info: subvolume "csi-vol-427774b4-340b-11ed-8d66-0242ac110005" deleted
```
```

remove extra line

**Review by subhamkrai — CHANGES_REQUESTED** [review]

File: pkg/filesystem/subvolume.go [hunk 65]
```diff
@@ -201,17 +203,107 @@ func unMarshaljson(list string) []fsStruct {
...
+		}
 	}
+	if omapkey != "" {
+		// remove omap key.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapkey, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to remove omap key for subvolume %q", subVol)
+		} else {
+			logging.Info("omap key:%q deleted", omapkey)
+		}
```
subhamkrai, line 264:
```suggestion
		} 
			logging.Info("omap key:%q deleted", omapkey)
		
```

we have `logging.Fatal` so no need to use else here

pkg/filesystem/subvolume.go
```diff
@@ -201,17 +203,107 @@ func unMarshaljson(list string) []fsStruct {
...
+		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
+	}
+	if omapval != "" {
+		// remove omap object.
+		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		if err != nil {
+			logging.Fatal(err, "failed to remove omap object for subvolume %q", subVol)
+		} else {
+			logging.Info("omap object:%q deleted", omapval)
+		}
```
subhamkrai, line 255:
```suggestion
		} else {
			logging.Info("omap object:%q deleted", omapval)
		}
```
same

subhamkrai (changes requested):

subhamkrai (approved):

PR #238, opened 2024-02-08
test: run rbd command with timeout after mon restore

satoru-takeuchi:

Sometimes running rbd command hangs up just after mon restore. Then {default,custom}-namespace jobs stucks like this:
    
```
Run set -ex
+ set -ex
+ kubectl rook-ceph rbd ls replicapool
```
    
Run this command with timeout to avoid this test flakiness.

**Checklist:**

- [x] CI tests has been updated, if necessary.

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -59,6 +59,20 @@ jobs:
           set -ex
           kubectl rook-ceph --context=$(kubectl config current-context) ceph status
 
+      - name: Rbd command
+        run: |
+          set -ex
+          SUCCEEDED=0
+          for ((i=0;i<30;i++)) ; do
+            timeout 10 kubectl rook-ceph rbd ls replicapool || continue
+            SUCCEEDED=1
+            break
+          done
+          if [ "${SUCCEEDED}" -ne 1 ]; then
+            echo "running rbd command timed out"
+            exit 1
+          fi
+
       - name: Mon restore
         run: |
           set -ex
@@ -69,11 +83,6 @@ jobs:
           tests/github-action-helper.sh wait_for_three_mons rook-ceph
           kubectl -n rook-ceph wait deployment rook-ceph-mon-d --for condition=Available=True --timeout=90s
           kubectl -n rook-ceph wait deployment rook-ceph-mon-e --for condition=Available=True --timeout=90s
-
-      - name: Rbd command
-        run: |
-          set -ex
-          kubectl rook-ceph rbd ls replicapool
 
       - name: Subvolume command
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/go-test.yaml
```diff
@@ -73,7 +73,15 @@ jobs:
       - name: Rbd command
         run: |
           set -ex
-          kubectl rook-ceph rbd ls replicapool
+          SUCCEEDED=0
+          for ((i=0;i<30;i++)) ; do
+            kubectl rook-ceph rbd ls replicapool --timeout=10 || continue
```
subhamkrai, line 78:
```suggestion
           timeout 30 kubectl rook-ceph rbd ls replicapool --timeout=10 || continue
```

just adding this works?

subhamkrai (changes requested):

.github/workflows/go-test.yaml
satoru-takeuchi: You are correct. I should use `timeout`. I misunderstood that `kubectl rook-ceph` has `--timeout` option.

.github/workflows/go-test.yaml
```diff
@@ -73,7 +73,15 @@ jobs:
...
-          kubectl rook-ceph rbd ls replicapool
+          SUCCEEDED=0
+          for ((i=0;i<30;i++)) ; do
+            timeout 10 kubectl rook-ceph rbd ls replicapool || continue
+            SUCCEEDED=1
+          done
+          if [ "${SUCCEEDED}" -ne 1 ]; then
+            echo "running rbd command timed out"
+            exit 1
+          fi
```
subhamkrai, line 84:
```suggestion
          SUCCEEDED=0
          for ((i=0;i<30;i++)) ; do
            timeout 10 kubectl rook-ceph rbd ls replicapool || continue
            SUCCEEDED=1
          done
          if [ "${SUCCEEDED}" -ne 1 ]; then
            echo "running rbd command timed out"
            exit 1
          fi
```


I think better option is check move the command before mon restore and check if `replicapool` status is ready other wise command will stuck and this is what happening
```bash
 replicapool   Failure
```

wdyt @satoru-takeuchi ?

satoru-takeuchi: Do you mean that the better way is just exchanging the order of "Mon restore" section and "Rbd command" section?

subhamkrai: yes

satoru-takeuchi:
Probably the root cause is neither mon restore nor rbd command. rbd command hangs iff the number of osd is 0. We should check before this number get 1. I close this PR for now.

PR #264, opened 2024-03-19
clean: increase the timeout to 15min

subhamkrai:
ci: fix namespace typo and add more log

fixing namespace typo in custom-ns test and also adding
more logs to print the resources which are not deleted after
running destroy-cluster command

clean: increase the timeout to 15min

ci can be slow sometimes so let's increase the
`destroy-cluster` timeout to 15min. As seen in
some ci clusters the resources are removed after some time.
Screenshot from 2024-03-19 16-18-10

https://github.com/rook/kubectl-rook-ceph/actions/runs/8340959485/job/22826020306?pr=259
Resolves #

**Checklist:**

### DIFF
--- a/.github/workflows/go-test.yaml
+++ b/.github/workflows/go-test.yaml
@@ -153,7 +153,7 @@ jobs:
           set -ex
           kubectl rook-ceph destroy-cluster
           sleep 1
-          kubectl get deployments -n rook-ceph --no-headers| wc -l | (read n && [ $n -le 1 ] || { echo "the crs could not be deleted"; exit 1;})
+          kubectl get deployments -n rook-ceph --no-headers| wc -l | (read n && [ $n -le 1 ] || { echo "the crs could not be deleted"; kubectl get all -n rook-ceph; exit 1;})
 
       - name: collect common logs
         if: always()
@@ -309,7 +309,7 @@ jobs:
           set -ex
           kubectl rook-ceph --operator-namespace test-operator -n test-cluster destroy-cluster
           sleep 1
-          kubectl get deployments -n rook-ceph --no-headers| wc -l | (read n && [ $n -le 1 ] || { echo "the crs could not be deleted"; exit 1;})
+          kubectl get deployments -n test-cluster --no-headers| wc -l | (read n && [ $n -le 1 ] || { echo "the crs could not be deleted"; kubectl get all -n test-cluster; exit 1;})
 
       - name: collect common logs
         if: always()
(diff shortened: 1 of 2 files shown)


.github/workflows/go-test.yaml
```diff
@@ -153,7 +153,7 @@ jobs:
           set -ex
           kubectl rook-ceph destroy-cluster
           sleep 1
-          kubectl get deployments -n rook-ceph --no-headers| wc -l | (read n && [ $n -le 1 ] || { echo "the crs could not be deleted"; exit 1;})
+          kubectl get deployments -n rook-ceph --no-headers| wc -l | (read n && [ $n -le 1 ] || { echo "the crs could not be deleted"; kubectl get all -n rook-ceph; exit 1;})
```
parth-gr, line 156: why we cant simply use `kubectl wait`

parth-gr (changes requested):

.github/workflows/go-test.yaml
subhamkrai: we don't need to wait as the wait time is in the main go code in last line we are just checking if any ceph cluster cr is not deleted

parth-gr: got it

pkg/crds/crds.go
```diff
@@ -61,7 +61,7 @@ const (
 	CephResourceCephClusters = "cephclusters"
 	toolBoxDeployment        = "rook-ceph-tools"
 	timeOutCheckResources    = 5 * time.Second
-	maxExecutionTime         = 10 * time.Minute
+	maxExecutionTime         = 15 * time.Minute
```
parth-gr, line 64: i didnt checked that code can you point me to that, I believe we still can use kubectl wait there

subhamkrai: so let me ask this, why you are suggesting to use `kubectl wait`? to wait for the resources to be deleted. right? but the above code that you mentioned is the time limit to wait for resources to be deleted after which context will be cancelled.

parth-gr: I saw the code now it makes sense, 
But just for to understand why we need a check with an infinite loop for deleting the resources, I have never seen this approach in rook operator or we do similar like there too?

subhamkrai: you see this line https://github.com/rook/kubectl-rook-ceph/blob/e8e209a623c03b774d0578647552abb1497b4afc/pkg/crds/crds.go#L181-L184

in the context we are setting the timeout after that it will come out so it is not infinite loop techinally

parth-gr (approved):

Madhu-1 (approved):

PR #265, opened 2024-03-20
csi: add setup for external mode for subvolume cleanup

yati1998:
This commit adds extra setup that if required for
subvolume cleanup incase of external mode

### DIFF
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -40,6 +40,11 @@ type subVolumeInfo struct {
 	state string
 }
 
+type monitor struct {
+	ClusterID string
+	Monitors  []string
+}
+
 const (
 	inUse             = "in-use"
 	stale             = "stale"
@@ -50,6 +55,65 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
 
 	subvolumeNames := getK8sRefSubvolume(ctx, clientsets)
 	listCephFSSubvolumes(ctx, clientsets, operatorNamespace, clusterNamespace, includeStaleOnly, subvolumeNames)
+}
+
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	cephclusters, err := clientsets.Rook.CephV1().CephClusters(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to list CephClusters: %q", err))
+	}
+	for i := range cephclusters.Items {
+		enable = cephclusters.Items[i].Spec.External.Enable
+		if enable == true {
+			break
+		}
+	}
+	return enable
+}
+
+// getExternalClusterDetails gets the required mon-ip, id and key to connect to the
+// ceph cluster.
+func getExternalClusterDetails(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, string, string) {
+	var adminId, adminKey, m string
+
+	scList, err := clientsets.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching secrets: %v\n", err))
+	}
+	for i := range scList.Items {
+		if strings.HasPrefix(scList.Items[i].ObjectMeta.Name, "rook-csi-cephfs-provisioner") {
+			data := scList.Items[i].Data
+			if data == nil {
+				logging.Fatal(fmt.Errorf("Secret data is empty"))
+			}
+			adminId = string(data["adminID"])
+			adminKey = string(data["adminKey"])
+			break
+
+		}
+	}
+
+	cm, err := clientsets.Kube.CoreV1().ConfigMaps(clusterNamespace).Get(ctx, "rook-ceph-mon-endpoints", v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching configmaps: %v\n", err))
+	}
+
+	if len(cm.Data) == 0 || cm.Data == nil {
+		logging.Fatal(fmt.Errorf("Configmap data is empty"))
+	}
+	monpoint := cm.Data["csi-cluster-config-json"]
+	var monip []monitor
+	json.Unmarshal([]byte(monpoint), &monip)
+	for _, mp := range monip {
+		if len(mp.Monitors) == 0 || mp.Monitors[0] == "" {
+			logging.Fatal(fmt.Errorf("mon ip is empty"))
+		}
+		m = mp.Monitors[0]
+	}
+
+	return m, adminId, adminKey
 }
 
 // getk8sRefSubvolume returns the k8s ref for the subvolumes
@@ -65,6 +129,17 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 		}
 	}
 	return subvolumeNames
+}
+
+// runCommand checks for the presence of externalcluster and runs the command accordingly.
+func runCommand(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, cmd string, args []string) (string, error) {
+	if checkForExternalStorage(ctx, clientsets, clusterNamespace) {
+		m, admin_id, admin_key := getExternalClusterDetails(ctx, clientsets, operatorNamespace, clusterNamespace)
+		args = append(args, "-m", m, "--id", admin_id, "--key", admin_key)
+	}
+	list, err := exec.RunCommandInOperatorPod(ctx, clientsets, cmd, args, operatorNamespace, clusterNamespace, true)
+
+	return list, err
 }
 
 // listCephFSSubvolumes list all the subvolumes
@@ -87,7 +162,9 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			continue
 		}
 		for _, svg := range subvolg {
-			svList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "ls", fs.Name, svg.Name}, operatorNamespace, clusterNamespace, true)
+			cmd := "ceph"
+			args := []string{"fs", "subvolume", "ls", fs.Name, svg.Name, "--format", "json"}
+			svList, err := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 			if err != nil {
 				logging.Error(err, "failed to get subvolumes of %q", fs.Name)
 				continue
@@ -136,7 +213,10 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 
 // getSubvolumeState returns the state of the subvolume
 func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
-	subVolumeInfo, errvol := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup}, operatorNamespace, clusterNamespace, true)
+	cmd := "ceph"
+	args := []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup, "--format", "json"}
+
+	subVolumeInfo, errvol := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 	if errvol != nil {
 		logging.Error(errvol, "failed to get filesystems")
 		return ""
@@ -155,7 +235,11 @@ func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, oper
 
 // gets list of filesystem
 func getFileSystem(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) ([]fsStruct, error) {
-	fsList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "ls", "--format", "json"}, operatorNamespace, clusterNamespace, true)
+
+	cmd := "ceph"
+	args := []string{"fs", "ls", "--format", "json"}
+
+	fsList, err := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 	if err != nil {
 		logging.Error(err, "failed to get filesystems")
 		return []fsStruct{}, err
@@ -170,7 +254,10 @@ func getFileSystem(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 // checkSnapshot checks if there are any snapshots in the subvolume
 func checkSnapshot(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fs, sv, svg string) bool {
 
-	snapList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "snapshot", "ls", fs, sv, svg}, operatorNamespace, clusterNamespace, true)
+	cmd := "ceph"
+	args := []string{"fs", "subvolume", "snapshot", "ls", fs, sv, svg, "--format", "json"}
+
+	snapList, err := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 	if err != nil {
 		logging.Error(err, "failed to get subvolume snapshots of %q/%q/%q", fs, sv, svg)
 		return false
@@ -185,7 +272,10 @@ func checkSnapshot(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 
 // gets the list of subvolumegroup for the specified filesystem
 func getSubvolumeGroup(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fs string) ([]fsStruct, error) {
-	svgList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolumegroup", "ls", fs, "--format", "json"}, operatorNamespace, clusterNamespace, true)
+	cmd := "ceph"
+	args := []string{"fs", "subvolumegroup", "ls", fs, "--format", "json"}
+
+	svgList, err := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 	if err != nil {
 		logging.Error(err, "failed to get subvolume groups for filesystem %q", fs)
 		return []fsStruct{}, err
@@ -211,7 +301,10 @@ func Delete(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNamespa
 	_, check := k8sSubvolume[subvol]
 	if !check {
 		deleteOmapForSubvolume(ctx, clientsets, OperatorNamespace, CephClusterNamespace, subvol, fs)
-		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}, OperatorNamespace, CephClusterNamespace, true)
+		cmd := "ceph"
+		args := []string{"fs", "subvolume", "rm", fs, subvol, svg, "--retain-snapshots"}
+
+		_, err := runCommand(ctx, clientsets, OperatorNamespace, CephClusterNamespace, cmd, args)
 		if err != nil {
 			logging.Fatal(err, "failed to delete subvolume of %s/%s/%s", fs, svg, subvol)
 		}
@@ -246,8 +339,11 @@ func deleteOmapForSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets,
 		logging.Fatal(fmt.Errorf("pool name not found: %q", err))
 	}
 	if omapval != "" {
+		cmd := "rados"
+		args := []string{"rm", omapval, "-p", poolName, "--namespace", "csi"}
+
 		// remove omap object.
-		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rm", omapval, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		_, err := runCommand(ctx, clientsets, OperatorNamespace, CephClusterNamespace, cmd, args)
 		if err != nil {
 			logging.Fatal(err, "failed to remove omap object for subvolume %q", subVol)
 		}
@@ -255,8 +351,11 @@ func deleteOmapForSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets,
 
 	}
 	if omapkey != "" {
+		cmd := "rados"
+		args := []string{"rmomapkey", "csi.volumes.default", omapkey, "-p", poolName, "--namespace", "csi"}
+
 		// remove omap key.
-		_, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", []string{"rmomapkey", "csi.volumes.default", omapkey, "-p", poolName, "--namespace", "csi"}, OperatorNamespace, CephClusterNamespace, true)
+		_, err := runCommand(ctx, clientsets, OperatorNamespace, CephClusterNamespace, cmd, args)
 		if err != nil {
 			logging.Fatal(err, "failed to remove omap key for subvolume %q", subVol)
 		}
@@ -279,8 +378,9 @@ func getOmapKey(ctx context.Context, clientsets *k8sutil.Clientsets, OperatorNam
 	}
 	omapval := getOmapVal(subVol)
 
-	cmd := []string{"getomapval", omapval, "csi.volname", "-p", poolName, "--namespace", "csi", "/dev/stdout"}
-	pvname, err := exec.RunCommandInOperatorPod(ctx, clientsets, "rados", cmd, OperatorNamespace, CephClusterNamespace, true)
+	args := []string{"getomapval", omapval, "csi.volname", "-p", poolName, "--namespace", "csi", "/dev/stdout"}
+	cmd := "rados"
+	pvname, err := runCommand(ctx, clientsets, OperatorNamespace, CephClusterNamespace, cmd, args)
 	if err != nil || pvname == "" {
 		logging.Info("No PV found for subvolume %s: %s", subVol, err)
 		return ""


pkg/filesystem/subvolume.go
```diff
@@ -87,7 +143,18 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
-			svList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "ls", fs.Name, svg.Name}, operatorNamespace, clusterNamespace, true)
+			cmd := "ceph"
+			args := []string{"fs", "subvolume", "ls", fs.Name, svg.Name, "--format", "json"}
+
+			if checkForExternalStorage() {
+				m, admin_id, admin_key := getExternalClusterDetails(ctx, clientsets, operatorNamespace, clusterNamespace)
+				cmd = "/bin/sh"
+				args = []string{
+					"-c",
+					fmt.Sprintf(`export CEPH_ARGS="-m=%s --id=%s --key=%s" && ceph fs subvolume ls %s %s --conf=/var/lib/rook/%s/%s.config --format json`, m, admin_id, admin_key, fs.Name, svg.Name, clusterNamespace, clusterNamespace),
```
nixpanic, line 154: What is the reason to use `CEPH_ARGS`, and not pass the arguments to the `ceph` command directly? This looks more complex than what is needed, I think.

Is it not the same as this, with enhanced quoting of the `--key=` parameter?
```
ceph -m=%s --id=%s --key=%q fs subvolume ls %s %s --conf=/var/lib/rook/%s/%s.config --format json
```

yati1998: exporting to CEPH_ARGS is taken as a priority, I have tried earlier to pass them directly as you mentioned. But in this case its starts looking for file in which keyring is stored and hence fails to get them.

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +54,78 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
 	listCephFSSubvolumes(ctx, clientsets, operatorNamespace, clusterNamespace, includeStaleOnly, subvolumeNames)
 }
 
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	resource := "storageclusters"
+	storageclusters, err := clientsets.ListResourcesDynamically(ctx, "ocs.openshift.io", "v1", resource, clusterNamespace)
```
Madhu-1, line 61: ocs.openshift,io should not use used in upstream Rook

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +54,78 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+
+	// This is the sample data of secret decoded from external-cluster-details above.
+	//{"name": "rook-ceph-mon-endpoints", "kind": "ConfigMap",
+	//"data": {"data": "ceph-ci-vavuthuextdevyp1-7w8tg-u3w3fp-node1-installer=10.0.209.62:6789", "maxMonId": "0", "mapping": "{}"}},
+	// {"name": "rook-csi-cephfs-provisioner", "kind": "Secret", "data": {"adminID": "csi-cephfs-provisioner",
+	//"adminKey": "AQDSVvhlMnPvABAAHqbM73kgNKYtXi9hwfDaBg=="}}, {"name": "rook-csi-cephfs-node", "kind": "Secret", "data":
+	//{"adminID": "csi-cephfs-node", "adminKey": "AQDSVvhl58U7ABAAtRkNDbAD4f1PXB0fNYlqwQ=="}}
+	// To connect to the ceph-cluster we need the mon-ip(10.0.209.62:6789), the adminID(csi-cephfs-node) and
+	// the key(AQDSVvhl58U7ABAAtRkNDbAD4f1PXB0fNYlqwQ==)
+	split := strings.Split(external_cluster_details, "},")
```
Madhu-1, line 103: cant we use json.unmarhsal to a struct?

pkg/filesystem/subvolume.go
Madhu-1: @yati1998 as @nixpanic mentioned you will have the option to pass `--id= --keyring= --monitor=` explicitly if you want to connect to a different ceph cluster.

pkg/filesystem/subvolume.go
```diff
@@ -185,7 +305,18 @@ func checkSnapshot(ctx context.Context, clientsets *k8sutil.Clientsets, operator
...
+
+	if checkForExternalStorage(ctx, clientsets, clusterNamespace) {
+		m, admin_id, admin_key := getExternalClusterDetails(ctx, clientsets, operatorNamespace, clusterNamespace)
+		cmd = "/bin/sh"
+		args = []string{
+			"-c",
+			fmt.Sprintf(`export CEPH_ARGS="-m=%s --id=%s --key=%s" && ceph fs subvolumegroup ls %s --conf=/var/lib/rook/%s/%s.config --format json`, m, admin_id, admin_key, fs, clusterNamespace, clusterNamespace),
+		}
+	}
+	svgList, err := exec.RunCommandInOperatorPod(ctx, clientsets, cmd, args, operatorNamespace, clusterNamespace, true)
```
Madhu-1, line 319: we should have helper function to run the command , it will become maintenance problem and easy to introduce bugs if we follow if/else approach to run same comment but with different ceph clusters.

Madhu-1 (changes requested):

pkg/filesystem/subvolume.go
yati1998: struct will be bit complex, using map was a bit direct I guess. Cause the format pasted in comment is just the part of what is required. The whole output is bit more . Map will be easy to lookup rather than iterating through an array of struct. what do you say?

Madhu-1: i believe struct will be straight forward but you need to know what  struct to unmarshal to, string split and traversing through it is kind of error prone

yati1998: okay, I will try to have a look at the struct and see if that can be done.

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +54,78 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+			return enable
+		}
+	}
+	return enable
+}
+
+// getExternalClusterDetails exports the CEPH_ARGS with required mon-ip, id and key to connect to the
+// ceph cluster.
+func getExternalClusterDetails(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, string, string) {
+	getsecret := fmt.Sprintf(`kubectl get secret rook-ceph-external-cluster-details -n %s -o=jsonpath='{.data.external_cluster_details}' |  base64 -d`, clusterNamespace)
```
Madhu-1, line 86: i might be missing something, i dont see any reference to `rook-ceph-external-cluster-details` secret in upstream Rook repo. when and who will create this in upstream Rook?

yati1998: This is not in the upstream rook, I was searching for the same, We need to add some check, this is only specific to the external cluster.

subhamkrai: please use api to fetch secrets

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +58,64 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
 	listCephFSSubvolumes(ctx, clientsets, operatorNamespace, clusterNamespace, includeStaleOnly, subvolumeNames)
 }
 
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	resource := "storageclusters"
```
subhamkrai, line 64:
```suggestion
	resource := "storageclusters"
```
in upstream it is `cephClusters` not `stroageClusters`

subhamkrai (changes requested):

Madhu-1 (commented):
@yati1998 we cannot add downstream specific things to the upstream tool which is in Rook repo.

yati1998:
> @yati1998 we cannot add downstream specific things to the upstream tool which is in Rook repo.

Got it, we need to have additional changes in odf-cli for downstream. Will update the PR here.

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +58,71 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+		}
+		if spec["enable"] != nil {
+			enable = spec["enable"].(bool)
+			return enable
+		}
+	}
+	return enable
+}
+
+// getExternalClusterDetails exports the CEPH_ARGS with required mon-ip, id and key to connect to the
```
Madhu-1, line 87: update the comment

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +58,71 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+// ceph cluster.
+func getExternalClusterDetails(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, string, string) {
+	var admin_id, admin_key, m string
+
+	scList, err := clientsets.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching secrets: %v\n", err))
+	}
+	for _, sc := range scList.Items {
+		if sc.ObjectMeta.Name == "rook-csi-cephfs-node" {
```
Madhu-1, line 97: why are we using cephfs node user? what if they are using someother secrets like restricted one? @parth-gr i hope its possible to use different secrets names for different filesystem/subvolumegroups

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +58,71 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+			admin_key = string(data["adminKey"])
+		}
+	}
+
+	cmList, err := clientsets.Kube.CoreV1().ConfigMaps(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching configmaps: %v\n", err))
+	}
+	for _, cm := range cmList.Items {
+		if cm.ObjectMeta.Name == "rook-ceph-mon-endpoints" {
```
Madhu-1, line 109: if we know the configmap name why are we doing list operations instead of get?

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +58,71 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+			// data is of the below format
+			// "ceph-ci-vavuthuextdevyp2-btr2b-hwxgpn-node1-installer=10.0.209.241:6789,
+			// ceph-ci-vavuthuextdevyp2-btr2b-hwxgpn-node2=10.0.210.220:6789,
+			// ceph-ci-vavuthuextdevyp2-btr2b-hwxgpn-node5=10.0.209.2:6789,
+			// ceph-ci-vavuthuextdevyp2-btr2b-hwxgpn-node3=10.0.211.109:6789"
```
Madhu-1, line 114: can we use `csi-cluster-config-json` from the `"rook-ceph-mon-endpoints"` or use `rook-ceph-csi-config` configmap as its in known format and we don't need to split anything, just need to unmarshal to a struct

pkg/filesystem/subvolume.go
```diff
@@ -87,7 +158,14 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			continue
 		}
 		for _, svg := range subvolg {
-			svList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "ls", fs.Name, svg.Name}, operatorNamespace, clusterNamespace, true)
+			cmd := "ceph"
+			args := []string{"fs", "subvolume", "ls", fs.Name, svg.Name, "--format", "json"}
+
+			if checkForExternalStorage(ctx, clientsets, clusterNamespace) {
+				m, admin_id, admin_key := getExternalClusterDetails(ctx, clientsets, operatorNamespace, clusterNamespace)
+				args = []string{"fs", "subvolume", "ls", fs.Name, svg.Name, "--format", "json", "-m", m, "--id", admin_id, "--key", admin_key}
```
Madhu-1, line 166: Do we need to run this CLI from the machine where the ceph packages are installed?

pkg/filesystem/subvolume.go
```diff
@@ -155,7 +240,16 @@ func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, oper
...
 func getFileSystem(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) ([]fsStruct, error) {
-	fsList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "ls", "--format", "json"}, operatorNamespace, clusterNamespace, true)
+
+	cmd := "ceph"
+	args := []string{"fs", "ls", "--format", "json"}
+
+	if checkForExternalStorage(ctx, clientsets, clusterNamespace) {
+		m, admin_id, admin_key := getExternalClusterDetails(ctx, clientsets, operatorNamespace, clusterNamespace)
+		args = []string{"fs", "ls", "--format", "json", "-m", m, "--id", admin_id, "--key", admin_key}
+	}
```
Madhu-1, line 250: why cant we run this from the operator pod? i hope rook also runs ceph commands form the operator pod. @subhamkrai can you please check this.

pkg/filesystem/subvolume.go
parth-gr: Yes that's correct, it will be variable,
the prefix will be always `rook-csi-cephfs-node`, 
Not sure about the current operation but we can see the name from the storage class parameter.

pkg/filesystem/subvolume.go
yati1998: Not sure what exactly is your query, we are running from operator pod only, this is just to get the arguements.

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +60,48 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
 	listCephFSSubvolumes(ctx, clientsets, operatorNamespace, clusterNamespace, includeStaleOnly, subvolumeNames)
 }
 
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	cephclusters, _ := clientsets.Rook.CephV1().CephClusters(clusterNamespace).List(ctx, v1.ListOptions{})
+	for _, cephcluster := range cephclusters.Items {
+		enable = cephcluster.Spec.External.Enable
```
Madhu-1, line 68: just return here of use break so that if we find an external cluster we don't need to go in the loop again.

pkg/filesystem/subvolume.go
Madhu-1: use provisioner secret and also check as per above comment we might not always create secret with exact names, you might need to use `strings.HasPrefix` or something similar

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +60,48 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+	}
+	for _, sc := range scList.Items {
+		if sc.ObjectMeta.Name == "rook-csi-cephfs-node" {
+			data := sc.Data
+			admin_id = string(data["adminID"])
+			admin_key = string(data["adminKey"])
+		}
+	}
+
+	cm, err := clientsets.Kube.CoreV1().ConfigMaps(clusterNamespace).Get(ctx, "rook-ceph-mon-endpoints", v1.GetOptions{})
```
Madhu-1, line 90: This will be in operator namespace or cluster Namespace?

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +60,48 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+func getExternalClusterDetails(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, string, string) {
+	var admin_id, admin_key, m string
+
+	scList, err := clientsets.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching secrets: %v\n", err))
+	}
+	for _, sc := range scList.Items {
+		if sc.ObjectMeta.Name == "rook-csi-cephfs-node" {
+			data := sc.Data
```
Madhu-1, line 84: what it `sc.Data` is nil? please add a check for it as well

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +60,48 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+			admin_key = string(data["adminKey"])
+		}
+	}
+
+	cm, err := clientsets.Kube.CoreV1().ConfigMaps(clusterNamespace).Get(ctx, "rook-ceph-mon-endpoints", v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching configmaps: %v\n", err))
+	}
+
+	data := cm.Data
```
Madhu-1, line 95: same here what if `cm.Data` is nil? please add a check for the same

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +60,48 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+	monpoint := data["csi-cluster-config-json"]
+	var monip []monitor
+	json.Unmarshal([]byte(monpoint), &monip)
+	for _, mp := range monip {
+		m = mp.Monitors[0]
```
Madhu-1, line 100: check the length of `mp` before accessing it. use break so that you don't need to go through the loop again?

pkg/filesystem/subvolume.go
```diff
@@ -67,6 +117,17 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	return subvolumeNames
 }
 
+// RunCommand checks for the presence of externalcluster and runs the command accordingly.
+func runCommand(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, cmd string, args []string) (string, error) {
+	if checkForExternalStorage(ctx, clientsets, clusterNamespace) {
+		m, admin_id, admin_key := getExternalClusterDetails(ctx, clientsets, operatorNamespace, clusterNamespace)
+		args = append(args, "-m", m, "--id", admin_id, "--key", admin_key)
+	}
+	List, err := exec.RunCommandInOperatorPod(ctx, clientsets, cmd, args, operatorNamespace, clusterNamespace, true)
```
Madhu-1, line 126: use small `l` in `List`

pkg/filesystem/subvolume.go
```diff
@@ -67,6 +117,17 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	return subvolumeNames
 }
 
+// RunCommand checks for the presence of externalcluster and runs the command accordingly.
```
Madhu-1, line 120: `RunCommand` to `runCommand`

pkg/filesystem/subvolume.go
```diff
@@ -170,7 +240,10 @@ func getFileSystem(ctx context.Context, clientsets *k8sutil.Clientsets, operator
 // checkSnapshot checks if there are any snapshots in the subvolume
 func checkSnapshot(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fs, sv, svg string) bool {
 
-	snapList, err := exec.RunCommandInOperatorPod(ctx, clientsets, "ceph", []string{"fs", "subvolume", "snapshot", "ls", fs, sv, svg}, operatorNamespace, clusterNamespace, true)
+	cmd := "ceph"
+	args := []string{"fs", "subvolume", "snapshot", "ls", fs, sv, svg}
```
Madhu-1, line 244: why not use `format=json` here?

pkg/filesystem/subvolume.go
```diff
@@ -39,6 +39,14 @@ type subVolumeInfo struct {
 	fs    string
 	state string
 }
+type externalClusterDetails struct {
+	Name string
+	Data map[string]string
+}
```
Madhu-1, line 45: Not used anymore and can be removed?

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +60,48 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
 	listCephFSSubvolumes(ctx, clientsets, operatorNamespace, clusterNamespace, includeStaleOnly, subvolumeNames)
 }
 
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	cephclusters, _ := clientsets.Rook.CephV1().CephClusters(clusterNamespace).List(ctx, v1.ListOptions{})
```
Madhu-1, line 66: why err is discarded here?

pkg/filesystem/subvolume.go
yati1998: It should be cluster Namespace only, @subhamkrai correct me if I am wrong!!

pkg/filesystem/subvolume.go
yati1998: ahh, missed it while rewriting, thanks.

pkg/filesystem/subvolume.go
Madhu-1: we should return err if Data is nil

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,59 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+	scList, err := clientsets.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching secrets: %v\n", err))
+	}
+	for _, sc := range scList.Items {
+		if strings.HasPrefix(sc.ObjectMeta.Name, "rook-csi-cephfs-provisioner") {
+			data := sc.Data
+			if data != nil {
+				admin_id = string(data["adminID"])
+				admin_key = string(data["adminKey"])
```
Madhu-1, line 90: lets not use variable name with `_` its not as part of standard practice

pkg/filesystem/subvolume.go
Madhu-1: same here

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,59 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+
+	data := cm.Data
+	if data != nil {
+		monpoint := data["csi-cluster-config-json"]
+		var monip []monitor
+		json.Unmarshal([]byte(monpoint), &monip)
+		for _, mp := range monip {
+			m = mp.Monitors[0]
+		}
+	}
```
Madhu-1, line 109: add a check to ensure `m` is not emtpty

Madhu-1 (commented):
small nit, LGTM

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,66 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
 }
 
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	cephclusters, err := clientsets.Rook.CephV1().CephClusters(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to list CephClusters: %q", err))
+	}
+	for _, cephcluster := range cephclusters.Items {
```
Madhu-1, line 67:
```suggestion
for i := range cephclusters.Items {
		enable = cephclusters.Items[i].Spec.External.Enable
		if enable == true {
			break
		}
	}
	return enable
```

avoid in-memory copy when doing for loop

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,66 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+func getExternalClusterDetails(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, string, string) {
+	var adminId, adminKey, m string
+
+	scList, err := clientsets.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching secrets: %v\n", err))
+	}
+	for _, sc := range scList.Items {
+		if strings.HasPrefix(sc.ObjectMeta.Name, "rook-csi-cephfs-provisioner") {
+			data := sc.Data
```
Madhu-1, line 87:
```suggestion
for i := range scList.Items {
		if strings.HasPrefix(scList.Items[i].ObjectMeta.Name, "rook-csi-cephfs-provisioner") {
			data := scList.Items[i].Data

**Review by Madhu-1 — COMMENTED** [review]
LGTM

**Review by Madhu-1 — APPROVED** [review]

File: pkg/filesystem/subvolume.go (hunk 14 above)
Comment (subhamkrai) [reply pkg/filesystem/subvolume.go:98]: This will be clusterNamespace only, only rook-operator related resources require operatorNamespace and this cm is related to cephCLuster.

File: pkg/filesystem/subvolume.go [hunk 27]
```diff
@@ -52,6 +57,66 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+
+	cm, err := clientsets.Kube.CoreV1().ConfigMaps(clusterNamespace).Get(ctx, "rook-ceph-mon-endpoints", v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching configmaps: %v\n", err))
+	}
+
+	data := cm.Data
+	if data == nil {
+		logging.Fatal(fmt.Errorf("Configmap data is empty"))
+	}
```
subhamkrai, line 106:
```suggestion
	data := cm.Data
	if data == nil {
		logging.Fatal(fmt.Errorf("Configmap data is empty"))
	}
```

First check len of `cm.Data` then proceed otherwise we'll get nil pointer error when it's empty

pkg/filesystem/subvolume.go
subhamkrai: @yati1998 please check len of `mp.Monitors` before proceeding

subhamkrai (changes requested):

subhamkrai (approved):

yati1998:
@nixpanic please do add your reviews as well, if any.

subhamkrai:
> @nixpanic please do add your reviews as well, if any.

going ahead with the PR, we can follow up later if have more comments. Thanks

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,65 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
 	listCephFSSubvolumes(ctx, clientsets, operatorNamespace, clusterNamespace, includeStaleOnly, subvolumeNames)
 }
 
+// checkForExternalStorage checks if the external mode is enabled.
+func checkForExternalStorage(ctx context.Context, clientsets *k8sutil.Clientsets, clusterNamespace string) bool {
+	enable := false
+	cephclusters, err := clientsets.Rook.CephV1().CephClusters(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("failed to list CephClusters: %q", err))
```
nixpanic, line 65: Maybe include the namespace for people that read the error and try to identify what's wrong?

`"failed to list CephClusters in namespace %q: %v", clusterNamespace, err)`

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,65 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+
+	scList, err := clientsets.Kube.CoreV1().Secrets(clusterNamespace).List(ctx, v1.ListOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching secrets: %v\n", err))
+	}
+	for i := range scList.Items {
+		if strings.HasPrefix(scList.Items[i].ObjectMeta.Name, "rook-csi-cephfs-provisioner") {
+			data := scList.Items[i].Data
+			if data == nil {
+				logging.Fatal(fmt.Errorf("Secret data is empty"))
```
nixpanic, line 89: include the namespace/secretname in the error

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,65 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+			adminId = string(data["adminID"])
+			adminKey = string(data["adminKey"])
+			break
+
+		}
+	}
+
+	cm, err := clientsets.Kube.CoreV1().ConfigMaps(clusterNamespace).Get(ctx, "rook-ceph-mon-endpoints", v1.GetOptions{})
+	if err != nil {
+		logging.Fatal(fmt.Errorf("Error fetching configmaps: %v\n", err))
```
nixpanic, line 100: include the namespace  here too, remove the `\n`

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,65 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+	}
+
+	if len(cm.Data) == 0 || cm.Data == nil {
+		logging.Fatal(fmt.Errorf("Configmap data is empty"))
```
nixpanic, line 104: include the namespace/configmap name

pkg/filesystem/subvolume.go
```diff
@@ -52,6 +57,65 @@ func List(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace
...
+	}
+	monpoint := cm.Data["csi-cluster-config-json"]
+	var monip []monitor
+	json.Unmarshal([]byte(monpoint), &monip)
+	for _, mp := range monip {
+		if len(mp.Monitors) == 0 || mp.Monitors[0] == "" {
+			logging.Fatal(fmt.Errorf("mon ip is empty"))
```
nixpanic, line 111: include the namespace/configmap name, maybe even clusterid too

nixpanic (commented):
Looks good to me. Only a few nits that would improve the user experience when an error is reported.

PR #279, opened 2024-04-15
core: add ceph daemon command support

### LINKED ISSUE #208: is possible to support "ceph daemon cmd"?
such as: "ceph daemon osd.x dump_historic_ops"?

subhamkrai:
adding `ceph daemon *` command support

Resolves #208

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,10 +50,11 @@ These are args currently supported:
     kubectl rook-ceph --context=$(kubectl config current-context) mons
     ```
 
-
 ### Commands
 
 - `ceph <args>` : Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](https://docs.ceph.com/en/pacific/start/intro/) for more.
+
+- `ceph daemon <daemon.id> <args>`: Run a Ceph daemon command by connecting to its admin socket.
 
 - `rados <args>` : Run a Rados CLI command. Supports any arguments the `rados` command supports. See [Rados docs](https://docs.ceph.com/en/latest/man/8/rados/) for more.
 
@@ -92,6 +93,7 @@ These are args currently supported:
 Visit docs below for complete details about each command and their flags uses.
 
 1. [Running ceph commands](docs/ceph.md)
+1. [Running ceph daemon commands](docs/ceph.md#Daemon)
 1. [Running rbd commands](docs/rbd.md)
 1. [Get mon endpoints](docs/mons.md#print-mon-endpoints)
 1. [Get cluster health status](docs/health.md)
@@ -107,7 +109,7 @@ Visit docs below for complete details about each command and their flags uses.
 1. [Disaster Recovery](docs/dr-health.md)
 1. [Restore deleted CRs](docs/crd.md)
 1. [Destroy cluster](docs/destroy-cluster.md)
-2. [Running rados commands](docs/rados.md)
+1. [Running rados commands](docs/rados.md)
 
 ## Examples
 
--- a/cmd/commands/ceph.go
+++ b/cmd/commands/ceph.go
@@ -17,6 +17,9 @@ limitations under the License.
 package command
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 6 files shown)


cmd/commands/ceph.go
```diff
@@ -33,9 +36,20 @@ var CephCmd = &cobra.Command{
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		logging.Info("running 'ceph' command with args: %v", args)
-		_, err := exec.RunCommandInOperatorPod(cmd.Context(), clientSets, cmd.Use, args, operatorNamespace, cephClusterNamespace, false)
-		if err != nil {
-			logging.Fatal(err)
+		if args[0] == "daemon" {
+			if len(args) > 2 {
+				cephLabel := strings.Replace(args[1], "-", ".", 1)
```
travisn, line 41: Could you factor out lines 41-42 to a helper method and then add unit tests for it? Anytime there is string parsing, I recommend unit tests.

travisn: Also can you add a comment that shows the expected input? It's helpful to see why we need to replace a dash.

travisn (changes requested):
How about an update to the readme with this new command?

subhamkrai:
> How about an update to the readme with this new command?

there is already `ceph *` mentioned in Readme, do we want to add specific for `ceph daemon`?

travisn:
> > How about an update to the readme with this new command?
> 
> there is already `ceph *` mentioned in Readme, do we want to add specific for `ceph daemon`?

Seems good to mention as a separate item because it fundamentally does something different. Instead of executing the ceph command in the operator pod, it would connect to the daemon pod.

README.md
```diff
@@ -50,11 +50,12 @@ These are args currently supported:
     kubectl rook-ceph --context=$(kubectl config current-context) mons
     ```
 
-
 ### Commands
 
 - `ceph <args>` : Run a Ceph CLI command. Supports any arguments the `ceph` command supports. See [Ceph docs](https://docs.ceph.com/en/pacific/start/intro/) for more.
 
+- `ceph daemon <cephDaemon.id> <args>`: Run a Ceph daemon command connecting to the admin-socket.
```
travisn, line 57:
```suggestion
- `ceph daemon <daemon.id> <args>`: Run a Ceph daemon command by connecting to its admin socket.
```

cmd/commands/ceph.go
```diff
@@ -33,9 +36,25 @@ var CephCmd = &cobra.Command{
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		logging.Info("running 'ceph' command with args: %v", args)
-		_, err := exec.RunCommandInOperatorPod(cmd.Context(), clientSets, cmd.Use, args, operatorNamespace, cephClusterNamespace, false)
-		if err != nil {
-			logging.Fatal(err)
+		if args[0] == "daemon" {
+			if len(args) > 2 {
```
travisn, line 40: If there are not two args do they get an error?

cmd/commands/ceph.go
```diff
@@ -33,9 +36,25 @@ var CephCmd = &cobra.Command{
...
+				cephDaemonNameAndID := splitCephDaemonNameAndID(args[1])
```
travisn, line 41: Is the error handled if they don't include a `.`?

cmd/commands/ceph.go
```diff
@@ -33,9 +36,25 @@ var CephCmd = &cobra.Command{
...
+			if err != nil {
+				logging.Fatal(err)
+			}
 		}
 	},
 }
+
+// splitCephDaemonNameAndID splits the arg based on '.' so that we use it with label, example osd.0 to osd and 0
+func splitCephDaemonNameAndID(args string) []string {
+	cephDaemonNameAndID := strings.Split(args, ".")
```
travisn, line 58:
```suggestion
	return strings.Split(args, ".")
```

docs/ceph.md
```diff
@@ -96,3 +96,18 @@ kubectl rook-ceph ceph status --format json-pretty
 #     "progress_events": {}
 # }
 ```
+
+## Daemon
+
+Run Ceph daemon command connecting to the admin-socket.
```
travisn, line 102:
```suggestion
Run the Ceph daemon command by connecting to its admin socket.
```

travisn (changes requested):
A few small suggestions

cmd/commands/ceph.go
subhamkrai: now it has
```bash
./bin/kubectl-rook-ceph ceph daemon
Warning: rook version 'rook: v1.14.0-alpha.0.80.gf84e932ec' is running a pre-release version of Rook.

Info: running 'ceph' command with args: [daemon]
Error: too few arguments to run the 'ceph daemon' command
```

cmd/commands/ceph.go
subhamkrai: now it has
```bash
./bin/kubectl-rook-ceph ceph daemon osd 0
Warning: rook version 'rook: v1.14.0-alpha.0.80.gf84e932ec' is running a pre-release version of Rook.

Info: running 'ceph' command with args: [daemon osd 0]
Error: Invalid argument to 'ceph daemon' command: osd. The arg should be in the format of '<daemon.id>'
```

travisn (approved):

PR #285, opened 2024-04-23
build: update rook and k8s api version

subhamkrai:
this commit updates rook and k8s api version

Resolves #

**Checklist:**


go.mod
```diff
@@ -1,19 +1,21 @@
 module github.com/rook/kubectl-rook-ceph
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.1
```
travisn, line 5: Why was the toochain needed as a separate version from go above? It's nice if we can just declare `1.22` and it automatically picks up the latest patch.

subhamkrai: Let's discuss this in huddle,
 the k8s  0.30 api are bringing those changes either we have to mention `toolchain go.*` or we have to use explicit version of go in go.mod like `go 1.22.1`

go.mod
```diff
@@ -1,19 +1,21 @@
 module github.com/rook/kubectl-rook-ceph
 
-go 1.21
+go 1.22.0
```
Madhu-1, line 3: lets use 1.22 not stick to x.y.z so that we always use latest released version and also we will get build problem in downstream if we use `x.y.z` here

subhamkrai: https://go.dev/doc/toolchain

Madhu-1: toolchain is automatically picked up based on the version right? i think that should be fine.

subhamkrai: yes it is auto picked by running go mod tidy

travisn: Go [toolchain doc](https://go.dev/doc/toolchain) says:
"That is, the `go` line sets the minimum required Go version necessary to use a module or workspace."

Sounds find to have 1.22.0 as the min version. But do we need to be opinionated about the toolchain that is chosen? Can we remove the `toolchain` line?

subhamkrai: we can remove the `toolchain` we just need to use the same go toolchain version `go *`. See the changes now

Madhu-1: we will get to problems like https://github.com/red-hat-storage/odf-multicluster-orchestrator/pull/197, sadly that PR doesn't contain details but that was done to fix the build issue in Downstream, for some reason we don't get x.y.z specific go version in downstream , and also this doesn't allow developers to use lower version of 1.22 to build the repo? just ensure that we are covering all the cases

subhamkrai: I think we can discuss this in today's meeting, but anyways we need to use specific version in go.mod either with/without toolchain
```
go 1.22.0

toolchain go1.22.1
```

is added when ran `go mod tidy`

travisn: This does allow multiple versions of Go to be used locally at least during development. If I build this branch then run go version I see:
```
% go version
go version go1.22.1 darwin/arm64
```
Which is interesting because I don't recall installing 1.22, it must have pulled it automatically. Then if I go back to the rook repo, I still see go 1.21:
```
% go version
go version go1.21.5 darwin/arm64
```
So the main requirement I assume is that the downstream build would need to use 1.22 for this repo

subhamkrai:
CLosing this one as this has been fixed in another PR.

PR #290, opened 2024-05-09
chore: improve goreleaser step

chenrui333:
**Checklist:**

---

match with what is in go-test

https://github.com/rook/kubectl-rook-ceph/blob/4521811a74af2ccb286cc7070ef19e07f9f6398a/.github/workflows/go-test.yaml#L42

- also relates to https://github.com/Homebrew/homebrew-core/pull/171289


.goreleaser.yaml
```diff
@@ -3,8 +3,9 @@ before:
     - go mod tidy
 builds:
   - id: kubectl-rook-ceph
+    ldflags: -s -w
     main: ./cmd
-    binary: kubectl-rook-ceph
+    binary: kubectl-rook_ceph
```
subhamkrai, line 8:
```suggestion
    binary: kubectl-rook_ceph
```

can you explain why this change is required and how it is helping if you use the krew plugin to install the tool above change is not required, IIUC you are consuming the release build image and directly calling the binary?

subhamkrai (changes requested):
Also, please describe your changes in PR description

.goreleaser.yaml
chenrui333: I dont know which is more desired, that is why I propose this change to check (as I see the discrepancy with the homebrew artifact, https://github.com/Homebrew/homebrew-core/pull/171289)

sounds like `kubectl-rook-ceph` is preferred over `kubectl-rook_ceph`?

chenrui333: also cc @travisn for thoughts

Skaronator: `kubectl-rook_ceph` is the correct Name of the binary in order to call the plugin with `kubectl rook-ceph`.

Offical kubectl Docs:
https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#names-with-dashes-and-underscores

travisn: That doc indicates that it is "possible" to use underscores, but it is not required. Seems like it is more natural to use dashes, as described earlier in the doc [here](https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#naming-a-plugin).

Skaronator: It seems to me that dashes in the filename represents a space in the kubectl command.

> the command `kubectl foo bar baz` is invoked by the user, would have the filename of `kubectl-foo-bar-baz`.

travisn: The commands work fine in our testing with the dash as in the [readme](https://github.com/rook/kubectl-rook-ceph?tab=readme-ov-file#usage). Are you seeing otherwise?

Skaronator: I'm currently on vacation, so I'm unable to test anything. I'm just referring to the Kubernetes documentation, and although dashes currently works, it's not guaranteed to work in the future. I recommend using an underscore.

Just my 2 cents.

subhamkrai: So, when using binary directly(not with krew) `kubectl-rook-ceph`  it doesn't read it to `kubectl rook-ceph` but using the binary name `kubectl- rook_ceph` it reads it to `kubectl rook-ceph`

nirs: The way to get the right name in help text for a command used as a plugin is documented here:
https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#creating-a-plugin

PR #305, opened 2024-06-26
core: Add status.phase=Running filter to WaitForPodToRun

### LINKED ISSUE #304: `waitForPodToRun` does not account for multiple pods
When I ran `kubectl rook-ceph ceph status` on my cluster, the command never succeeded as it did not pick up on the currently running `ceph-operator` pod:

```
$ kubectl rook-ceph --operator-namespace=rook-ceph ceph status -h
Info: waiting for pod with label "app=rook-ceph-operator" in namespace "rook-ceph" to be running
Info: waiting for pod with label "app=rook-ceph-operator" in namespace "rook-ceph" to be running
[...]
```

But the `rook-ceph` operator was running fine:
```
$ kubectl get pods -n rook-ceph
[...]
rook-ceph-operator-695f48c6bd-r85m8                      0/1     Completed   0               7d19h
rook-ceph-operator-d7fcdf58b-6wnm5                       1/1     Running     0               20h
[...]
```

I noticed however that the previous operator pod `rook-ceph-operator-695f48c6bd-r85m8` was still around. This pod belonged to a previous `ReplicaSet` (I recently updated the cluster). Once deleted, the check was now satisfied and the plugin recognized the operator as being alive.

After checking the code, the root cause seems to be that `waitForPodToRun` does request the list of pods but it only ever checks the first pod returned in that list. The `Completed` pod apparently always came first back from the API, so the check never succeeded. Reworking this into iterating over all the returned pods should fix this. Another approach could be to filter for `Running` pods.

Luminger:
To prevent pods in the `Completed` phase to show up in the list pulled by `WaitForPodToRun` (which resulted in an indefinitely hang), add a filter to only return pods in the `Running` phase.

**Issue resolved by this Pull Request:**
Resolves #304 

**Checklist:**

### DIFF
--- a/pkg/k8sutil/k8sutil.go
+++ b/pkg/k8sutil/k8sutil.go
@@ -44,7 +44,7 @@ func RestartDeployment(ctx context.Context, k8sclientset kubernetes.Interface, n
 }
 
 func WaitForPodToRun(ctx context.Context, k8sclientset kubernetes.Interface, namespace, labelSelector string) (corev1.Pod, error) {
-	opts := v1.ListOptions{LabelSelector: labelSelector}
+	opts := v1.ListOptions{LabelSelector: labelSelector, FieldSelector: "status.phase=Running"}
 	for i := 0; i < 60; i++ {
 		pod, err := k8sclientset.CoreV1().Pods(namespace).List(ctx, opts)
 		if err != nil {


Luminger:
The test failure seems to be a transient error during deletion of the test cluster, can we redo the test run?

subhamkrai:
> The test failure seems to be a transient error during deletion of the test cluster, can we redo the test run?

restarted, now

subhamkrai (approved):

subhamkrai:
Thanks for the PR @Luminger

PR #311, opened 2024-07-11
csi: add check for pending clone

yati1998:
Incase any pvc is in pending state and the backend subvolume is create, it will be counted as stale.
this commit adds the error check to make sure user doesn't delete the subvolume if it the pvc is in
pending state.

### DIFF
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -19,8 +19,10 @@ package subvolume
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
+	"syscall"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
@@ -175,8 +177,19 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil {
+					if errors.Is(err, syscall.EAGAIN) {
+						logging.Warning("Found pending clone: %q", sv.Name)
+						logging.Warning("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
+						logging.Warning("To avoid stale resources, please scale down the cephfs deployment before deleting the subvolume.")
+						continue
+					}
+					logging.Fatal(fmt.Errorf("failed to get subvolume state: %q %q", sv.Name, err))
 
+				}
 				// Assume the volume is stale unless proven otherwise
 				stalevol := true
 				// lookup for subvolume in list of the PV references
@@ -195,7 +208,6 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 					// check the state of the stale subvolume
 					// if it is snapshot-retained then skip listing it.
 					if state == "snapshot-retained" {
-						status = state
 						continue
 					}
 					// check if the stale subvolume has snapshots.
@@ -212,14 +224,14 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 }
 
 // getSubvolumeState returns the state of the subvolume
-func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) string {
+func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace, fsName, SubVol, SubvolumeGroup string) (string, error) {
 	cmd := "ceph"
 	args := []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup, "--format", "json"}
 
 	subVolumeInfo, errvol := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 	if errvol != nil {
-		logging.Error(errvol, "failed to get filesystems")
-		return ""
+		logging.Error(errvol, "failed to get subvolume info")
+		return "", errvol
 	}
 	var info map[string]interface{}
 	err := json.Unmarshal([]byte(subVolumeInfo), &info)
@@ -230,7 +242,7 @@ func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, oper
 	if !ok {
 		logging.Fatal(fmt.Errorf("failed to get the state of subvolume: %q", SubVol))
 	}
-	return state
+	return state, nil
 }
 
 // gets list of filesystem


pkg/filesystem/subvolume.go
```diff
@@ -175,8 +175,13 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
-
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil {
+					logging.Info("Subvolume info not found for: %q", err)
+					logging.Info("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
```
Madhu-1, line 183: is it missing `continue` `return` or break ?

pkg/filesystem/subvolume.go
```diff
@@ -175,8 +175,13 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
-
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil {
```
Madhu-1, line 181: Considering the `err!=nil` as pending is not the right assumption, it can fail for other reasons as well. please add a valid check to look for specific strings in error

Madhu-1 (changes requested):

pkg/filesystem/subvolume.go
yati1998: yes, as per now it considers the subvolume as stale along with that it prints the above text, but yeah we can skip this subvolume.

pkg/filesystem/subvolume.go
```diff
@@ -175,8 +177,14 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
-
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil && errors.Is(err, syscall.EAGAIN) {
+					logging.Info("Subvolume info not found for: %q", err)
+					logging.Info("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
+					continue
+				}
```
Madhu-1, line 187: what if the error is not nil and not EAGAIN?

can you please test this and see if we are getting `EAGAIN` here to make sure we are good 👍🏻

pkg/filesystem/subvolume.go
```diff
@@ -195,12 +203,14 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
 					// if it is snapshot-retained then skip listing it.
 					if state == "snapshot-retained" {
-						status = state
 						continue
 					}
-					// check if the stale subvolume has snapshots.
-					if checkSnapshot(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name) {
-						status = staleWithSnapshot
+					// check for snapshot only if the subvolume info doesn't return any error.
+					if err == nil {
```
Madhu-1, line 209: why we need this check here? i assume `err` should be handled above not here

pkg/filesystem/subvolume.go
yati1998:
```
yatipadia:kubectl-rook-ceph$ ./bin/kubectl-rook-ceph subvolume ls
Filesystem  Subvolume  SubvolumeGroup  State
myfs csi-vol-9e760985-c5cd-46d1-8484-8ecfa1e3d6bc csi in-use
myfs csi-vol-5d13c38b-2263-46bc-826b-6c0df2ad4bcb csi in-use
Error: Error EAGAIN: subvolume 'csi-vol-71bb4f78-1373-4b7e-8400-ea5fd96a22dc' is not ready for operation info
. failed to run command. command terminated with exit code 11%!(EXTRA string=failed to get subvolume info)
Info: Subvolume info not found for: "Error EAGAIN: subvolume 'csi-vol-71bb4f78-1373-4b7e-8400-ea5fd96a22dc' is not ready for operation info\n. failed to run command. command terminated with exit code 11"
Info: Please delete the pending pv if any before deleting the 
```

Here's the output if the clone is in pending state , the info command is not ready incase the subvolume is not properly created, so this should be fine.

pkg/filesystem/subvolume.go
yati1998: in case there is error running subvolume info command, we shouldn't check for the snapshot also,  that would just be an addition to  run time.

yati1998: okay, after adding continue to above case, we no more need this check as in case of err it will not go further. Thanks for notifying.

pkg/filesystem/subvolume.go
Madhu-1: @yati1998 in case of err !=nil and error is not syscall.EAGAIN we need to log it as error and let the user run the tool again, we should not ignore errors unless we know why we are doing like that

yati1998: done

yati1998: please review it again, thank you @Madhu-1

pkg/filesystem/subvolume.go
```diff
@@ -175,8 +177,18 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil {
+					if errors.Is(err, syscall.EAGAIN) {
+						logging.Info("Subvolume info not found for: %q %q", sv.Name, err)
+						logging.Info("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
```
Madhu-1, line 186: they also need to scale down the cephfs csi deployment to avoid stale resources isn't it?

yati1998: added the log @Madhu-1

pkg/filesystem/subvolume.go
```diff
@@ -175,8 +177,19 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 			// append the subvolume which doesn't have any snapshot attached to it.
 			for _, sv := range subvol {
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil {
+					if errors.Is(err, syscall.EAGAIN) {
+						logging.Info("Subvolume info not found for: %q %q", sv.Name, err)
```
Madhu-1, line 185: `logging.Info("Found Pending clone: %q %q", sv.Name, err)`
log this as warning as well.

Madhu-1 (approved):
LGTM

Madhu-1 (approved):

pkg/filesystem/subvolume.go
```diff
@@ -175,8 +177,19 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
...
 			for _, sv := range subvol {
-				state := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				state, err := getSubvolumeState(ctx, clientsets, operatorNamespace, clusterNamespace, fs.Name, sv.Name, svg.Name)
+				// subvolume info returns error in case of pending clone or if it is not ready
+				// it is suggested to delete the pvc before deleting the subvolume.
+				if err != nil {
+					if errors.Is(err, syscall.EAGAIN) {
+						logging.Warning("Found pending clone: %q", sv.Name)
+						logging.Warning("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
+						logging.Warning("To avoid stale resources, please scale down the cephfs deployment and then run the cleanup script.")
```
travisn, line 187: What cleanup script?

yati1998: the subvolume delete command, maybe this could confuse the user, let me update.

PR #312, opened 2024-07-12
csi: look for subvolumepath instead of subvolumename

yati1998:
In case of ROX pvc, the subvolume name is different and is not listed in the backend as it is just a shallow volume and it links to the source path.
incase the source volume is delete the subvolme will be shown stale, but actually that is being used by the ROX pvc.
Soinstead getting the subvolumename to look for stale subvolumes, it is better to retrive the name from the subvolumepath.

### DIFF
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -125,10 +125,29 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	subvolumeNames := make(map[string]subVolumeInfo)
 	for _, pv := range pvList.Items {
 		if pv.Spec.CSI != nil {
-			subvolumeNames[pv.Spec.CSI.VolumeAttributes["subvolumeName"]] = subVolumeInfo{}
+			subvolumePath := pv.Spec.CSI.VolumeAttributes["subvolumePath"]
+			name, err := getSubvolumeNameFromPath(subvolumePath)
+			if err != nil {
+				logging.Error(err, "failed to get subvolume name")
+				continue
+			}
+			subvolumeNames[name] = subVolumeInfo{}
 		}
 	}
 	return subvolumeNames
+}
+
+// getSubvolumeNameFromPath get the subvolumename from the path.
+// subvolumepath: /volumes/csi/csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d/5f4e4caa-f835-41ba-83c1-5bbd57f6aedf
+// subvolumename: csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d
+func getSubvolumeNameFromPath(path string) (string, error) {
+	splitSubvol := strings.Split(path, "/")
+	if len(splitSubvol) < 4 {
+		return "", fmt.Errorf("failed to get name from subvolumepath: %s", path)
+	}
+	name := splitSubvol[3]
+
+	return name, nil
 }
 
 // runCommand checks for the presence of externalcluster and runs the command accordingly.
--- a/pkg/filesystem/subvolume_test.go
+++ b/pkg/filesystem/subvolume_test.go
@@ -16,7 +16,12 @@ limitations under the License.
 
 package subvolume
 
-import "testing"
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
 
 func TestGetOmapVal(t *testing.T) {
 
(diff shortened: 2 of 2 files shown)


pkg/filesystem/subvolume.go
```diff
@@ -125,12 +125,27 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
...
 	}
 	return subvolumeNames
 }
 
+// getSubvolumeNameFromPath get the subvolumename from the path.
+// subvolumepath: /volumes/csi/csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d/5f4e4caa-f835-41ba-83c1-5bbd57f6aedf
+// subvolumename: csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d
+func getSubvolumeNameFromPath(path string) string {
+	splitSubvol := strings.Split(path, "/")
+	if len(splitSubvol) < 3 {
```
travisn, line 141: If you're accessing `splitSubvol[3]` below, then we need at least 4 entries to avoid an index exception
```suggestion
	if len(splitSubvol) < 4 {
```

pkg/filesystem/subvolume_test.go
```diff
@@ -59,3 +63,35 @@ func TestGetOmapVal(t *testing.T) {
...
+			name: "5f4e4caa-f835-41ba-83c1-5bbd57f6aedf",
+		},
+		{
+			path: "csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d/5f4e4caa-f835-41ba-83c1-5bbd57f6aedf",
+			name: "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.path, func(t *testing.T) {
+			if name := getSubvolumeNameFromPath(tt.path); name != tt.name {
```
Madhu-1, line 92: i think we dont need if check here, asser.Equal will fail if we don't get the expected value

pkg/filesystem/subvolume.go
```diff
@@ -125,12 +125,27 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	subvolumeNames := make(map[string]subVolumeInfo)
 	for _, pv := range pvList.Items {
 		if pv.Spec.CSI != nil {
-			subvolumeNames[pv.Spec.CSI.VolumeAttributes["subvolumeName"]] = subVolumeInfo{}
+			subvolumePath := pv.Spec.CSI.VolumeAttributes["subvolumePath"]
+			name := getSubvolumeNameFromPath(subvolumePath)
+			subvolumeNames[name] = subVolumeInfo{}
```
Madhu-1, line 130: if the `name` is empty dont we need to skip it here or log it here?

yati1998: I don't think it is required. If there are no subvolumes, the map will be empty and while lookup all the subvolumes from backend will be considered as stale.

Madhu-1: why need to have `subvolumeNames[""]=subVolumeInfo{}` this initialization doesnt make sense. we need to log and skip this if we don't find subvolumename

yati1998: oh sorry, got it.

pkg/filesystem/subvolume_test.go
```diff
@@ -59,3 +63,34 @@ func TestGetOmapVal(t *testing.T) {
...
+		{
+			path: "/volumes/csi/csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d/5f4e4caa-f835-41ba-83c1-5bbd57f6aedf",
+			name: "csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d",
+		},
+		{
+			path: "",
+			name: "",
+		},
+		{
+			path: "/volumes/csi-vol-6a99b552-fdcc-441d-b1e6-a522a85a503d/5f4e4caa-f835-41ba-83c1-5bbd57f6aedf",
```
travisn, line 82: Does this unit test pass? If there are fewer than 4 slashes, it's supposed to return the empty string, so I expect it to fail. 
@subhamkrai Are we running the unit tests in the CI?

yati1998: No it will not fail, the extra slash at initial adds an empty string as first member of list, hence the total length is still 4, and due to this it is not failing. @travisn

pkg/filesystem/subvolume.go
```diff
@@ -125,12 +125,29 @@ func getK8sRefSubvolume(ctx context.Context, clientsets *k8sutil.Clientsets) map
 	subvolumeNames := make(map[string]subVolumeInfo)
 	for _, pv := range pvList.Items {
 		if pv.Spec.CSI != nil {
-			subvolumeNames[pv.Spec.CSI.VolumeAttributes["subvolumeName"]] = subVolumeInfo{}
+			subvolumePath := pv.Spec.CSI.VolumeAttributes["subvolumePath"]
+			name := getSubvolumeNameFromPath(subvolumePath)
```
Madhu-1, line 129: @yati1998 if we are not able to get the name from the PV we need to log it as an error so that the user knows something is wrong, silently continuing can cause problems with expectation miss match

pkg/filesystem/subvolume.go
Madhu-1: need to add continue in `if err!=nil` check to check for other PV

yati1998: added

pkg/filesystem/subvolume.go
yati1998: @Madhu-1 addressed the change.

Madhu-1 (approved):

pkg/filesystem/subvolume_test.go
travisn: oh of course, my error

travisn (commented):
Please investigate the CI, it is failing with:
```
waiting for the pvc clone to be in bound state
```
I'll rerun to see if it was intermittent though. LGTM after the CI passes.

yati1998:
> Please investigate the CI, it is failing with:
> 
> ```
> waiting for the pvc clone to be in bound state
> ```
> 
> I'll rerun to see if it was intermittent though. LGTM after the CI passes.

@travisn  The CI seems to have passed, can you please check again?

PR #354, opened 2025-01-30
maintenance: remove hardcoded crd constants in restoreCrd

OdedViner:
Refactored the RestoreCrd function to accept groupName and
versionResource as parameters instead of relying on the hardcoded
constants crds.CephRookIoGroup and crds.CephRookResourcesVersion.
This improves flexibility and allows the function to handle
different CRDs dynamically.

**Checklist:**

### DIFF
--- a/cmd/commands/restore.go
+++ b/cmd/commands/restore.go
@@ -17,6 +17,7 @@ limitations under the License.
 package command
 
 import (
+	"github.com/rook/kubectl-rook-ceph/pkg/crds"
 	"github.com/rook/kubectl-rook-ceph/pkg/restore"
 	"github.com/spf13/cobra"
 )
@@ -31,6 +32,6 @@ var RestoreCmd = &cobra.Command{
 		verifyOperatorPodIsRunning(cmd.Context(), clientSets)
 	},
 	Run: func(cmd *cobra.Command, args []string) {
-		restore.RestoreCrd(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, args)
+		restore.RestoreCrd(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, crds.CephRookIoGroup, crds.CephRookResourcesVersion, args)
 	},
 }
--- a/pkg/restore/crd.go
+++ b/pkg/restore/crd.go
@@ -32,7 +32,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace,
+	groupName, versionResource string, args []string) {
 	crd := args[0]
 
 	var crName string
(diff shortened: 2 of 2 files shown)


travisn (changes requested):
In the kubectl-rook-ceph repo we will only want to include upstream Rook resources. Instead, I would suggest that this `ResourceCrd()` method could accept arguments for a resource group name. In this repo the command will only set the `ceph.rook.io` group, while external repos that are extending this code could pass a different group name.

cmd/commands/restore.go
```diff
@@ -31,6 +31,6 @@ var RestoreCmd = &cobra.Command{
 		verifyOperatorPodIsRunning(cmd.Context(), clientSets)
 	},
 	Run: func(cmd *cobra.Command, args []string) {
-		restore.RestoreCrd(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, args)
+		restore.RestoreCrd(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, nil, nil, args)
```
travisn, line 34:
```suggestion
		restore.RestoreCrd(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, crds.CephRookIoGroup, crds.CephRookResourcesVersion, args)
```

pkg/restore/crd.go
```diff
@@ -32,7 +32,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string,
+	groupNameOptional, versionResourceOptional *string, args []string) {
```
travisn, line 36: Let's just require these names, it seems simpler this way.
```suggestion
	groupName, versionResource string, args []string) {
```

pkg/restore/crd.go
```diff
@@ -42,7 +43,16 @@ func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorN
 	}
 
 	logging.Info("Detecting which resources to restore for crd %q", crd)
-	crdList, err := k8sclientset.ListResourcesDynamically(ctx, crds.CephRookIoGroup, crds.CephRookResourcesVersion, crd, clusterNamespace)
+	var groupName, versionResource string
+	// Set default values for groupName and versionResource
+	if groupNameOptional == nil || versionResourceOptional == nil {
```
travisn, line 48: No need for these checks if we just require them

travisn (changes requested):

subhamkrai (commented):
Also, fix the CI issues

cmd/commands/restore.go
OdedViner: done

pkg/restore/crd.go
OdedViner: done

pkg/restore/crd.go
OdedViner: done

pkg/restore/crd.go
```diff
@@ -32,7 +32,8 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, args []string) {
+func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace string,
```
travisn, line 35:
```suggestion
func RestoreCrd(ctx context.Context, k8sclientset *k8sutil.Clientsets, operatorNamespace, clusterNamespace,
```

travisn (changes requested):
Still a CI issue, but not sure it's related or intermittent. When you push this small fix, let's see if it passes

pkg/restore/crd.go
OdedViner: done

travisn (approved):

PR #366, opened 2025-04-17
kubectl-ceph-rook: Fix IPv6 address handling in monitor endpoint parsing

aagogolev:
## Description

This PR fixes an issue with the `kubectl-ceph-rook` plugin when working with Ceph clusters configured with IPv6 addresses. The current implementation of the `getMonDetails()` function fails to correctly parse IPv6 monitor endpoints, causing commands like `mons restore-quorum` to fail.

When running `kubectl rook-ceph mons restore-quorum` with IPv6 monitor addresses, the command fails with an error:

```
Info: mon=a, endpoints=[2a02:5501:31:c0a::3]:6789
Info: mon=b, endpoints=[2a02:5501:31:c0a::4]:6789
...
Info: Restoring mon quorum to mon b [2a02
...
parse error setting 'public_addr' to '[2a02'
too many arguments: [--setuser-match-path=/var/lib/ceph/mon/ceph-b/store.db]
Error: failed to run command. failed to run command. command terminated with exit code 1
%!(EXTRA string=failed to extract monmap)
```

The issue occurs because the `getMonDetails()` function uses a simple string split by the first colon to extract the IP address and port from monitor endpoints. This approach works for IPv4 addresses but fails for IPv6 addresses which contain multiple colons.

The fix enhances the `getMonDetails()` function to properly handle IPv6 addresses by checking if the endpoint contains more than one colon (indicating a possible IPv6 address) and using the last colon as the separator between the IP address and port for IPv6 addresses, while maintaining the existing behavior for IPv4 addresses.

### DIFF
--- a/pkg/mons/restore_quorum.go
+++ b/pkg/mons/restore_quorum.go
@@ -297,15 +297,19 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 		if !ok {
 			return []string{}, "", "", fmt.Errorf("failed to fetch mon endpoint")
 		} else if monName == goodMon {
-			goodMonPublicIp, goodMonPort, ok = strings.Cut(monEndpoint, ":")
-			if !ok {
-				return []string{}, "", "", fmt.Errorf("failed to get good mon endpoint and port")
+			host, port, err := ParseMonEndpoint(monEndpoint)
+			if err != nil {
+				return []string{}, "", "", err
 			}
+
+			goodMonPublicIp = host
+			goodMonPort = port
 		} else {
 			badMons = append(badMons, monName)
 		}
 		logging.Info("mon=%s, endpoints=%s\n", monName, monEndpoint)
 	}
+
 	return badMons, goodMonPublicIp, goodMonPort, nil
 }
 
--- a/pkg/mons/mon_endpoints.go
+++ b/pkg/mons/mon_endpoints.go
@@ -18,6 +18,7 @@ package mons
 import (
 	"context"
 	"fmt"
+	"net"
 	"regexp"
 
 	"github.com/rook/kubectl-rook-ceph/pkg/logging"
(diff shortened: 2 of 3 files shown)


pkg/mons/restore_quorum.go
```diff
@@ -297,15 +297,22 @@ func getMonDetails(goodMon string, monEndpoints []string) ([]string, string, str
 		if !ok {
 			return []string{}, "", "", fmt.Errorf("failed to fetch mon endpoint")
 		} else if monName == goodMon {
-			goodMonPublicIp, goodMonPort, ok = strings.Cut(monEndpoint, ":")
-			if !ok {
-				return []string{}, "", "", fmt.Errorf("failed to get good mon endpoint and port")
+			if strings.Count(monEndpoint, ":") > 1 {
```
Madhu-1, line 300: can we use SplitHostPort and ParseIP golang packages to get the ip and port?

Madhu-1 (commented):
@aagogolev changes looks good, can you please squash the commits into 1.

aagogolev:
@Madhu-1 I've squashed the commits into a single commit as requested.

travisn (commented):
Thanks for the fix, the changes look good. Would you might adding a unit test for this? Whenever there is string parsing, unit tests really help prevent future regressions. To focus the unit test only on parsing, this might be simplest:
- Create a helper method in mon_endpoints.go such as `ParseMonEndpoint()`, the implementation would be lines 301-311.
- Add a unit test for that helper in mon_endpoints_test.go

aagogolev:
@travisn Thank you for the feedback! I've implemented the requested changes

travisn:
Great, thanks for adding the unit tests! Now we just need to fix #369 separately to get the CI running again...

travisn:
@aagogolev One thing to look into, it seems the tabs changed to spaces. Did you use gofmt? When the CI is running again it will likely catch this inconsistency.

aagogolev:
@travisn @subhamkrai I've made all the requested changes.Ready for review =)

travisn (approved):
LGTM
@subhamkrai let's discuss when to have the next release

Madhu-1 (approved):

subhamkrai:
> LGTM @subhamkrai let's discuss when to have the next release

Thanks for Merging, Travis I was travelling last week

Issue #388, opened 2025-08-20
csi: subvolume ls should not fail if volumesnapshotcontents CRD is missing


OdedViner:
currently `kubectl-rook-ceph subvolume ls` fails when the
volumesnapshotcontents.snapshot.storage.k8s.io resource does not
exist. the command should continue and log a warning instead of
exiting, so clusters without snapshot CRDs can still use the tool.

PR #389, opened 2025-08-20
csi: skip failure when volumesnapshotcontents crd is missing

OdedViner:
subvolume ls should not fail if volumesnapshotcontents resource does not exist, instead log a warning and continue

Fixes: #388 

**Checklist:**

### DIFF
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -28,6 +28,7 @@ import (
 	"github.com/rook/kubectl-rook-ceph/pkg/exec"
 	"github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
 	"github.com/rook/kubectl-rook-ceph/pkg/logging"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/filesystem/subvolume.go
```diff
@@ -172,7 +173,12 @@ func getK8sRefSnapshotHandle(ctx context.Context, clientsets *k8sutil.Clientsets
 	}
 	snapList, err := snapConfig.VolumeSnapshotContents().List(ctx, v1.ListOptions{})
 	if err != nil {
-		logging.Fatal(fmt.Errorf("Error fetching volumesnapshotcontents: %v\n", err))
+		// ignore only NotFound
```
travisn, line 176: This will only print once, right? Just confirming this method is not called in a loop.

OdedViner: We only call the `getK8sRefSnapshotHandle` function here https://github.com/rook/kubectl-rook-ceph/blob/30d5643e26df579bc62951343af825b8db90d778/pkg/filesystem/subvolume.go#L66

travisn (approved):

PR #399, opened 2025-10-19
csi: skip not-ready cephfs subvolumes in ls and summarize errors

OdedViner:
- avoid aborting when ceph reports eagain for subvolume info
- list healthy entries; skip not-ready ones and keep going
- collect skipped items and print a brief summary warning
- tidy error text in getSubvolumeState to reduce extra noise

**Checklist:**

### DIFF
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	osexec "os/exec"
 	"strconv"
 	"strings"
 	"syscall"
@@ -250,6 +251,9 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 	}
 	fmt.Println("Filesystem  Subvolume  SubvolumeGroup  State")
 
+	// collect subvolumes that are not ready (EAGAIN) to show a summary at the end
+	var notReadyErrors []string
+
 	// this iterates over the filesystems and subvolumegroup to get the list of subvolumes that exist
 	for _, fs := range fsstruct {
 		// gets the subvolumegroup in the filesystem
@@ -276,10 +280,9 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 				// subvolume info returns error in case of pending clone or if it is not ready
 				// it is suggested to delete the pvc before deleting the subvolume.
 				if err != nil {
-					if errors.Is(err, syscall.EAGAIN) {
-						logging.Warning("Found pending clone: %q", sv.Name)
-						logging.Warning("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
-						logging.Warning("To avoid stale resources, please scale down the cephfs deployment before deleting the subvolume.")
+					if isSubvolumeNotReady(err) {
+						// Skip listing this subvolume but remember to report it later
+						notReadyErrors = append(notReadyErrors, fmt.Sprintf("%s/%s/%s: %v", fs.Name, svg.Name, sv.Name, err))
 						continue
 					}
 					logging.Fatal(fmt.Errorf("failed to get subvolume state: %q %q", sv.Name, err))
@@ -317,6 +320,15 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 		}
 	}
+
+	// After listing, show a concise summary of skipped subvolumes due to not-ready state
+	if len(notReadyErrors) > 0 {
+		logging.Warning("%d subvolumes were skipped because they are not ready (pending clone or in progress):", len(notReadyErrors))
+		for _, errMsg := range notReadyErrors {
+			logging.Warning("  %s", errMsg)
+		}
+		logging.Warning("To avoid stale resources, you may scale down the cephfs deployment before deleting such subvolumes.")
+	}
 }
 
 // getSubvolumeState returns the state of the subvolume
@@ -326,7 +338,8 @@ func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, oper
 
 	subVolumeInfo, errvol := runCommand(ctx, clientsets, operatorNamespace, clusterNamespace, cmd, args)
 	if errvol != nil {
-		logging.Error(errvol, "failed to get subvolume info")
+		// Avoid fmt EXTRA artifacts by formatting the error ourselves
+		logging.Error(fmt.Errorf("failed to get subvolume info for %s/%s/%s: %v", fsName, SubvolumeGroup, SubVol, errvol))
 		return "", errvol
 	}
 	var info map[string]interface{}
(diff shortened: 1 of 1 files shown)


pkg/filesystem/subvolume.go
```diff
@@ -259,10 +262,9 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 				// subvolume info returns error in case of pending clone or if it is not ready
 				// it is suggested to delete the pvc before deleting the subvolume.
 				if err != nil {
-					if errors.Is(err, syscall.EAGAIN) {
-						logging.Warning("Found pending clone: %q", sv.Name)
-						logging.Warning("Please delete the pending pv if any before deleting the subvolume %s", sv.Name)
-						logging.Warning("To avoid stale resources, please scale down the cephfs deployment before deleting the subvolume.")
+					if isSubvolumeNotReady(err) || errors.Is(err, syscall.EAGAIN) {
```
BlaineEXE: Is there a reason not to move the `errors.Is(err, syscall.EAGAIN)` into the `isSubvolumeNotReady()` function? It seems like it still makes sense as part of the not-ready detection.

pkg/filesystem/subvolume.go
```diff
@@ -324,6 +336,28 @@ func getSubvolumeState(ctx context.Context, clientsets *k8sutil.Clientsets, oper
...
+	if strings.Contains(msg, "Error EAGAIN") {
+		return true
+	}
+	if strings.Contains(msg, "is not ready for operation") {
+		return true
+	}
+	if strings.Contains(msg, "exit code 11") {
+		// Ceph often maps EAGAIN to exit code 11 for subvolume 'info'
+		return true
+	}
```
BlaineEXE: I have a couple thoughts related here:
1. As a general statement, error handling is fragile when it relies on specific error message text, and we should avoid it where it's not required
2. Ceph doesn't have very good error code returns, and so there are times when Rook does need to do message text checking

Overall, what I would prefer to see is us checking return code ints or error types as much as possible. When we can't, we should try string matching for specific err code names that are not likely to change across Ceph versions or with different localized Ceph builds in a different language. 

If we include the outer `errors.Is(err, syscall.EAGAIN)`, we actually have the start of a good pattern

```go
if errors.Is(err, syscall.EAGAIN) {
   return true
}
if strings.Contains(msg, "EAGAIN") { // I removed "Error" since "EAGAIN" is the critical string
   return true
}
```

In this pattern, we do our best to identify the error using the error type using `errors.Is()`. If that somehow fails, we can fall back to looking for the EAGAIN string.

Also of note, EAGAIN is code 11 -- sometimes output as `-11` by Ceph.

In order to make Rook as robust as possible, we should try to do as much non-string checking as possible before resorting to string checking.

Things we can do better here:
1. Leave string equality checking for last always
2. Try to figure out what error type(s) is actually returned. This isn't always straightforward. When repro'ing an issue, I print the type of the returned error using `%T`. That will give one data point. I also inspect the called code (even dependencies) and see what errors they build. I tried that here but didn't see anything very helpful. With all that, it's possible to have a reasonable idea of what different error types might be getting returned. If any of them codify the integer form of the return code, we should try to get it. [For example.](https://github.com/rook/rook/blob/3cb343f62ad048955e14adf437e019da18c5fae4/pkg/operator/ceph/object/s3-handlers.go#L102-L112)

At the end, the rough flow might look something like:
1. If error `Is(err, syscall.EAGAIN)`
2. If `error.(type).Code` == 11 || `error.(type).Code` == -11 (if code extraction is possible)
3. If error.Error() string contains "EAGAIN"
4. If error.Error() string contains `exit code 11`

1 and 2 are the most ideal checks we can do. 3 is a decent fallback. 4 isn't great, but I am guessing that `exit code %d` is a wrapped message from inside a dependency somewhere that isn't likely to change over time.

I would suggest removing the string check for `is not ready for operation` unless we really, really need that fallback. That text seems somewhat likely to change over enough time.

pkg/filesystem/subvolume.go
OdedViner: done

pkg/filesystem/subvolume.go
OdedViner: image


I manually tested `exitCodeFromError` function.

To force a non-existent subcommand and verify exit-code extraction, I changed the args:
```
// Before
args := []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup, "--format", "json"}

// After (intentional typo to trigger failure)
args := []string{"fs", "subvolume", "info1", fsName, SubVol, SubvolumeGroup, "--format", "json"}
```

Result: the helper correctly extracted exit code 22.
errvol (trimmed):
```
no valid command found; 10 closest matches:
fs subvolume ls <vol_name> [<group_name>]
fs subvolume create <vol_name> <sub_name> ...
...
failed to run command. command terminated with exit code 22
(underlying: k8s.io/client-go/util/exec.CodeExitError)
```

ExitStatus():
```
k8s.io/client-go/util/exec.CodeExitError{Code:22}
```

Parsed code:
22


This demonstrates exitCodeFromError handles wrapped errors (e.g., CodeExitError) and reliably returns the numeric exit code.

BlaineEXE (commented):
The error handling code here looks much more robust. ~I am wondering if you were able to determine which code path is followed when running the command in manual testing. That isn't of critical importance as long as we know it works, but it could be good to have a note of.~

I also see [this comment](https://github.com/rook/kubectl-rook-ceph/pull/399#discussion_r2455030006) describing the testing. Very thorough notes; thank you.

LGTM

pkg/filesystem/subvolume.go
```diff
@@ -317,6 +320,15 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			}
 		}
 	}
+
+	// After listing, show a concise summary of skipped subvolumes due to not-ready state
+	if len(notReadyErrors) > 0 {
+		logging.Warning("Some subvolumes were not ready and were skipped (pending clone or in-progress):")
+		for _, m := range notReadyErrors {
```
subhamkrai:
```suggestion
		for _, m := range notReadyErrors {
```

Could merge the below warnings?

subhamkrai (commented):
Overall looks good, could have meaning full variable. For example in the for loop.

pkg/filesystem/subvolume.go
OdedViner: To force a non-existent subcommand and verify exit-code extraction, I changed the args:
```
// Before
args := []string{"fs", "subvolume", "info", fsName, SubVol, SubvolumeGroup, "--format", "json"}

// Next best: exit codes 11 or -11 (common Ceph EAGAIN)
if code, ok := exitCodeFromError(err); ok {
	if code == 11 || code == -11 {
		return true
	}
}

// After (intentional typo to trigger failure)
args := []string{"fs", "subvolume", "info1", fsName, SubVol, SubvolumeGroup, "--format", "json"}

// Next best: exit codes 11 or -11 (common Ceph EAGAIN)
if code, ok := exitCodeFromError(err); ok {
	if code == 22 || code == -22 {
		return true
	}
}
```

Run the command:
```
$ ./bin/kubectl-rook-ceph subvolume ls
Info: volumesnapshotcontents resource not found, skipping snapshot checks
Filesystem  Subvolume  SubvolumeGroup  State
Error: failed to get subvolume info for myfs/csi/csi-vol-874dc7c7-8d0e-478a-a2e5-16952905e92b: no valid command found; 10 closest matches:
fs subvolume ls <vol_name> [<group_name>]
fs subvolume create <vol_name> <sub_name> [<size:int>] [<group_name>] [<pool_layout>] [<uid:int>] [<gid:int>] [<mode>] [--namespace-isolated] [<earmark>]
fs subvolume rm <vol_name> <sub_name> [<group_name>] [--force] [--retain-snapshots]
fs subvolume authorize <vol_name> <sub_name> <auth_id> [<group_name>] [<access_level>] [<tenant_id>] [--allow-existing-id]
fs subvolume deauthorize <vol_name> <sub_name> <auth_id> [<group_name>]
fs subvolume authorized_list <vol_name> <sub_name> [<group_name>]
fs subvolume evict <vol_name> <sub_name> <auth_id> [<group_name>]
fs subvolume getpath <vol_name> <sub_name> [<group_name>]
fs subvolume info <vol_name> <sub_name> [<group_name>]
fs subvolume exist <vol_name> [<group_name>]
Error EINVAL: invalid command
. failed to run command. command terminated with exit code 22
Warning: 1 subvolumes were skipped because they are not ready (pending clone or in progress):
Warning:   myfs/csi/csi-vol-874dc7c7-8d0e-478a-a2e5-16952905e92b: no valid command found; 10 closest matches:
fs subvolume ls <vol_name> [<group_name>]
fs subvolume create <vol_name> <sub_name> [<size:int>] [<group_name>] [<pool_layout>] [<uid:int>] [<gid:int>] [<mode>] [--namespace-isolated] [<earmark>]
fs subvolume rm <vol_name> <sub_name> [<group_name>] [--force] [--retain-snapshots]
fs subvolume authorize <vol_name> <sub_name> <auth_id> [<group_name>] [<access_level>] [<tenant_id>] [--allow-existing-id]
fs subvolume deauthorize <vol_name> <sub_name> <auth_id> [<group_name>]
fs subvolume authorized_list <vol_name> <sub_name> [<group_name>]
fs subvolume evict <vol_name> <sub_name> <auth_id> [<group_name>]
fs subvolume getpath <vol_name> <sub_name> [<group_name>]
fs subvolume info <vol_name> <sub_name> [<group_name>]
fs subvolume exist <vol_name> [<group_name>]
Error EINVAL: invalid command
. failed to run command. command terminated with exit code 22
Warning: To avoid stale resources, you may scale down the cephfs deployment before deleting such subvolumes.

```

@BlaineEXE @subhamkrai  @yati1998  do you think we need to fix the output?

subhamkrai: We should not print
```
Filesystem  Subvolume  SubvolumeGroup  State
```

We are getting an error, could we improve this?

pkg/filesystem/subvolume.go
OdedViner: done

PR #417, opened 2025-12-02
csi: set default svg as "csi"

ShravaniVangur:
For the subvolume command, If no subvolumegroup is specified, the tool must only scan within the internal subvolumegroup. This PR sets the default subvolumegroup as `csi`. 

**Checklist:**

### DIFF
--- a/cmd/commands/subvolume.go
+++ b/cmd/commands/subvolume.go
@@ -45,7 +45,6 @@ var deleteCmd = &cobra.Command{
 	Use:     "delete",
 	Short:   "Deletes a stale subvolume.",
 	Args:    cobra.RangeArgs(2, 3),
-	Hidden:  true,
 	Example: "kubectl rook-ceph delete <filesystem> <subvolume> [subvolumegroup]",
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx := cmd.Context()


cmd/commands/subvolume.go
```diff
@@ -30,12 +30,14 @@ var SubvolumeCmd = &cobra.Command{
 }
 
 var listCmd = &cobra.Command{
-	Use:   "ls",
-	Short: "Print the list of subvolumes.",
+	Use:     "ls",
+	Short:   "Print the list of subvolumes.",
+	Example: "kubectl rook-ceph subvolume ls --svg svgname --stale",
```
subhamkrai:
```suggestion
	Example: "kubectl rook-ceph subvolume ls",
```
Let's have simple example for the command?

subhamkrai (commented):
@ShravaniVangur also, `subvolume delete` bug is fixed, could you revert this in separate commit or PR?

pkg/filesystem/subvolume.go
```diff
@@ -263,6 +263,9 @@ func listCephFSSubvolumes(ctx context.Context, clientsets *k8sutil.Clientsets, o
 			continue
 		}
 		for _, svg := range subvolg {
+			if subvolgName != "" && svg.Name != subvolgName {
```
Rakshith-R, line 266: There's no need to fetch and loop through subvolumegroups if we expect a subvolumegroup to be always set now ?

let's just directly use it?

ShravaniVangur: In case there's an issue with the svg name/it does not exist, an error such as `Error ENOENT: SVG 'a' not found . failed to run command. command terminated with exit code 2%!(EXTRA string=failed to get subvolumes of %q, string=a)` is encountered and the tool abruptly ends. By checking if it exists in the loop, the tool ignores if not found. 
If the former condition is acceptable then we do not have to check for it's existence.

Rakshith-R: > In case there's an issue with the svg name/it does not exist, an error such as `Error ENOENT: SVG 'a' not found . failed to run command. command terminated with exit code 2%!(EXTRA string=failed to get subvolumes of %q, string=a)` is encountered and the tool abruptly ends. By checking if it exists in the loop, the tool ignores if not found. If the former condition is acceptable then we do not have to check for it's existence.

Listing will fail when user only has permission for a particular subvolumegroup.

Failing with not found is okay.

docs/subvolume.md
```diff
@@ -7,13 +7,16 @@ This would consider all the cases where we can have stale subvolume
 and delete them without impacting other resources and attached volumes.
 
 The subvolume command will require the following sub commands:
+
 * `ls` : [ls](#ls) lists all the subvolumes
-    * `--stale`: lists only stale subvolumes
+  * `--stale`: lists only stale subvolumes
+  * `--svg <subvolumegroupname`: lists subvolumes in a particular subvolume(default is "csi")
```
travisn, line 13: I am a bit late to review, but you could open a follow-up PR to fix this text?
```suggestion
  * `--svg <subvolumegroupname>`: lists subvolumes in a particular subvolume(default is "csi")
```

ShravaniVangur: Sure. Follow-up PR: https://github.com/rook/kubectl-rook-ceph/pull/418

PR #422, opened 2026-01-09
core: add new command for multus validation

subhamkrai:
this commit export the existing multus validation
command from the rook/rook repo https://github.com/rook/rook/tree/master/cmd/rook/userfacing/multus to new command the plugin. This will be easer users when try to validate multus without exec in rook op pod.

**Checklist:**


README.md
```diff
@@ -112,6 +117,7 @@ Visit docs below for complete details about each command and their flags uses.
 1. [Restore deleted CRs](docs/crd.md)
 1. [Destroy cluster](docs/destroy-cluster.md)
 1. [Running rados commands](docs/rados.md)
+1. [Validating Multus configuration](https://rook.github.io/docs/rook/latest/CRDs/Cluster/network-providers/?h=valid#validating-multus-configuration)
```
travisn: Currently I don't see an example in for how to run the new command. The doc in this link only explains how to run it by exec-ing to the pod. Seems like we need a doc in this repo that shows an example CLI.

travisn (changes requested):

README.md
BlaineEXE: Fair point. IMO, a full example is probably more duplication than needed. Do you think it would be sufficient to tell users that they can use this CLI for all `multus validation` commands without having to exec into the operator pod? Maybe with as simple of an example as `kubectl rook-ceph multus validation --help`?

subhamkrai: @travisn @BlaineEXE added new doc should be good now.

@BlaineEXE One question should we remove this from the doc saying `before you install a CephCluster` as we now support day-2 multus support?

BlaineEXE: I would lean toward keeping it. I think it's a good reminder for most users, and I hope it won't distract from users who do use it day-2.

I forget if we have written upstream Rook docs explaining how to support Multus for an already-installed CephCluster. Do you remember?

travisn (approved):

README.md
subhamkrai: I don't think we added a separate doc for day-2 support. 

Btw are we good to merge this PR?

BlaineEXE: I think it's good to merge when you're ready

PR #439, opened 2026-03-24
core: add consumer-context and rados-namespace flag for subvolume

Rakshith-R:
**Description:**

- core: add --consumer-context flag for cross-cluster PVC/Snap lookups

Add a `--consumer-context` global flag that allows PV and VolumeSnapshotContent lookups to target a different
Kubernetes cluster context while Ceph commands execute in the default context.

This supports stretched/external storage scenarios where PVs exist in a consumer cluster separate
from the Ceph cluster. When omitted, behavior is identical to before (consumer clients alias the primary). Currently wired for `subvolume ls` and `subvolume delete` commands.

- test: add integration tests for --consumer-context flag

Adds integration tests with a `create_consumer_context()` helper that creates a kubeconfig
context alias for single-cluster CI testing, verifying `subvolume ls` works with a
valid consumer context and fails cleanly with an invalid one.

---
- core: add --rados-namespace flag to subvolume commands

Add --rados-namespace as a PersistentFlag on the subvolume command so both ls and delete inherit it. The flag specifies the rados namespace for OMAP object/key lookups (default "csi").

Thread radosNamespace through the full snapshot OMAP call chain:
List → listCephFSSubvolumes → checkSnapshot → deleteSnapshot → deleteOmapForSnapshot and getSnapOmapKey, replacing previously hardcoded "csi" values.

Also add unit tests for isSubvolumeNotReady and exitCodeFromError, and update docs with --rados-namespace usage examples.

- test: add integration test for subvolume delete with custom SVG

Add create_stale_subvolume_in_svg() helper that creates a CephFilesystemSubVolumeGroup CR, reads status.info.clusterID, builds a Retain StorageClass with subvolumeGroup set to the SVG
name, then creates and deletes a PVC to produce a stale subvolume.

**Checklist:**

- [x] **Commit Message Formatting**: Commit titles and
  messages follow guidelines in the
  [developer guide](https://rook.io/docs/rook/latest/Contributing/development-flow/#commit-structure).
- [x] Reviewed the developer guide on
  [Submitting a Pull Request](https://rook.io/docs/rook/latest/Contributing/development-flow/#submitting-a-pull-request)

### DIFF
--- a/.github/workflows/go-test-config/action.yaml
+++ b/.github/workflows/go-test-config/action.yaml
@@ -135,6 +135,20 @@ runs:
         kubectl rook-ceph ${NS_OPT} cephfs-snap ls --filesystem myfs --orphaned | grep csi-snap
         kubectl rook-ceph ${NS_OPT} cephfs-snap delete $subVol $snap --filesystem myfs
 
+    - name: Subvolume with consumer-context
+      shell: bash --noprofile --norc -eo pipefail -x {0}
+      run: |
+        set -ex
+        tests/github-action-helper.sh create_consumer_context consumer-ctx
+        kubectl rook-ceph ${NS_OPT} subvolume ls --consumer-context=consumer-ctx
+        kubectl rook-ceph ${NS_OPT} subvolume ls --stale --consumer-context=consumer-ctx
+        # verify invalid context fails
+        if kubectl rook-ceph ${NS_OPT} subvolume ls --consumer-context=nonexistent-ctx 2>/dev/null; then
+          echo "ERROR: expected failure with invalid --consumer-context"
+          exit 1
+        fi
+        echo "Invalid consumer-context correctly rejected"
+
     - name: Get rbd list
       shell: bash --noprofile --norc -eo pipefail -x {0}
       run: |
--- a/.github/workflows/go-test-config/action.yaml
+++ b/.github/workflows/go-test-config/action.yaml
@@ -114,7 +114,7 @@ runs:
         kubectl rook-ceph ${NS_OPT} subvolume delete myfs test-subvol group-a
         tests/github-action-helper.sh create_sc_with_retain_policy ${{ inputs.op-ns }} ${{ inputs.cluster-ns }}
         tests/github-action-helper.sh create_stale_subvolume
-        subVol=$(kubectl rook-ceph ${NS_OPT} subvolume ls --stale | awk '{print $2}' | grep csi-vol)
+        subVol=$(kubectl rook-ceph ${NS_OPT} subvolume ls --stale | awk '{print $2}' | grep csi-vol | head -1)
         kubectl rook_ceph ${NS_OPT} subvolume delete myfs $subVol
 
     - name: CephFS Snapshot command
@@ -148,6 +148,33 @@ runs:
           exit 1
         fi
         echo "Invalid consumer-context correctly rejected"
+
+    - name: Subvolume delete with custom SVG and explicit rados-namespace
+      shell: bash --noprofile --norc -eo pipefail -x {0}
+      env:
+        OP_NS: ${{ inputs.op-ns }}
+        CLUSTER_NS: ${{ inputs.cluster-ns }}
+      run: |
+        set -ex
+        SVG=test-svg
+        # Creates SVG CR, matching Retain StorageClass with
+        # subvolumeGroup=$SVG, PVC, then deletes PVC/PV so the
+        # subvolume is stale.  CephFS CSI always stores OMAP in
+        # the "csi" rados namespace (cluster-level ConfigMap default).
+        tests/github-action-helper.sh create_stale_subvolume_in_svg \
+          "$SVG" "$OP_NS" "$CLUSTER_NS"
+        subVol=$(kubectl rook-ceph ${NS_OPT} subvolume ls --stale \
+          --svg "$SVG" | awk '{print $2}' | grep csi-vol | head -1)
+        # Pass --rados-namespace=csi explicitly to verify the flag
+        # is wired through the delete path.
+        kubectl rook-ceph ${NS_OPT} subvolume delete myfs \
+          "$subVol" "$SVG" --rados-namespace=csi
+        # verify subvolume is gone
+        remaining=$(kubectl rook-ceph ${NS_OPT} subvolume ls --stale \
+          --svg "$SVG" \
+          | awk '{print $2}' | grep -c "^${subVol}$" || true)
+        [ "$remaining" -eq 0 ] || \
+          { echo "ERROR: subvolume $subVol still listed after delete"; exit 1; }
 
     - name: Get rbd list
       shell: bash --noprofile --norc -eo pipefail -x {0}
--- a/docs/cephfs-snapshots.md
+++ b/docs/cephfs-snapshots.md
@@ -13,6 +13,7 @@ The cephfs-snap command will require the following sub commands:
   * `--orphaned`: lists only orphaned snapshots
   * `--svg <subvolumegroupname>`: lists snapshots in a particular subvolume group (default is "csi")
   * `--filesystem <filesystemname>`: lists snapshots in a particular filesystem (default is "myfs")
+  * `--consumer-context <context>`: Kubernetes context for PV and VolumeSnapshotContent lookups (default is the current context)
 * `delete <filesystem> <subvolume> <snapshot>`:
     [delete](#delete) an orphaned snapshot.
   * filesystem: filesystem name to which the snapshot belongs.
@@ -20,6 +21,7 @@ The cephfs-snap command will require the following sub commands:
   * snapshot: snapshot name.
   * `--svg <subvolumegroupname>`: subvolume group name (default is "csi")
   * `--filesystem <filesystemname>`: lists snapshots in a particular filesystem (default is "myfs")
+  * `--consumer-context <context>`: Kubernetes context for PV and VolumeSnapshotContent lookups (default is the current context)
 
 ## ls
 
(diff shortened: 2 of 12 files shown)


docs/subvolume.md
```diff
@@ -8,14 +8,21 @@ and delete them without impacting other resources and attached volumes.
 
 The subvolume command will require the following sub commands:
 
+**Shared flags** (available on both `ls` and `delete`):
```
travisn: For the documentation, since there are only two commands, it would be simpler for the user if the flags are just listed under each command, even if it seems duplicative for us.
This will also avoid the current rendering issue that the shared flags and `ls` and `delete` all show as a single bulleted list.

docs/subvolume.md
```diff
@@ -8,14 +8,21 @@ and delete them without impacting other resources and attached volumes.
...
+  used for OMAP lookups (default is "csi")
+
 * `ls` : [ls](#ls) lists all the subvolumes
   * `--stale`: lists only stale subvolumes
-  * `--svg <subvolumegroupname>`: lists subvolumes in a particular subvolume(default is "csi")
 * `delete <filesystem> <subvolume> [subvolumegroup]`:
     [delete](#delete) a stale subvolume.
   * subvolume: subvolume name.
-  * filesystem: filesystem name to which the subvolume belongs.
-  * subvolumegroup: subvolumegroup name to which the subvolume belong(default is "csi")
```
travisn, line 18: nit: Keep them as a single line. Why wrap the text so short?

docs/subvolume.md
```diff
@@ -60,3 +95,34 @@ $ kubectl rook-ceph subvolume delete myfs csi-vol-427774b4-340b-11ed-8d66-0242ac
...
+
+```bash
+$ kubectl rook-ceph subvolume delete myfs \
+    csi-vol-427774b4-340b-11ed-8d66-0242ac110005 csi \
+    --rados-namespace=my-namespace
+
+Info: Deleting the omap object and key for subvolume
+  "csi-vol-427774b4-340b-11ed-8d66-0242ac110005"
+Info: omap object:"csi.volume.427774b4-340b-11ed-8d66-0242ac110005"
+  deleted
```
travisn: Can we keep lines similar to this as a single line? It will look better rendered too.
```suggestion
Info: omap object:"csi.volume.427774b4-340b-11ed-8d66-0242ac110005" deleted
```

README.md
```diff
@@ -50,6 +50,15 @@ These are args currently supported:
     kubectl rook-ceph --context=$(kubectl config current-context) mons
     ```
 
+5. `--consumer-context`: the name of the Kubernetes context
```
travisn: When the `subvolume ls` or `subvolume delete` commands run, are both of the `context` and `consume-context` potentially used at the same time? In this PR, I only see where the subvolume commands are using the `consumer-context`. And if only using that context, why not just use the default context or the `context` param that is already defined? But perhaps I'm overlooking where both contexts are used by the same command.

travisn (changes requested):
LGTM, mainly a question about the context, and a few nits on the docs.

README.md
Rakshith-R: > When the `subvolume ls` or `subvolume delete` commands run, are both of the `context` and `consume-context` potentially used at the same time? In this PR, I only see where the subvolume commands are using the `consumer-context`. And if only using that context, why not just use the default context or the `context` param that is already defined? But perhaps I'm overlooking where both contexts are used by the same command.

Removed it from main README.md to prevent confusion.

yes, the default cluster context or the one that is passed as --context will be considered to have the cephcluster and used to execute ceph commands.

This new --consumer-context will be used only to list PV & SnapContent to compare against backend ceph objects in the default cluster.

We need both:
- default-context: to execute commands in rook-ceph pod
- consumer-context: list PV and SnapContent

This way we don't have the need to execute ceph cmds from consumer cluster.

docs/subvolume.md
Rakshith-R: done :+1:

docs/subvolume.md
```diff
@@ -10,12 +10,16 @@ The subvolume command will require the following sub commands:
 
 * `ls` : [ls](#ls) lists all the subvolumes
   * `--stale`: lists only stale subvolumes
-  * `--svg <subvolumegroupname>`: lists subvolumes in a particular subvolume(default is "csi")
-* `delete <filesystem> <subvolume> [subvolumegroup]`:
-    [delete](#delete) a stale subvolume.
-  * subvolume: subvolume name.
+  * `--svg <subvolumegroupname>`: target a specific subvolume group (default is "csi")
+  * `--rados-namespace <namespace>`: rados namespace used for OMAP lookups (default is "csi")
+  * `--consumer-context <context>`: Kubernetes context for PV and VolumeSnapshotContent lookups (default is the current context)
```
subhamkrai, line 14: it make sense to go add --consumer-context doc changes with the first commit

docs/subvolume.md
```diff
@@ -10,12 +10,16 @@ The subvolume command will require the following sub commands:
...
+* `delete <filesystem> <subvolume> [subvolumegroup]`: [delete](#delete) a stale subvolume.
   * filesystem: filesystem name to which the subvolume belongs.
-  * subvolumegroup: subvolumegroup name to which the subvolume belong(default is "csi")
+  * subvolume: subvolume name.
+  * subvolumegroup: subvolumegroup name to which the subvolume belongs (default is "csi")
+  * `--svg <subvolumegroupname>`: target a specific subvolume group (default is "csi")
+  * `--rados-namespace <namespace>`: rados namespace used for OMAP lookups (default is "csi")
+  * `--consumer-context <context>`: Kubernetes context for PV and VolumeSnapshotContent lookups (default is the current context)
```
subhamkrai, line 21: same

subhamkrai (changes requested):
Also, squash tests are committed into one commit.

docs/cephfs-snapshots.md
```diff
@@ -46,6 +49,34 @@ Filesystem                         Subvolume
...
+(e.g. stretched or external storage), use
+`--consumer-context` to look up PVs and
+VolumeSnapshotContents on the consumer cluster.
+The default is the current context, which is
+typically the same cluster where the CephCluster
+resides:
+
+```bash
+$ kubectl rook-ceph \
+    --consumer-context=consumer-cluster cephfs-snap ls
```
travisn: IMO it's more readable as a single line. Can you change all the example commands to be a single line? If they are really wide, they could still wrap like this, but most commands should not need to wrap. The output is anyway quite wide.
```suggestion
$ kubectl rook-ceph --consumer-context=consumer-cluster cephfs-snap ls
```

docs/cephfs-snapshots.md
```diff
@@ -46,6 +49,34 @@ Filesystem                         Subvolume
...
+
+Filesystem  Subvolume                                     SubvolumeGroup  Snapshot                                       State
+myfs        csi-vol-aa0099b5-f7a0-49c2-bc97-a810005a9654  csi             csi-snap-3936435c-a14a-4a76-9d0f-71321ac084a9  bound
+myfs        csi-vol-aa0099b5-f7a0-49c2-bc97-a810005a9654  csi             csi-snap-4047546d-b25b-5b87-da08-82432bd195ba  orphaned
+```
+
+```bash
+$ kubectl rook-ceph \
+    --consumer-context=consumer-cluster \
+    cephfs-snap ls --orphaned
```
travisn:
```suggestion
$ kubectl rook-ceph --consumer-context=consumer-cluster cephfs-snap ls --orphaned
```

docs/cephfs-snapshots.md
```diff
@@ -46,6 +49,34 @@ Filesystem                         Subvolume
 myfs                               csi-vol-aa0099b5-f7a0-49c2-bc97-a810005a9654  csi             csi-snap-3936435c-a14a-4a76-9d0f-71321ac084a9  orphaned
 ```
 
+### Using --consumer-context
```
travisn: What about this title name to be more descriptive, instead of using the param name?
```suggestion
### Remote Cluster Context
```

docs/subvolume.md
```diff
@@ -43,6 +47,31 @@ myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110005 svg01 in-use
 myfs csi-vol-427774b4-340b-11ed-8d66-0242ac110007 svg01 stale
 ```
 
+### Using --consumer-context
```
travisn:
```suggestion
### Remote Cluster Context
```

travisn (changes requested):
just a few more small suggestions for the doc

.github/workflows/go-test-config/action.yaml
```diff
@@ -135,6 +135,47 @@ runs:
         kubectl rook-ceph ${NS_OPT} cephfs-snap ls --filesystem myfs --orphaned | grep csi-snap
         kubectl rook-ceph ${NS_OPT} cephfs-snap delete $subVol $snap --filesystem myfs
 
+    - name: Subvolume with consumer-context
```
travisn, line 138: How long do the new tests take? It's difficult to tell since the CI does not time actions individually. Overall, this action takes ~17 minutes recently, but seems to take 25 minutes in this PR. Do these new tests really take ~8 minutes?

travisn (approved):
LGTM, just a question about CI

.github/workflows/go-test-config/action.yaml
Rakshith-R: > How long do the new tests take? It's difficult to tell since the CI does not time actions individually. Overall, this action takes ~17 minutes recently, but seems to take 25 minutes in this PR. Do these new tests really take ~8 minutes?

The svg creation adds a minute or 2, the long wait was because cephcsi container did not pick up the configmap entry for new clusterID.
With the new added restart, it works faster.
Ran it twice comes to ~18-19 mins.

subhamkrai (approved):

PR #440, opened 2026-03-26
csi: add custom pod execution for subvolume and snapshot commands

Rakshith-R:
> For users who do not have access to the parent kubernetes cluster that hosts the ceph cluster.

Add --pod-name, --pod-namespace, --pod-container, --mon-ip, --user-id, and --user-key flags to subvolume and cephfs-snap commands, enabling command execution in any pod with Ceph CLI tools instead of only the rook operator pod.

Refactor pkg/filesystem to use CephFilesystem struct with methods, replacing standalone functions that repeatedly passed ctx, clientsets, and namespace parameters.

**Checklist:**

### DIFF
--- a/.github/workflows/go-test-config/action.yaml
+++ b/.github/workflows/go-test-config/action.yaml
@@ -117,6 +117,37 @@ runs:
         subVol=$(kubectl rook-ceph ${NS_OPT} subvolume ls --stale | awk '{print $2}' | grep csi-vol | head -1)
         kubectl rook_ceph ${NS_OPT} subvolume delete myfs $subVol
 
+        # --- Test with custom pod flags ---
+        CLUSTER_NS=${{ inputs.cluster-ns }}
+        OP_NS=${{ inputs.op-ns }}
+        POD_NAME=$(kubectl -n "$OP_NS" get pod \
+            -l "app=${OP_NS}.cephfs.csi.ceph.com-ctrlplugin" \
+            -o jsonpath='{.items[0].metadata.name}')
+        POD_NS="$OP_NS"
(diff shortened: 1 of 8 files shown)


travisn (changes requested):
These arguments to run in a different pod seem difficult to use, but I understand the scenario is for some external tool to drive it in case the operator pod is not available. So I would suggest we actually don't document these changes in this repo since we don't expect them to be used directly by users of this tool. How about reverting the .md files with this change and we just go with the code changes?

Rakshith-R:
> These arguments to run in a different pod seem difficult to use, but I understand the scenario is for some external tool to drive it in case the operator pod is not available. So I would suggest we actually don't document these changes in this repo since we don't expect them to be used directly by users of this tool. How about reverting the .md files with this change and we just go with the code changes?

:+1:  Works for me, I've removed the .md files.

travisn (approved):
@subhamkrai ptal

subhamkrai (commented):
@Rakshith-R 

Question: Do we have the details(mon-ip, user-id, user-key) stored in some k8s resource or somewhere in the pod from there, if we could fetch the details and only take pod name and namespace as an input?

Rakshith-R:
> @Rakshith-R
> 
> Question: Do we have the details(mon-ip, user-id, user-key) stored in some k8s resource or somewhere in the pod from there, if we could fetch the details and only take pod name and namespace as an input?

I did check out the external cluster fetch in this repo but I don't want to hard code and limit/restrict how mon-ip, user-id, user-key are fetched and used.

The location may change in the future and I would like to preserve the flexibility this tool provides.

subhamkrai (approved):

PR #445, opened 2026-04-24
update getSnapshotHandleId to correctly retrieve snapshotID

ShravaniVangur:
Problem statement: 
Snapshot handles with UUID-based clusterIDs were being parsed incorrectly, causing the snapshot ID extraction to fail. This resulted in bound snapshots being reported as orphaned.

Solution: 
Using binary encoding to decode the snapshot ID from the handle instead of assuming a fixed number of dashes.

The PR also adds the subvolume and snapshot commands to the `README.md` file.

### DIFF
--- a/pkg/filesystem/subvolume.go
+++ b/pkg/filesystem/subvolume.go
@@ -165,7 +165,7 @@ func (f *CephFilesystem) getK8sRefSubvolume() map[string]subVolumeInfo {
 			if strings.Contains(driverName, "cephfs.csi.ceph.com") {
 				volumeHandle := pv.Spec.CSI.VolumeHandle
 				prefix := pv.Spec.CSI.VolumeAttributes["volumeNamePrefix"]
-				name, err := generateSubvolumeNameFromVolumeHandle(prefix, volumeHandle)
+				name, err := generateResourceNameFromCSIHandle(prefix, volumeHandle, "csi-vol-")
 				if err != nil {
 					logging.Error(err, "failed to get subvolume name")
 					continue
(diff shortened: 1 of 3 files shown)


pkg/filesystem/subvolume.go
```diff
@@ -235,19 +246,42 @@ func (f *CephFilesystem) getK8sRefSnapshotHandle() map[string]snapshotInfo {
 	return snapshotHandles
 }
 
-// getSnapshotHandleId gets the id from snapshothandle
-// SnapshotHandle: 0001-0009-rook-ceph-0000000000000001-17b95621-
-// 58e8-4676-bc6a-39e928f19d23
-// SnapshotHandleId: 17b95621-58e8-4676-bc6a-39e928f19d23
+// getSnapshotHandleId extracts the ObjectUUID from a CSI snapshot handle.
+// Handle format: <version:4hex>-<clusterIDLen:4hex>-<clusterID:variable>-<poolID:16hex>-<objectUUID:36chars>
```
travisn: Consider using a [Regular expression parser](https://pkg.go.dev/regexp) to extract the parts of the string. It will be generally simpler and more reliable than a custom parsing approach.

ShravaniVangur: The current parsing mirrors `DecomposeCSIID` from https://github.com/ceph/ceph-csi/blob/e6ca34ab07841647ff3d0a7e17f13dcbbed081ef/internal/util/volid.go#L114 
The handle format is length-prefixed because the clusterID can itself contain dashes, making regex/split-based parsing ambiguous.

Rakshith-R: https://github.com/rook/kubectl-rook-ceph/blob/7903fc91d10f736c369d1b72779a363c4bdd7384/pkg/filesystem/subvolume.go#L180-L206

use this pre-exiting logic, maybe make the default prefix(csi-vol-/csi-snap-) a parameter and pass it too

ShravaniVangur: Done PTAL @Rakshith-R

Rakshith-R (approved):

PR #453, opened 2026-07-14
Format health command for a health report

yati1998:
this commit modifies the health command to print
the details in a good format

Just to note [OK] is in green color

```
ypadia:kubectl-rook-ceph$ ./bin/kubectl-rook-ceph -n openshift-storage health --verbose
Checking Mon Distribution...
Checking Ceph Cluster Health...
Checking OSD Distribution...
Checking Pods Status...
Checking PG Status...
Checking MGR Status...

========================================================================
CLUSTER HEALTH REPORT
========================================================================
Generated: 2026-07-22 04:34:52 UTC
Namespace: openshift-storage

========================================================================
Storage
========================================================================

[OK] Ceph Cluster Health [OK]
	Status: HEALTH_OK

[OK] PG Status [OK]
	Status: 185 PGs in healthy state
	Details:
		- PgState: active+clean, PgCount: 185

========================================================================
K8s Resources
========================================================================

[OK] Mon Distribution [OK]
	Status: 3 mon pods running on 3 different nodes
	Details:
		- 3/3 mons in quorum
		- rook-ceph-mon-a-75cdfcbbcc-pfd96 -> compute-0 (Running)
		- rook-ceph-mon-b-5bf5669d75-5vwlh -> compute-1 (Running)
		- rook-ceph-mon-c-54b867cd8f-ktppn -> compute-2 (Running)

[OK] OSD Distribution [OK]
	Status: 3 osd pods running on 3 different nodes
		- rook-ceph-osd-0-748b4d8c5c-nl4bm -> compute-0 (Running)
		- rook-ceph-osd-1-8457868b5-scwq6 -> compute-2 (Running)
		- rook-ceph-osd-2-6547d7dfc9-gnb6m -> compute-1 (Running)

[OK] Pods Status [OK]
	Status: All 58 pods are Running/Succeeded
	Details:
		- 58 Running/Succeeded across checked namespaces

[OK] MGR Status [OK]
	Status: 2 mgr pod(s) running
		- rook-ceph-mgr-a-678c99b95c-w6vlr -> compute-1 (Running)
		- rook-ceph-mgr-b-84f8554bb5-5mrdq -> compute-2 (Running)

========================================================================
SUMMARY
========================================================================
Total Checks: 6
OK: 6
ypadia:kubectl-rook-ceph$ ./bin/kubectl-rook-ceph -n openshift-storage health
Checking Mon Distribution...
Checking Ceph Cluster Health...
Checking OSD Distribution...
Checking Pods Status...
Checking PG Status...
Checking MGR Status...

========================================================================
CLUSTER HEALTH REPORT
========================================================================
Generated: 2026-07-22 04:35:37 UTC
Namespace: openshift-storage

========================================================================
Storage
========================================================================

[OK] Ceph Cluster Health [OK]
	Status: HEALTH_OK

[OK] PG Status [OK]
	Status: 185 PGs in healthy state
	Details:
		- PgState: active+clean, PgCount: 185

========================================================================
K8s Resources
========================================================================

[OK] Mon Distribution [OK]
	Status: 3 mon pods running on 3 different nodes
	Details:
		- 3/3 mons in quorum

[OK] OSD Distribution [OK]
	Status: 3 osd pods running on 3 different nodes

[OK] Pods Status [OK]
	Status: All 58 pods are Running/Succeeded
	Details:
		- 58 Running/Succeeded across checked namespaces

[OK] MGR Status [OK]
	Status: 2 mgr pod(s) running

========================================================================
SUMMARY
========================================================================
Total Checks: 6
OK: 6
```

### DIFF
--- a/cmd/commands/health.go
+++ b/cmd/commands/health.go
@@ -21,15 +21,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var healthVerbose bool
+
 var Health = &cobra.Command{
-	Use:                "health",
-	Short:              "check health of the cluster and common configuration issues",
-	DisableFlagParsing: true,
-	Args:               cobra.NoArgs,
+	Use:   "health",
+	Short: "check health of the cluster and common configuration issues",
+	Args:  cobra.NoArgs,
 	PreRun: func(cmd *cobra.Command, args []string) {
 		verifyOperatorPodIsRunning(cmd.Context(), clientSets)
 	},
 	Run: func(cmd *cobra.Command, _ []string) {
-		health.Health(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace)
+		health.Health(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, healthVerbose)
 	},
+}
+
+func init() {
+	Health.Flags().BoolVar(&healthVerbose, "verbose", false, "shows detailed check for pods")
 }
--- a/docs/health.md
+++ b/docs/health.md
@@ -1,61 +1,93 @@
 # Health
 
-Health command check health of the cluster and common configuration issues. Health command currently validates these things configurations (let us know if you would like to add other validation in health command):
+The `health` command checks the health of a Ceph cluster and reports common configuration issues. It currently validates the following (let us know if you would like to add other validations):
 
-1. at least three mon pods should running on different nodes
-2. mon quorum and ceph health details
-3. at least three osd pods should running on different nodes
-4. all pods 'Running' status
-5. placement group status
-6. at least one mgr pod is running
+1. **Mon Distribution** — At least three mon pods should be running on different nodes, and mon quorum is intact
+2. **Ceph Cluster Health** — Overall Ceph health status (HEALTH_OK, HEALTH_WARN, HEALTH_ERR)
+3. **OSD Distribution** — At least three OSD pods should be running on different nodes
+4. **All Pods Status** — All pods are in Running/Succeeded state
+5. **PG Status** — Placement groups are in a healthy state
+6. **MGR Status** — At least one mgr pod is running
 
-Health commands logs have three ways of logging:
+Results are organized by category (Storage, K8s Resources) and each check reports one of:
 
-1. `Info`: This is just a logging information for the users.
-2. `Warning`: which mean there is some improvement required in the cluster.
-3. `Error`: This requires immediate user attention to get the cluster in healthy state.
+- **OK** — The check passed.
+- **Warning** — There is a potential issue that should be addressed.
+- **Critical** — Immediate attention is required.
 
-## Output
+### PG States
+
+The following PG states are considered healthy:
+
+- `active+clean` — Normal healthy state
+- `active+clean+scrubbing` — Routine data integrity check (runs daily)
+- `active+clean+scrubbing+deep` — Deep data integrity check (runs approximately weekly)
+- `active+clean+snaptrim` — Routine snapshot trimming
+- `active+clean+snaptrim_wait` — Waiting to start snapshot trimming
+
+Any other PG state is flagged as a warning or critical issue.
+
+## Usage
 
 ```bash
 kubectl rook-ceph health
+kubectl rook-ceph health --verbose
+```
 
-# Info:  Checking if at least three mon pods are running on different nodes
-# Warning:  At least three mon pods should running on different nodes
-# rook-ceph-mon-a-5988949b9f-kfshx                1/1     Running       0          26s
-# rook-ceph-mon-a-debug-6bc9d99979-4q2hd          1/1     Terminating   0          32s
-# rook-ceph-mon-b-69c8cb6d85-vg6js                1/1     Running       0          2m29s
-# rook-ceph-mon-c-6f6754bff5-746rp                1/1     Running       0          2m18s
-#
-# Info:  Checking mon quorum and ceph health details
-# Warning:  HEALTH_WARN 1/3 mons down, quorum b,c
-# [WRN] MON_DOWN: 1/3 mons down, quorum b,c
-#     mon.a (rank 0) addr [v2:10.98.95.196:3300/0,v1:10.98.95.196:6789/0] is down (out of quorum)
-#
-# Info:  Checking if at least three osd pods are running on different nodes
-# Warning:  At least three osd pods should running on different nodes
-# rook-ceph-osd-0-debug-6f6f5496d8-m2nbp          1/1     Terminating   0          19s
-#
-# Info:  Pods that are in 'Running' status
-# NAME                                            READY   STATUS        RESTARTS   AGE
-# csi-cephfsplugin-provisioner-5f978bdb5b-7hbtr   5/5     Running       0          3m
-# csi-cephfsplugin-vjl4c                          2/2     Running       0          3m
-# csi-rbdplugin-cwkc2                             2/2     Running       0          3m
-# csi-rbdplugin-provisioner-578f847bc-2j9ct       5/5     Running       0          3m
-# rook-ceph-mgr-a-7b78b4b4b8-ndpmt                1/1     Running       0          2m7s
-# rook-ceph-mon-a-5988949b9f-kfshx                1/1     Running       0          28s
-# rook-ceph-mon-a-debug-6bc9d99979-4q2hd          1/1     Terminating   0          34s
-# rook-ceph-mon-b-69c8cb6d85-vg6js                1/1     Running       0          2m31s
-# rook-ceph-mon-c-6f6754bff5-746rp                1/1     Running       0          2m20s
-# rook-ceph-operator-78cbdb59bd-4zcsh             1/1     Running       0          62s
-# rook-ceph-osd-0-debug-6f6f5496d8-m2nbp          1/1     Terminating   0          19s
-#
-# Warning:  Pods that are 'Not' in 'Running' status
-# NAME                                          READY   STATUS      RESTARTS   AGE
-#
-# Info:  checking placement group status
-# Info:  2 pgs: 2 active+clean; 449 KiB data, 21 MiB used, 14 GiB / 14 GiB avail
-#
-# Info:  checking if at least one mgr pod is running
-# rook-ceph-mgr-a-7b78b4b4b8-ndpmt                Running     fv-az290-487
+Use `--verbose` to include individual resource details (e.g., pod names, nodes) for each check.
+
+## Example Output
+
+```
+Checking Ceph status ...
+Checking Mon Distribution...
+Checking Ceph Cluster Health...
+Checking OSD Distribution...
+Checking Pods Status...
+Checking PG Status...
+Checking MGR Status...
+
+========================================================================
+CLUSTER HEALTH REPORT
+========================================================================
+Generated: 2026-07-22 04:31:13 UTC
+Namespace: rook-ceph
+
+========================================================================
+Storage
+========================================================================
+
+[OK] Ceph Cluster Health [OK]
+	Status: HEALTH_OK
+
+[OK] PG Status [OK]
+	Status: 185 PGs in healthy state
+	Details:
+		- PgState: active+clean, PgCount: 185
+
+========================================================================
+K8s Resources
+========================================================================
+
+[OK] Mon Distribution [OK]
+	Status: 3 mon pods running on 3 different nodes
+	Details:
+		- 3/3 mons in quorum
+
+[OK] OSD Distribution [OK]
+	Status: 3 osd pods running on 3 different nodes
+
+[OK] Pods Status [OK]
+	Status: All 58 pods are Running/Succeeded
+	Details:
+		- 58 Running/Succeeded across checked namespaces
+
+[OK] MGR Status [OK]
+	Status: 2 mgr pod(s) running
+
+========================================================================
+SUMMARY
+========================================================================
+Total Checks: 6
+OK:           6
 ```
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -18,7 +18,6 @@ package health
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -33,8 +32,20 @@ import (
 )
 
 type cephStatus struct {
-	PgMap  pgMap        `json:"pgmap"`
-	Health healthStatus `json:"health"`
+	PgMap       pgMap        `json:"pgmap"`
+	Health      healthStatus `json:"health"`
+	QuorumNames []string     `json:"quorum_names"`
+	MonMap      monMap       `json:"monmap"`
+}
+
+type monMap struct {
+	NumMons int        `json:"num_mons"`
+	Mons    []monEntry `json:"mons"`
+}
+
+type monEntry struct {
+	Name string `json:"name"`
+	Rank int    `json:"rank"`
 }
 
 type healthStatus struct {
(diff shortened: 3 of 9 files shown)


pkg/health/health.go
```diff
@@ -151,37 +213,85 @@ func getPodRunningStatus(ctx context.Context, k8sclientset kubernetes.Interface,
...
-		} else {
-			logging.Warning("\tPgState: %s, PgCount: %d", pgStatus.StateName, pgStatus.Count)
+func checkPGStatus(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) CheckResult {
+	result := CheckResult{
+		Name:     "PGStatus",
+		Category: "Storage",
+	}
+
+	_, pgStateEntries := unMarshalCephStatus(ctx, clientsets, operatorNamespace, clusterNamespace)
+
```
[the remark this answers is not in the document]
yati1998: That's already fixed in the last change — unMarshalCephStatus now returns an error, and checkPGStatus handles it:

  _, pgStateEntries, err := unMarshalCephStatus(ctx, clientsets, operatorNamespace, clusterNamespace)
  if err != nil {
      result.Status = StatusUnknown
      result.Message = err.Error()
      return result
  }

travisn (commented):
A few thoughts/questions:
- For the report, perhaps we have a `--verbose` flag, and only print the list of pods in that case. By default, it doesn't seem necessary to show all of the mon, mgr, osd, etc pods. They can just query for pods directly if they want to see the details. 
- While the report is being collected, is there any progress display? It maybe helpful to print what is being collected, in case some check gets stuck.

yati1998:
> A few thoughts/questions:
> 
> * For the report, perhaps we have a `--verbose` flag, and only print the list of pods in that case. By default, it doesn't seem necessary to show all of the mon, mgr, osd, etc pods. They can just query for pods directly if they want to see the details.
- yeah that would be better, will update it
> * While the report is being collected, is there any progress display? It maybe helpful to print what is being collected, in case some check gets stuck.
- as of now, there is nothing, let me try to add something

yati1998:
For the report, perhaps we have a --verbose flag, and only print the list of pods in that case. By default, it doesn't seem necessary to show all of the mon, mgr, osd, etc pods. They can just query for pods directly if they want to see the details.
@travisn since verbose was not present from start, it actually would need a lot of change to this as well as odf-cli repo as we have to add flag to the root command. i would suggest lets do it in seperate PR as an enhancement once important features have been implemented.
please share your idea on this

cmd/commands/health.go
```diff
@@ -21,15 +21,20 @@ import (
...
 		verifyOperatorPodIsRunning(cmd.Context(), clientSets)
 	},
 	Run: func(cmd *cobra.Command, _ []string) {
-		health.Health(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace)
+		health.Health(cmd.Context(), clientSets, operatorNamespace, cephClusterNamespace, healthVerbose)
 	},
 }
+
+func init() {
+	Health.Flags().BoolVar(&healthVerbose, "verbose", false, "show individual pod details for each check")
```
subhamkrai:
```suggestion
	Health.Flags().BoolVar(&healthVerbose, "verbose", false, "shows detailed check for pods")
```

pkg/health/health.go
```diff
@@ -50,98 +49,188 @@ type PgStateEntry struct {
...
-	checkMonQuorum(ctx, clientsets, operatorNamespace, clusterNamespace)
+func Health(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, verbose bool) {
+	var results []CheckResult
+
+	checks := []struct {
+		name string
+		run  func() CheckResult
+	}{
+		{"MonDistribution", func() CheckResult {
+			return checkMonDistribution(ctx, clientsets.Kube, clusterNamespace)
```
subhamkrai: Let's include mon quorum status as well

pkg/health/health.go
```diff
@@ -50,98 +49,188 @@ type PgStateEntry struct {
...
 	podList, err := k8sclientset.CoreV1().Pods(clusterNamespace).List(ctx, opts)
 	if err != nil {
-		logging.Error(fmt.Errorf("failed to list %s pods with label %s: %v", daemonType, opts.LabelSelector, err))
+		result.Status = StatusUnknown
+		result.Message = fmt.Sprintf("Failed to list %s pods: %v", daemonType, err)
+		return result
 	}
 
-	var nodeList = make(map[string]string)
+	runningNodeSet := make(map[string]bool)
```
subhamkrai:
```suggestion
	uniqueNodeSet := make(map[string]bool)
```

let's make it clear this map stores unique map

pkg/health/health.go
```diff
@@ -151,55 +240,102 @@ func getPodRunningStatus(ctx context.Context, k8sclientset kubernetes.Interface,
...
+	if runningCount < 1 {
+		result.Status = StatusWarning
+		result.Message = "No mgr pods running (at least 1 required)"
 	}
+
+	return result
 }
 
-func unMarshalCephStatus(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, []PgStateEntry) {
+func unMarshalCephStatus(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) (string, []PgStateEntry, error) {
```
subhamkrai: Instead of calling this twice, `checkPGStatus` and `clusterhealth` can we call this once and store the output, as this method is returning both the required output?

pkg/health/health.go
```diff
@@ -151,55 +240,102 @@ func getPodRunningStatus(ctx context.Context, k8sclientset kubernetes.Interface,
 			podRunning = append(podRunning, podList.Items[i])
 		}
 	}
-	return podRunning, podNotRunning
+	return podRunning, podNotRunning, nil
```
subhamkrai, line 269: @travisn should we list the `podNotRunning` without `verbose` as well?

subhamkrai (changes requested):
Please update the doc and README.md as well.

Overall logging suggestion, let's use `logging.Info/Warning/Fatal/Error`

pkg/health/health.go
yati1998: for now, I am planning revert the verbose implementation as it might cause issue in odf-cli as the we do call Health() there and changing the signature would show error there

subhamkrai: if we implement the same `verbose` in odf-cli it should not cause issue.

pkg/health/health.go
```diff
@@ -50,98 +49,188 @@ type PgStateEntry struct {
...
-	checkPodsOnNodes(ctx, clientsets.Kube, clusterNamespace, "app=rook-ceph-osd")
+	for _, c := range checks {
+		fmt.Fprintf(os.Stderr, "Checking %s...\n", c.name)
+		results = append(results, c.run())
+	}
 
-	fmt.Println()
-	CheckAllPodsStatus(ctx, clientsets.Kube, operatorNamespace, clusterNamespace)
+	printReport(clusterNamespace, results, verbose)
+}
```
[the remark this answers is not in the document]
yati1998: it should be fine, we will update the implementation in odf-cli too, which uses this api

travisn:
@yati1998 Is the example output in the PR description updated since any changes in the PR? I just wanted to understand the output as I'm reviewing, thanks.

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+	"sort"
+	"strings"
+	"time"
+)
+
+var categoryOrder = []string{"Storage", "Compute", "Network", "Object Storage"}
+
+func printReport(clusterNamespace string, results []CheckResult, verbose bool) {
+	printHeader(clusterNamespace)
+	printSummary(results)
```
travisn: Let's print the summary last. With a CLI tool, the last output is what the user sees first.

pkg/health/health.go
```diff
@@ -151,55 +240,102 @@ func getPodRunningStatus(ctx context.Context, k8sclientset kubernetes.Interface,
...
-		if pgStatus.StateName == "active+clean" {
-			logging.Info("\tPgState: %s, PgCount: %d", pgStatus.StateName, pgStatus.Count)
-		} else if strings.Contains(pgStatus.StateName, "down") || strings.Contains(pgStatus.StateName, "incomplete") || strings.Contains(pgStatus.StateName, "snaptrim_error") {
-			logging.Error(fmt.Errorf("\tPgState: %s, PgCount: %d", pgStatus.StateName, pgStatus.Count))
-		} else {
-			logging.Warning("\tPgState: %s, PgCount: %d", pgStatus.StateName, pgStatus.Count)
+func checkPGStatus(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) CheckResult {
+	result := CheckResult{
+		Name:     "PGStatus",
+		Category: "Storage",
```
travisn: Create consts for the category names

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+
+	grouped := groupByCategory(results)
+	printed := make(map[string]bool, len(grouped))
+	for _, category := range categoryOrder {
+		if checks, ok := grouped[category]; ok {
+			printCategory(category, checks, verbose)
+			printed[category] = true
+		}
+	}
+	var remaining []string
```
travisn: How about if we print an error if there are any unknown categories? Seems like we should be able to require a list of known categories since it's all in code. This will keep the code simpler so we don't even need this code that checks for remaining categories.

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, category := range remaining {
+		printCategory(category, grouped[category], verbose)
+	}
+
+	printFooter(results)
+}
+
+func printHeader(clusterNamespace string) {
+	fmt.Println()
+	fmt.Println("CLUSTER HEALTH REPORT")
```
travisn: Add separators around this header, otherwise it's not obvious

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+	fmt.Printf("Namespace: %s\n", clusterNamespace)
+}
+
+func printSummary(results []CheckResult) {
+	ok, warn, crit, unknown := countByStatus(results)
+	fmt.Println()
+	fmt.Println(separator())
+	fmt.Println("SUMMARY")
+	fmt.Println(separator())
+	fmt.Printf("Total Checks: %d  |  OK: %d  |  Warning: %d  |  Critical: %d",
```
travisn: Why print these on the same line? I would suggest to print each status on its own line.

travisn: If a count is 0, then no need to print that line. For example, if everything is OK, the output would be: 
```
Total Checks: 6 
OK: 6
```

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+		len(results), ok, warn, crit)
+	if unknown > 0 {
+		fmt.Printf("  |  Unknown: %d", unknown)
```
travisn: What is an example of an unknown status? That sounds unexpected.

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+			fmt.Printf("     - %s\n", d)
+		}
+	}
+
+	if verbose && len(r.Items) > 0 {
+		fmt.Println("   Items:")
+		for _, item := range r.Items {
+			label := item.Name
+			if item.Namespace != "" {
+				label = item.Namespace + "/" + item.Name
```
travisn: Whenever there is a slash separator, use `path.Join()`
```suggestion
				label = path.Join(item.Namespace, item.Name)
```

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if len(r.Details) > 0 {
+		fmt.Println("   Details:")
+		for _, d := range r.Details {
+			fmt.Printf("     - %s\n", d)
+		}
+	}
+
+	if verbose && len(r.Items) > 0 {
+		fmt.Println("   Items:")
```
travisn: No need to print this line
```suggestion
```

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+			if item.Node != "" {
+				fmt.Printf("     - %s -> %s (%s)\n", label, item.Node, item.Status)
+			} else {
+				fmt.Printf("     - %s (%s)\n", label, item.Status)
+			}
+		}
+	}
+}
+
+func printFooter(results []CheckResult) {
```
travisn: If the summary is printed at the end, no need for a footer since it's basically the same info.

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+	case StatusUnknown:
+		return 2
+	case StatusOK:
+		return 3
+	default:
+		return 4
+	}
+}
+
+func countByStatus(results []CheckResult) (ok, warn, crit, unknown int) {
```
travisn: let's use complete words in most cases like this
```suggestion
func countByStatus(results []CheckResult) (ok, warn, critical, unknown int) {
```

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+)
+
+var categoryOrder = []string{"Storage", "Compute", "Network", "Object Storage"}
+
+func printReport(clusterNamespace string, results []CheckResult, verbose bool) {
```
travisn, line 34: As a follow-up PR it would be nice to allow creation of a json report.

yati1998:
> @yati1998 Is the example output in the PR description updated since any changes in the PR? I just wanted to understand the output as I'm reviewing, thanks.

@travisn updated the description, please check

pkg/health/reporter.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func separator() string {
+	return strings.Repeat("=", 72)
+}
+
+func statusIcon(s CheckStatus) string {
+	switch s {
+	case StatusOK:
+		return "[OK]"
```
travisn, line 167: In the previous version of the report, we had colored output (info=blue, warning=yellow, error=red). Let's print the status in these same colors as well.

travisn (changes requested):

pkg/health/reporter.go
yati1998: noted

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
+Info:  [OK] CephClusterHealth                                       [OK]
+Info:     Status: HEALTH_OK
+Info:
+Info:  [OK] PGStatus                                                [OK]
+Info:     Status: 128 PGs active+clean
+Info:     Details:
+Info:       - PgState: active+clean, PgCount: 128
+Info:
+Info:  ========================================================================
+Info:  Compute
```
subhamkrai:
```suggestion
Info:  <>
```

Can we use other terms like k8s resources or ceph pods, compute indicate calculaitng things.

pkg/health/reporter.go
```diff
@@ -0,0 +1,206 @@
...
+	logging.Info("Total Checks: %d", len(results))
+	if ok > 0 {
+		logging.Info("OK:           %d", ok)
+	}
+	if warning > 0 {
+		logging.Info("Warning:      %d", warning)
+	}
+	if critical > 0 {
+		logging.Info("Critical:     %d", critical)
+	}
```
subhamkrai, line 81:
```suggestion
	if ok > 0 {
		logging.Info("OK:           %d", ok)
	}else if warning > 0 {
		logging.Info("Warning:      %d", warning)
	}else if critical > 0 {
		logging.Info("Critical:     %d", critical)
	}
```

can we use `else if ` or `switch`?

pkg/health/reporter.go
```diff
@@ -0,0 +1,206 @@
...
+		printCheckResult(r, verbose)
+	}
+}
+
+func printCheckResult(r CheckResult, verbose bool) {
+	colorFn := statusColor(r.Status)
+	icon := colorFn(statusIcon(r.Status))
+	label := colorFn(statusLabel(r.Status))
+	logging.Info("")
+	logging.Info("%s %-55s [%s]", icon, r.Name, label)
```
subhamkrai:
```suggestion
	logging.Info("%s %s [%s]", icon, r.Name, label)
```
Any reason for 55 char specific?

pkg/health/reporter_test.go
```diff
@@ -0,0 +1,348 @@
+/*
+Copyright 2023 The Rook Authors. All rights reserved.
```
subhamkrai:
```suggestion
Copyright 2026 The Rook Authors. All rights reserved.
```
same in other places

subhamkrai (commented):
The output looks good to me, more cosmetics chanages are requested.

And overall, when printing let's not use `.<space> ` let's use `%t` for tab space or use `text/tabwriter"` if it make sense.

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
-#
-# Info:  checking placement group status
-# Info:  2 pgs: 2 active+clean; 449 KiB data, 21 MiB used, 14 GiB / 14 GiB avail
-#
-# Info:  checking if at least one mgr pod is running
-# rook-ceph-mgr-a-7b78b4b4b8-ndpmt                Running     fv-az290-487
+## Example Output
+
+```
+Info:  Checking MonDistribution...
```
travisn: It seems odd to print `Info` on every line. With the new report format, I would vote to not have the Info prefix. Let's have a new helper such as `logging.Plain()` that does not have a prefix.

pkg/health/reporter.go
```diff
@@ -0,0 +1,222 @@
...
+func printSummary(results []CheckResult) {
+	ok, warning, critical, _ := countByStatus(results)
+	logging.Info("")
+	logging.Info("%s", separator())
+	logging.Info("SUMMARY")
+	logging.Info("%s", separator())
+	var lines []string
+	lines = append(lines, fmt.Sprintf("Total Checks:\t%d", len(results)))
+	if ok > 0 {
+		lines = append(lines, fmt.Sprintf("OK:\t%d", ok))
```
travisn, line 71: IMO a single space looks better than a tab as there is no need to align the output, but if you prefer with tabs we can keep it. 
```suggestion
		lines = append(lines, fmt.Sprintf("OK: %d", ok))
```

pkg/health/reporter.go
travisn: It is possible to have ok, warning, and critical all greater than 0 and we need to print them all in that case, so we should change it back to individual `if` checks

travisn (changes requested):
The CI is failing, here is the simplified message:
```
Total Checks: 3\nInfo: OK:           2\n" does not contain "Warning:"
```

docs/health.md
subhamkrai: @yati1998 Please update the doc as well with the new command output.

And one more small thing
The command output, header says
```
Checking MonDistribution...
Checking CephClusterHealth...
Checking OSDDistribution...
Checking AllPodsStatus...
Checking PGStatus...
Checking MGRStatus..
```

But the sequence doesn't match the output below
```
Storage
========================================================================

[OK] CephClusterHealth [OK]
	....

[OK] PGStatus [OK]
	....
K8s Resources
========================================================================

[OK] MonDistribution [OK]
	...
[OK] OSDDistribution [OK]
	...
[OK] AllPodsStatus [OK]
	...

[OK] MGRStatus [OK]
...
```

Is this something we can fix? Also,
I think the order should be
1. Checking MonDistribution...
2. Checking OSDDistribution...
3. Checking MGRStatus..
4. Checking CephClusterHealth...
5. Checking PGStatus...

@travisn thoughts?

subhamkrai (changes requested):

docs/health.md
travisn: I don't think it's important for the order of the checking to match the ordering of the health report. Since the report has categories, they can be different.

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
+Info:  [OK] PGStatus                                                [OK]
+Info:     Status: 128 PGs active+clean
+Info:     Details:
+Info:       - PgState: active+clean, PgCount: 128
+Info:
+Info:  ========================================================================
+Info:  K8s Resources
+Info:  ========================================================================
+Info:
+Info:  [OK] MonDistribution                                         [OK]
```
travisn: Let's make the descriptions more human readable
```suggestion
Info:  [OK] Mon Distribution                                         [OK]
```

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
+Info:     Status: 3 mon pods running on 3 different nodes
+Info:     Details:
+Info:       - 3/3 mons in quorum
+Info:
+Info:  [OK] OSDDistribution                                         [OK]
```
travisn: OSD Distribution

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
+Info:     Status: 3 osd pods running on 3 different nodes
+Info:
+Info:  [OK] AllPodsStatus                                            [OK]
```
travisn:
```suggestion
Info:  [OK] Pod Status                                            [OK]
```

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
+Info:     Status: All 12 pods are Running/Succeeded
+Info:     Details:
+Info:       - 12 Running/Succeeded across checked namespaces
+Info:
+Info:  [OK] MGRStatus                                                [OK]
```
travisn: MGR Status

travisn (changes requested):

pkg/health/health.go
```diff
@@ -50,98 +61,201 @@ type PgStateEntry struct {
 	Count     int    `json:"count"`
 }
 
-func Health(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) {
-	logging.Info("Checking if at least three mon pods are running on different nodes")
-	checkPodsOnNodes(ctx, clientsets.Kube, clusterNamespace, "app=rook-ceph-mon")
+func Health(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, verbose bool) {
+	var results []CheckResult
+
+	cephStatus, cephStatusErr := unMarshalCephStatus(ctx, clientsets, operatorNamespace, clusterNamespace)
```
travisn, line 68: Before this line, log a message that it is checking for ceph status. Otherwise, if ceph is not responding, it looks like nothing is happening for ~10s until it times out.

pkg/health/health.go
```diff
@@ -50,98 +61,201 @@ type PgStateEntry struct {
...
-	} else if cephHealthDetails == "HEALTH_ERR" {
-		logging.Error(fmt.Errorf("%s", cephHealthDetails))
+func checkCephClusterHealth(status cephStatus, statusErr error) CheckResult {
+	result := CheckResult{
+		Name:     CheckCephClusterHealth,
+		Category: CategoryStorage,
 	}
+
+	if statusErr != nil {
+		result.Status = StatusUnknown
```
travisn: Unknown still doesn't seem like a valid state. If some status fails, it should be an error. 

I tested the health in minikube where the mon quorum was down, and it results in this report:
```
[??] Ceph Cluster Health [UNKNOWN]
	Status: failed to get ceph status: timed out
. failed to run command. command terminated with exit code 1

[??] PG Status [UNKNOWN]
	Status: failed to get ceph status: timed out
. failed to run command. command terminated with exit code 1
```

Instead, I would expect errors like this:
```
[xx] Ceph Cluster Health [ERROR]
	Status: failed to get ceph status: timed out
. failed to run command. command terminated with exit code 1

[xx] PG Status [ERROR]
	Status: failed to get ceph status: timed out
. failed to run command. command terminated with exit code 1
```

travisn (changes requested):
The report is looking good, but an important change IMO is to report errors instead of unknown.

pkg/health/health.go
```diff
@@ -50,98 +61,202 @@ type PgStateEntry struct {
 	Count     int    `json:"count"`
 }
 
-func Health(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string) {
-	logging.Info("Checking if at least three mon pods are running on different nodes")
-	checkPodsOnNodes(ctx, clientsets.Kube, clusterNamespace, "app=rook-ceph-mon")
+func Health(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespace, clusterNamespace string, verbose bool) {
+	var results []CheckResult
+
+	logging.Plain("Fetching Ceph status...")
```
travisn: "Checking" will be consistent with other logs
```suggestion
	logging.Plain("Checking Ceph status...")
```

docs/health.md
```diff
@@ -1,61 +1,92 @@
...
 kubectl rook-ceph health
+kubectl rook-ceph health --verbose
+```
+
+Use `--verbose` to include individual resource details (e.g., pod names, nodes) for each check.
+
+## Example Output
+
+```
+Checking Mon Distribution...
```
travisn, line 43: Something to consider for a follow-up PR when we have the json health report... They would want to capture the json report separately from this logging that shows "Checking" steps. So perhaps the json report would be printed to stdout (instead of stderr). I also wonder if the text report should anyway be printed to stdout.

travisn (changes requested):
Just one more nit then I'll be ready to approve...

docs/health.md
yati1998: sure, we can take this up in follow up pr

travisn (approved):

subhamkrai (approved):

PR #454, opened 2026-07-15
ci: test custom namespace flaky test

subhamkrai:
with csi-operator managing the csi deployment
no need for different driver/pod names/label
based on different operator namespace. It is
rook-ceph only.

Also, moved the mon restore command to later in the
ci command list as sometime restore mon command take
longer time or ceph status takes time to be healthy.

**Checklist:**

### DIFF
--- a/.github/workflows/go-test-config/action.yaml
+++ b/.github/workflows/go-test-config/action.yaml
@@ -77,18 +77,6 @@ runs:
         set -ex
         kubectl rook-ceph ${NS_OPT} radosgw-admin user create --display-name="johnny rotten" --uid=johnny
 
-    - name: Mon restore
-      shell: bash --noprofile --norc -eo pipefail -x {0}
-      run: |
-        set -ex
-        # test the mon restore to restore to mon a, delete mons b and c, then add d and e
-        kubectl rook-ceph ${NS_OPT} mons restore-quorum a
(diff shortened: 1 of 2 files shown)


.github/workflows/go-test-config/action.yaml
```diff
@@ -121,7 +109,7 @@ runs:
         CLUSTER_NS=${{ inputs.cluster-ns }}
         OP_NS=${{ inputs.op-ns }}
         POD_NAME=$(kubectl -n "$OP_NS" get pod \
-            -l "app=${OP_NS}.cephfs.csi.ceph.com-ctrlplugin" \
+            -l "app=rook-ceph.cephfs.csi.ceph.com-ctrlplugin" \
```
travisn, line 112: The "custom namespace" CI test still installs in a different namespace. Why is it that we don't need the proper operator namespace? I'm missing why the CI passes now, if it still uses a different namespace.

subhamkrai: CSI driver pods in custom namespaces also have `rook-ceph` prefix, example
```yaml
Name:           rook-ceph.cephfs.csi.ceph.com-nodeplugin
Namespace:      test-operator
Selector:       app=rook-ceph.cephfs.csi.ceph.com-nodeplugin
```

travisn: I see now, these are just the pod names, not the provisioner name.

travisn (approved):

PR #455, opened 2026-07-24
core: Add Network MTU Config health check

yati1998:
Adds a new health check that validates network MTU configuration for Ceph clusters. Ceph recommends
jumbo frames (MTU >= 9000) for OSD replication
traffic,and undersized MTU silently degrades
throughput.

MTU Sources
The check collects MTU from the following resources:
- OpenShift Network (config.openshift.io/v1/networks/cluster) — reads status.clusterNetworkMTU and status.networkType from the cluster-scoped singleton
```
./bin/kubectl-rook-ceph -n openshift-storage health 
Checking Ceph status...
Checking Mon Distribution...
Checking Ceph Cluster Health...
Checking OSD Distribution...
Checking Pods Status...
Checking PG Status...
Checking MGR Status...
Checking Network MTU Config...

========================================================================
CLUSTER HEALTH REPORT
========================================================================
Generated: 2026-07-27 12:55:23 UTC
Namespace: openshift-storage

========================================================================
Storage
========================================================================

[OK] Ceph Cluster Health [OK]
	Status: HEALTH_OK

[OK] PG Status [OK]
	Status: 169 PGs in healthy state
	Details:
		- PgState: active+clean, PgCount: 169

========================================================================
K8s Resources
========================================================================

[OK] Mon Distribution [OK]
	Status: 3 mon pods running on 3 different nodes
	Details:
		- 3/3 mons in quorum

[OK] OSD Distribution [OK]
	Status: 3 osd pods running on 3 different nodes

[OK] Pods Status [OK]
	Status: All 43 pods are Running/Succeeded
	Details:
		- 43 Running/Succeeded across checked namespaces

[OK] MGR Status [OK]
	Status: 2 mgr pod(s) running

========================================================================
Network
========================================================================

[!!] Network MTU Config [WARNING]
	Status: Cluster network MTU 1400 is below recommended threshold 8900
	Details:
		- [WARN] OpenShift cluster network MTU: 1400 (network type: OVNKubernetes)

========================================================================
SUMMARY
========================================================================
Total Checks: 7
OK:           6
Warning:      1
```
```
./bin/kubectl-rook-ceph -n openshift-storage health --verbose
Checking Ceph status...
Checking Mon Distribution...
Checking Ceph Cluster Health...
Checking OSD Distribution...
Checking Pods Status...
Checking PG Status...
Checking MGR Status...
Checking Network MTU Config...

========================================================================
CLUSTER HEALTH REPORT
========================================================================
Generated: 2026-07-24 09:10:35 UTC
Namespace: openshift-storage

========================================================================
Storage
========================================================================

[OK] Ceph Cluster Health [OK]
	Status: HEALTH_OK

[OK] PG Status [OK]
	Status: 200 PGs in healthy state
	Details:
		- PgState: active+clean, PgCount: 200

========================================================================
K8s Resources
========================================================================

[OK] Mon Distribution [OK]
	Status: 3 mon pods running on 3 different nodes
	Details:
		- 3/3 mons in quorum
	Items:
		- rook-ceph-mon-a-574ccf8dbd-rxhbb -> compute-2 (Running)
		- rook-ceph-mon-b-55bf7bd758-9rw7h -> compute-1 (Running)
		- rook-ceph-mon-c-59b7b955f8-h2npb -> compute-0 (Running)

[OK] OSD Distribution [OK]
	Status: 3 osd pods running on 3 different nodes
	Items:
		- rook-ceph-osd-0-6cd448cdf8-w9qzl -> compute-0 (Running)
		- rook-ceph-osd-1-6d678c897c-jwdtp -> compute-2 (Running)
		- rook-ceph-osd-2-869d4d5db9-ttvjd -> compute-1 (Running)

[OK] Pods Status [OK]
	Status: All 58 pods are Running/Succeeded
	Details:
		- 58 Running/Succeeded across checked namespaces

[OK] MGR Status [OK]
	Status: 2 mgr pod(s) running
	Items:
		- rook-ceph-mgr-a-7554d5698d-fdxvj -> compute-2 (Running)
		- rook-ceph-mgr-b-7f787bdfd8-28bz6 -> compute-0 (Running)

========================================================================
Network
========================================================================

[!!] Network MTU Config [WARNING]
	Status: 1 MTU source(s) below 9000
	Details:
		- [WARN] OpenShift cluster network MTU: 1400 (network type: OVNKubernetes)
             
========================================================================
SUMMARY
========================================================================
Total Checks: 7
OK:           6
Warning:      1
```

### DIFF
--- a/pkg/health/network.go
+++ b/pkg/health/network.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2026 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package health
+
+import (
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+const mtuWarningThreshold = 8900
+
+func isOpenShiftCluster(ctx context.Context, dynamicClient dynamic.Interface) bool {
+	gvr := schema.GroupVersionResource{
+		Group:    "config.openshift.io",
+		Version:  "v1",
+		Resource: "networks",
+	}
+	_, err := dynamicClient.Resource(gvr).Get(ctx, "cluster", metav1.GetOptions{})
+	return err == nil
+}
+
+func checkNetworkMTUConfig(ctx context.Context, dynamicClient dynamic.Interface) CheckResult {
+	result := CheckResult{
+		Name:     CheckNetworkMTUConfig,
+		Category: CategoryNetwork,
+	}
+
+	osMTU, osNetType, osErr := getOpenShiftClusterMTU(ctx, dynamicClient)
+
+	if osErr == nil {
+		tag := "[INFO]"
+		if osMTU < mtuWarningThreshold {
+			tag = "[WARN]"
+		}
+		result.Details = append(result.Details, fmt.Sprintf("%s OpenShift cluster network MTU: %d (network type: %s)", tag, osMTU, osNetType))
+		result.Status, result.Message = evaluateClusterMTU(osMTU, mtuWarningThreshold)
+	} else {
+		result.Status = StatusOK
+		result.Message = fmt.Sprintf("MTU information not available via cluster APIs: %v", osErr)
+	}
+
+	return result
+}
+
+func getOpenShiftClusterMTU(ctx context.Context, dynamicClient dynamic.Interface) (int, string, error) {
+	gvr := schema.GroupVersionResource{
+		Group:    "config.openshift.io",
+		Version:  "v1",
+		Resource: "networks",
+	}
+
+	obj, err := dynamicClient.Resource(gvr).Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return 0, "", err
+	}
+
+	mtu, _, err := unstructured.NestedInt64(obj.Object, "status", "clusterNetworkMTU")
+	if err != nil {
+		return 0, "", fmt.Errorf("failed to read clusterNetworkMTU: %v", err)
+	}
+	netType, _, err := unstructured.NestedString(obj.Object, "status", "networkType")
+	if err != nil {
+		return 0, "", fmt.Errorf("failed to read networkType: %v", err)
+	}
+
+	return int(mtu), netType, nil
+}
+
+func evaluateClusterMTU(mtu, threshold int) (CheckStatus, string) {
+	if mtu >= threshold {
+		return StatusOK, fmt.Sprintf("Cluster network MTU %d meets recommended threshold %d", mtu, threshold)
+	}
+	return StatusWarning, fmt.Sprintf("Cluster network MTU %d is below recommended threshold %d", mtu, threshold)
+}
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -107,6 +107,15 @@ func Health(ctx context.Context, clientsets *k8sutil.Clientsets, operatorNamespa
 		}},
 	}
 
+	if isOpenShiftCluster(ctx, clientsets.Dynamic) {
+		checks = append(checks, struct {
+			name string
+			run  func() CheckResult
+		}{CheckNetworkMTUConfig, func() CheckResult {
+			return checkNetworkMTUConfig(ctx, clientsets.Dynamic)
+		}})
+	}
+
 	for _, c := range checks {
 		logging.Plain("Checking %s...", c.name)
 		results = append(results, c.run())
--- a/pkg/health/network_test.go
+++ b/pkg/health/network_test.go
@@ -1,0 +1,145 @@
+/*
+Copyright 2026 The Rook Authors. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package health
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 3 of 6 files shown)


travisn (commented):
I see this is still in Draft state, but here is some early feedback...

If running on minikube, I see the following network status:
```
========================================================================
Network
========================================================================

[OK] Network MTU Config [OK]
	Status: MTU information not available via cluster APIs
	Details:
		- OpenShift network config not available (non-OpenShift cluster)
		- CephCluster network provider: default
```

Since this is not an openshift cluster, no network status was actually checked. If not running on that platform, let's skip printing that status. In other words, let's not print the Networking section if we are not on the platform that would allow us to detect the status.

BlaineEXE:
```
========================================================================
Network
========================================================================

[!!] Network MTU Config [WARNING]
	Status: 1 MTU source(s) below 9000
	Details:
		- OpenShift cluster network MTU: 1400 (network type: OVNKubernetes)
		- CephCluster network provider: multus
	Items:
		- default/public-net: role=public, MTU not specified, inherits from host NIC ens224
		- default/private-net: role=cluster, MTU not specified, inherits from host NIC ens224
```

Something I appreciate here is that this is a warning and not an error. Network is so complicated, and there are so many ways of doing things in a valid manner that detecting errors clearly is hard. We should expect false positives on issues.

Is it worth separating "details" and "items"? These seem like pretty similar things to me. especially since the cluster network MTU is reported as a detail but multus items report no MTU as items.

Maybe a more helpful distinction would be "information" and "followup recommended" or something like that. Essentially info vs warn.

Alternately, and possibly additionally, I think it could help users to tell what kind of concern the line item is. For example: 

```
========================================================================
Network
========================================================================

[!!] Network MTU Config [WARNING]
	Status: 1 MTU source(s) below 9000
	Details:
		- [WARN] OpenShift cluster network MTU: 1400 (network type: OVNKubernetes)
		- [INFO] CephCluster network provider: multus
		- [INFO] default/public-net: role=public, MTU not specified, inherits from host NIC ens224
		- [INFO] default/private-net: role=cluster, MTU not specified, inherits from host NIC ens224
```

Since we can detect the `master` from Multus networks (ens224 in this example), should we then inspect the host interface's MTU to see if it has MTU at/above our threshold?

Another layer of challenge that we might face in customer clusters is that it's very rare for customers to specify the raw interface like ens224 for their multus network. Frequently, customer clusters use a VLAN interface or bond interface. The VLAN can also be a child of the bond. I think bonds almost always specify MTU, but I'm not sure about VLANs. They may inherit from their parent. When we look at the host data we have, I'm not sure if the interfaces always say the MTU, or if we would need to do multiple layers of inspection.

pkg/health/network.go
```diff
@@ -0,0 +1,297 @@
...
+
+func evaluateMTU(sources []mtuSource, networkType string, cephNet cephNetworkInfo) (CheckStatus, string, []string) {
+	if len(sources) == 0 {
+		return StatusOK, "MTU information not available via cluster APIs", nil
+	}
+
+	threshold := 9000
+	if isOverlayNetwork(networkType) && !cephNet.IsMultus {
+		threshold = 8900
+	}
```
BlaineEXE: I would recommend not hardcoding the thresholds in the code. Customer deployments may have valid reasons for setting MTU at a value slightly lower than 9000, like 8986 (accounts for overhead). We are likely to get a feature request in the future to make the threshold configurable, so I would suggest passing it through from the top level CLI call all the way down.

Alternately, we can use the more relaxed threshold of 8900 for everything. The difference between 8900 and 9000 is small compared to 1500 and 9000.

pkg/health/network.go
```diff
@@ -0,0 +1,88 @@
...
+	return err == nil
+}
+
+func checkNetworkMTUConfig(ctx context.Context, dynamicClient dynamic.Interface) CheckResult {
+	result := CheckResult{
+		Name:     CheckNetworkMTUConfig,
+		Category: CategoryNetwork,
+	}
+
+	const defaultThreshold = 8900
```
travisn: Is it possible to configure the threshold? If not, let's not call it a default.
```suggestion
	const mtuWarningThreshold = 8900
```

subhamkrai (commented):
@travisn do wee need this for non-openshift cluster as well? can we do it only for odf-cli if this MTU is openshift thing only?

yati1998:
> @travisn do wee need this for non-openshift cluster as well? can we do it only for odf-cli if this MTU is openshift thing only?

@subhamkrai i have already updated and included this change

travisn:
> @travisn do wee need this for non-openshift cluster as well? can we do it only for odf-cli if this MTU is openshift thing only?

@subhamkrai From the earlier comment, the suggestion was to skip the check if not on openshift, so non-openshift users will not see any checks about the network mtu.

travisn:
@yati1998 There are merge conflicts to resolve

pkg/health/network.go
```diff
@@ -0,0 +1,87 @@
...
+	return err == nil
+}
+
+func checkNetworkMTUConfig(ctx context.Context, dynamicClient dynamic.Interface) CheckResult {
+	result := CheckResult{
+		Name:     CheckNetworkMTUConfig,
+		Category: CategoryNetwork,
+	}
+
+	const mtuWarningThreshold = 8900
```
subhamkrai: Let's move this to the top of the file.

pkg/health/network.go
```diff
@@ -0,0 +1,87 @@
...
+	if osErr == nil {
+		tag := "[INFO]"
+		if osMTU < mtuWarningThreshold {
+			tag = "[WARN]"
+		}
+		result.Details = append(result.Details, fmt.Sprintf("%s OpenShift cluster network MTU: %d (network type: %s)", tag, osMTU, osNetType))
+		result.Status, result.Message = evaluateClusterMTU(osMTU, mtuWarningThreshold)
+	} else {
+		result.Status = StatusOK
+		result.Message = "MTU information not available via cluster APIs"
```
subhamkrai: Should we log the error from above?

pkg/health/network.go
```diff
@@ -0,0 +1,87 @@
...
+		Version:  "v1",
+		Resource: "networks",
+	}
+
+	obj, err := dynamicClient.Resource(gvr).Get(ctx, "cluster", metav1.GetOptions{})
+	if err != nil {
+		return 0, "", err
+	}
+
+	mtu, _, _ := unstructured.NestedInt64(obj.Object, "status", "clusterNetworkMTU")
```
subhamkrai:
```suggestion
	mtu, _, err := unstructured.NestedInt64(obj.Object, "status", "clusterNetworkMTU")
```
Handle the error, let's not skip the errors.

pkg/health/network.go
```diff
@@ -0,0 +1,87 @@
...
+	netType, _, _ := unstructured.NestedString(obj.Object, "status", "networkType")
```
subhamkrai:
```suggestion
	netType, _, err := unstructured.NestedString(obj.Object, "status", "networkType")
```

subhamkrai (changes requested):

pkg/health/network.go
```diff
@@ -0,0 +1,93 @@
...
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+)
+
+const mtuWarningThreshold = 8900
```
subhamkrai, line 29: Just one question, and it looks good to me.

Where is the above number coming from?

yati1998: https://docs.redhat.com/en/documentation/red_hat_ceph_storage/3/html/configuration_guide/network_configuration_reference#verifying-and-configuring-the-mtu-value_conf
9000 is the jumbo framework that is recommended to use in all official documents, the 100 less is to catch up with the overlays

BlaineEXE: Most users will want 9000, but I have seen users ask for (and read about recommendations) to use values like 8986 to account for metadata overhead on packets. I'm not sure exactly when users need this, but I don't believe it's necessary to be strict about the 9000 check. 8900 is within 11% of 9000, and I think that's a reasonable threshold where we won't raise false positives when users are deliberately using a lower-padded value for some reason

subhamkrai (approved):

travisn (commented):
@yati1998 Please rebase to resolve the merge conflicts

travisn (approved):

PR #461, opened 2026-07-29
operator: add logs command to read the operator log

jhoblitt:
**Motivation**

Watching the Rook operator log means restarting the watch every time the operator restarts: `kubectl logs -f` exits as soon as the container it attached to does, which is usually the event most worth watching.

**What changed**

`kubectl rook-ceph operator logs` reads the operator log from the namespace the plugin already resolves, with `-f/--follow`, `--tail`, `-p/--previous`, `--since` and `--timestamps`. `--follow` re-establishes the stream after it ends, so it rides through a container crash or a rollout onto a replacement pod.

**Notable decisions**

- A replacement pod is read from the start of its log; a container restarted in place resumes from a checkpoint one second back, so a line may repeat rather than be lost to clock skew.
- Skips the exec-based operator readiness check the sibling subcommands run, since it fails in precisely the crash-looping case where the log matters most.
- Reads one pod (the Deployment is single-replica); when several match, a running non-terminating one is preferred and named on stderr.

AI assistance: Claude wrote the implementation, tests and docs from an interactively agreed design; an adversarial review pass it ran surfaced the missing flag validation and the follow-stops-at-container-exit limitation this PR then addressed.

**Checklist:**

### DIFF
--- a/cmd/commands/operator.go
+++ b/cmd/commands/operator.go
@@ -17,8 +17,24 @@ limitations under the License.
 package command
 
 import (
+	"fmt"
+	"math"
+	"os"
+	"time"
+
 	k8sutil "github.com/rook/kubectl-rook-ceph/pkg/k8sutil"
+	"github.com/rook/kubectl-rook-ceph/pkg/logging"
 	"github.com/spf13/cobra"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	logsFollow     bool
+	logsPrevious   bool
+	logsTimestamps bool
+	logsTail       int64
+	logsSince      time.Duration
 )
 
 // OperatorCmd represents the operator commands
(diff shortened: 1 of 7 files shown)


travisn (commented):
I've struggled with the thought of adding commands like this to the kubectl plugin. It's very simple to use aliases in one's environment to define commands that will make life more convenient. For example, I have aliases such as:
```
alias k=kubectl
alias kn=kubectl -n rook-ceph
alias oplog=kubectl -n rook-ceph logs -l app=rook-ceph-operator
```

Instead of adding this new command with 650 lines of code to maintain, I'd rather encourage sharing of common aliases that people find convenient.

jhoblitt:
The main motivation here is to diverge from kubectl's logs -f behavior and not having to restart the log watch every time the operator restarts.

travisn:
> The main motivation here is to diverge from kubectl's logs -f behavior and not having to restart the log watch every time the operator restarts.

With that motivation in mind I could see this feature adding value, thanks.

cmd/commands/operator.go
```diff
@@ -51,7 +67,60 @@ var setCmd = &cobra.Command{
...
+	Use:     "logs",
+	Short:   "Print the logs of the rook-ceph operator pod",
+	Args:    cobra.NoArgs,
+	Example: "kubectl rook-ceph operator logs -f",
+	PreRunE: func(_ *cobra.Command, _ []string) error {
+		return validateLogsFlags(logsTail, logsSince)
+	},
+	Run: func(cmd *cobra.Command, _ []string) {
+		opts := logOptions(logsFollow, logsPrevious, logsTimestamps, logsTail, logsSince)
+		err := k8sutil.StreamPodLogs(cmd.Context(), clientSets.Kube, operatorNamespace, "app=rook-ceph-operator", opts, os.Stdout)
```
travisn, line 80: Idea for a future PR... Could we generalize this to apply to other pods as well? It could be useful for watching ceph daemon logs. Perhaps it would be done with a command such as:
`kubectl rook-ceph daemon logs -l "app=rook-ceph-mon,mon=a"`

travisn (approved):

PR #463, opened 2026-08-03
core: include noobaa check in health report

yati1998:
this commit includes noobaa check in health report

```
 ./bin/kubectl-rook-ceph -n openshift-storage health --verbose
Checking Ceph status...
Checking Mon Distribution...
Checking Ceph Cluster Health...
Checking OSD Distribution...
Checking Pods Status...
Checking PG Status...
Checking MGR Status...
Checking Node Resource Pressure...
Checking Cluster Capacity...
Checking Network MTU Config...
Checking NooBaa Health...

========================================================================
CLUSTER HEALTH REPORT
========================================================================
Generated: 2026-08-03 10:10:53 UTC
Namespace: openshift-storage

========================================================================
Storage
========================================================================

[OK] Ceph Cluster Health [OK]
	Status: HEALTH_OK

[OK] PG Status [OK]
	Status: 81 PGs in healthy state
	Details:
		- PgState: active+clean, PgCount: 81

[OK] Cluster Capacity [OK]
	Status: Cluster capacity 5.5% used (16.5 GiB / 300.0 GiB)
	Items:
		- .mgr: 0.0% used, 448.5 KiB stored
		- ocs-storagecluster-cephblockpool: 6.3% used, 5.3 GiB stored
		- ocs-storagecluster-cephfilesystem-metadata: 0.0% used, 369.2 KiB stored
		- ocs-storagecluster-cephfilesystem-data0: 0.0% used, 1.2 KiB stored

========================================================================
K8s Resources
========================================================================

[OK] Mon Distribution [OK]
	Status: 3 mon pods running on 3 different nodes
	Details:
		- 3/3 mons in quorum
	Items:
		- rook-ceph-mon-a-78cfdf9c5c-g6pmp -> ypadia-deploy-jwptc-worker-1-pncgx (Running)
		- rook-ceph-mon-b-6967d775bc-4x6t2 -> ypadia-deploy-jwptc-worker-2-rjgww (Running)
		- rook-ceph-mon-c-7bffdc65f9-ctmjg -> ypadia-deploy-jwptc-worker-3-ts4sq (Running)

[OK] OSD Distribution [OK]
	Status: 3 osd pods running on 3 different nodes
	Items:
		- rook-ceph-osd-0-fb9656d89-kkkbp -> ypadia-deploy-jwptc-worker-2-rjgww (Running)
		- rook-ceph-osd-1-f455b9d4c-t7z2z -> ypadia-deploy-jwptc-worker-3-ts4sq (Running)
[...]

PR #465, opened 2026-08-06
core: revert noobaa check in health report

yati1998:
Noobaa is completely separate from Rook.
this commit reverts the changes back


travisn (approved):
LGTM to revert #463, per the comment [here](https://github.com/rook/kubectl-rook-ceph/pull/463#issuecomment-5205685979).

